PR #12, opened 2020-04-17
(WEP-2) OAuth2 authentication

roosnic1:
Finally the first version of the OAuth2 Provider. I abandoned the idea of using [hydra](https://www.ory.sh/hydra/docs/index) because it requires a Postgres database and it needed many additional express routes to fully function. I instead used https://github.com/panva/node-oidc-provider and created a new package called oauth2. It's a full OAuth2 provider with it's own database (mongoDB) and it has access to the user collection in the wepublish database.

It still needs work. For example there needs to be a way to include your own templates for the login screens. But the whole thing is at a stage where it's ready and in dire need to be reviewed.


examples/api/src/index.ts
```diff
@@ -52,9 +59,29 @@ async function asyncMain() {
...
 
+  const oauth2Providers: Oauth2Provider[] = [
+    {
+      name: 'google',
+      discoverUrl: 'https://accounts.google.com',
+      clientId: '617896178757-i6ldn0nni9qtle8o6eu76lv93d78nvfi.apps.googleusercontent.com',
+      clientKey: 't267ZLqkV9dacrkPQp_pF-G2',
+      redirectUri: ['http://localhost:4000/auth/google'],
+      scopes: ['openid profile email']
+    },
```
per2plex, line 70: Get `clientId` and `clientKey` from environment and provide defaults inside `nodemon.json` for development.

examples/api/nodemon.json
```diff
@@ -9,7 +9,5 @@
   ],
   "delay": 250,
   "ignoreRoot": [".git"],
-  "env": {
-    "MONGO_URL": "mongodb://localhost:27017/wepublish"
-  }
```
per2plex, line 14: Leave this in as default for development.

examples/api/package.json
```diff
@@ -12,24 +12,25 @@
     "build": "run-s copy-license sync-package-info build:*",
     "build:typescript": "tsc -b ./src/tsconfig.json",
     "watch": "npm-run-all copy-license sync-package-info --parallel watch:*",
-    "watch:server": "nodemon --inspect=0.0.0.0:9240 ./dist/index.js",
+    "watch:server": "nodemon --exec 'node -r dotenv/config' --inspect=0.0.0.0:9240 ./dist/index.js",
```
per2plex, line 15: `-r` option is directly supported by `nodemon` e.g.: `nodemon -r dotenv/config`. But `dotenv/config` should probably be loaded directly in `index.ts`

examples/api/src/index.ts
per2plex, line 70: Get `clientId` and `clientKey` from enviroment and put defaults into `nodemon.json`.

packages/oauth2/src/server/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+import * as path from 'path'
+import {Provider} from 'oidc-provider'
+import {MongoDBAdapter as OAuth2DBAdapter} from './adapter'
+import {routes} from './routes'
+import {configuration} from './configuration'
+import url from 'url'
+import {MongoDBAdapter as WepublishDBAdapter} from '@wepublish/api-db-mongodb'
+import set from 'lodash/set'
+
+import * as jwks_keys from './keys/keys.json'
```
per2plex, line 13: Get keys from environment so it can be easily set for development or inside docker containers.

packages/oauth2/src/server/routes.ts
```diff
@@ -0,0 +1,220 @@
...
+      }
+    }
+  )
+
+export function routes(app: Application, provider: Provider, dbAdapter: MongoDBAdapter): void {
+  //const { constructor: { errors: { SessionNotFound } } } = provider;
+
+  app.use((req, res, next) => {
+    const orig = res.render
+    // you'll probably want to use a full blown render engine capable of layouts
```
per2plex, line 36: We should consider using some pluggable React components for this.

packages/oauth2/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "main": "./dist/server/index.js",
+  "license": "MIT",
+  "scripts": {
+    "build": "run-p build:*",
+    "build:server": "tsc -b ./src/server",
+    "watch": "npm-run-all copy-license sync-package-info copy-assets --parallel watch:*",
+    "watch:server": "run-p watch:server:*",
+    "watch:server:start": "nodemon --exec 'node -r dotenv/config' --inspect=0.0.0.0:9241 ./dist/server/index.js",
+    "watch:server:build": "tsc -b ./src/server --watch",
+    "copy-assets": "shx cp -R ./src/server/views ./dist/server",
```
per2plex, line 16: Errors if `./dist/server` doesn't exists, create it first.

examples/api/nodemon.json
```diff
@@ -9,7 +9,5 @@
   ],
   "delay": 250,
   "ignoreRoot": [".git"],
-  "env": {
-    "MONGO_URL": "mongodb://localhost:27017/wepublish"
```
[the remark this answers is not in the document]
roosnic1: The env variables declared in the `nodemon.json` file overwrite the variables in the `.env` file. I advise against env vars declaration in files that are tracked by git. I understand the desire to have a project up and running with `install`, `build` and `dev` but in this project you have to do some setup work anyway. I propose we put a `.env.example` file in all the examples and packages. In the README we can describe all the variables. If this is okay for you i would do this after merging in a new PR for the whole project.

examples/api/package.json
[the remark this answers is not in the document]
roosnic1: Didn't know about the direct support. thanks :)
I would not include the `dotenv/config` in the `index.ts` since it shouldn't be used in production. The env vars will be either inside the docker container or delivered by the kubernetes deployment. If you do not include the module it will also not load `.env` files checked into git by mistake.

packages/oauth2/package.json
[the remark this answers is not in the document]
roosnic1: fixed that in the main `package.json`

packages/oauth2/src/server/index.tsx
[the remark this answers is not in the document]
roosnic1: I agree with getting the keys via environment variables. I will change this. But i strongly advise against putting a default inside `nodemon.json` or the `.env.example`. This should fail if you don't know hot to make your own keys.

examples/api/src/index.ts
roosnic1: the `Oauth2Providers` should be in the database anyways. I created a new issue for that. https://wepublish.atlassian.net/browse/WEP-66

packages/oauth2/src/server/routes.ts
roosnic1: Agree. Also created a new issue for that. https://wepublish.atlassian.net/browse/WEP-67

examples/api/nodemon.json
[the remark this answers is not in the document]
per2plex: `nodemon.json` is only used for development, but `nodemon` overriding the `.env` variables is indeed problematic, I've created a `.env.example` file in all relevant packages as you proposed and provided a `setup` script to copy those to `.env`.

packages/oauth2/src/server/index.tsx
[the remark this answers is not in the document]
per2plex: I've provided some default keys inside `.env.example` for now, we could provide some interactive `setup` for the future which allows keys to be generated on the fly.

PR #14, opened 2020-04-28
WEP-21 User mgmt and roles

roosnic1:
This PR introduces roles and permissions for users.


packages/api-db-mongodb/src/migration.ts
```diff
@@ -17,7 +18,7 @@ export const Migrations: Migration[] = [
         strict: true
       })
 
-      await migrations.createIndex({name: 1}, {unique: true})
+      await migrations.createIndex({version: 1}, {unique: true})
```
roosnic1, line 21: @per2plex i guess this needs to be fixed manually in bajour. Otherwise migrations will break.

packages/api-db-mongodb/src/migration.ts
```diff
@@ -94,6 +95,40 @@ export const Migrations: Migration[] = [
...
+  {
+    version: 1,
+    async migrate(db, locale) {
+      const userRoles = await db.createCollection<v1.DBUserRole>(v1.CollectionName.UserRoles, {
+        strict: true
+      })
+
+      await userRoles.createIndex({name: 1}, {unique: true})
+
+      userRoles.insertMany([
```
roosnic1, line 108: @per2plex not sure where to insert entries like that. I guess it makes sense here but that means this has to be copied to the other db adapters as well. What do you think?

packages/api-db-mongodb/src/schema/index.ts
```diff
@@ -1 +1,4 @@
 export * from './0'
+export * from './1'
+export {DBUser} from './1'
+export {CollectionName} from './1'
```
roosnic1, line 4: This doesn't look right to me. Was the idea to copy the whole schema `0` and then add the changes and make a new version or what was the plan?

packages/api-db-mongodb/src/migration.ts
per2plex: I think this is fine for now. Another way would be to standardize the initialization process. But I also thought about abandoning DB adaptability in the future in favor of using MongoDB directly which would allow for easier pluggability.

packages/api-db-mongodb/src/schema/index.ts
per2plex: They way you've implemented it was the way I've planned. Because we don't have an explicit schema in MongoDB I thought having types for all the "schema" versions would be nice and help in migrating. But I can see how this would get convoluted quickly and there's no way to add new values into the `CollectionName` enum.

packages/api/src/graphql/permissions.ts
```diff
@@ -0,0 +1,178 @@
...
+  CanGetImages,
+  CanCreateArticle,
+  CanGetArticle,
+  CanGetArticles,
+  CanPublishArticle,
+  CanCreatePage,
+  CanGetPage,
+  CanGetPages,
+  CanPublishPage
+]
```
per2plex, line 178: The typing on the right side (e.g: `<Permission>`) shouldn't be necessary and can be removed.

roosnic1: Done.

PR #17, opened 2020-06-04
E2E Testing

roosnic1:


packages/api/src/server.ts
```diff
@@ -35,7 +35,8 @@ export class WepublishServer {
     })
 
     const corsOptions = {
-      origin: '*',
+      origin: true,
+      credentials: true,
```
per2plex, line 39: Setting `credentials: true` might be dangerous in combination with `origin: true` if we support cookie authorization in the future, that's one of the reason it isn't allowed in combination with `origin: "*"`. It shouldn't be needed at the moment anyway `Authorization` is in the `allowedHeaders`.

The only way to add `credentials: true` securely would be to allow the configuration of the `origin`.

package.json
```diff
@@ -58,7 +58,8 @@
     "clean:website-example": "yarn workspace @wepublish/website-example clean",
     "build:website-example": "yarn workspace @wepublish/website-example build",
     "prettier": "prettier --write '{packages,services}/**/*.{js,ts,tsx,json}'",
-    "prettier-check": "prettier --check '{packages,services}/**/*.{js,ts,tsx,json}'"
+    "prettier-check": "prettier --check '{packages,services}/**/*.{js,ts,tsx,json}'",
+    "test:e2e:chromium": "node -r dotenv/config node_modules/.bin/testcafe chromium ./tests/e2e.ts"
```
per2plex, line 62: Is the idea to add a script for all popular browsers, or should we make this configurable via an environment variable?

PR #21, opened 2020-06-10
Setup docs

dev7ch:
Hey hey,

During setup I stumbled to some issue while getting the api running, the local api domain was not exposed.

After this adjustment in composing files i got it working ... what do you think about this changes @roosnic1 ?

### DIFF
--- a/packages/editor/src/client/api/index.ts
+++ b/packages/editor/src/client/api/index.ts
@@ -13,11 +13,7 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
-  /**
-   * A date-time string at UTC, such as 2007-12-03T10:15:30Z, compliant with the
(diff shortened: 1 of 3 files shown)


packages/editor/src/client/api/index.ts
```diff
@@ -28,7 +24,7 @@ export type Scalars = {
 };
 
 export type Article = {
-   __typename?: 'Article';
+  __typename?: 'Article';
```
roosnic1, line 27: Did your editor do this kind of linting? because there's a git pre commit hook that does the linting

dev7ch: actually i assume it was done by IDE (JetBrains) ... i could drop this if its not correct

roosnic1: yes please. All the linting and formating should be done by the pre-commit hook.

PR #25, opened 2020-06-30
Iframe style height

dev7ch:
## What was changed

- allow iframe to set style (`width` and `height`)
- disable auto ratio if styles are set
- make `title, width, height, style` optional
- shorten and update iframe embed edit instructions in editor
- introduce `styleCustom` as `style` attribute for `<iframe>` input

## What problems are solved

- overwriting the auto ratio calculation for custom iframe by setting styles
- more flexibility for iframe embeds
- make the publisher able to remove `style` and `width` and `title` to avoid invalid values on reedit
- removes wrong padding from in editor if `style` is  used


packages/api-db-karma/src/model.ts
```diff
@@ -166,7 +166,10 @@ export const EmbedBlockModel = struct({
   url: optional(string),
   title: optional(string),
   width: optional(int32),
-  height: optional(int32)
+  height: optional(int32),
```
roosnic1, line 169: the `api-db-karma` database adapter is currently not maintained

packages/api/src/db/block.ts
```diff
@@ -85,6 +85,9 @@ export interface EmbedBlock {
   title?: string
   width?: number
   height?: number
+  styleHeight?: string
+  styleWidth?: string
+  ratio?: boolean
```
roosnic1, line 90: maybe a more explaining name like `useRatio` and i'm not sure if a boolean should be optional. I'd suggest a default in this case maybe true or wathever it was before.

packages/editor/src/client/blocks/embeds/iframe.tsx
```diff
@@ -6,23 +6,43 @@ export interface IframeEmbedProps {
...
+  width = 300,
+  height = 300,
+  styleHeight,
+  styleWidth,
+  ratio
+}: IframeEmbedProps) {
+  const ratioVal = width / height
+  const styles = {}
+  if (!ratio) {
+    Object.assign(styles, {
```
roosnic1, line 26: Why the `Object.assign(...)`

packages/editor/src/client/blocks/embeds/iframe.tsx
```diff
@@ -6,23 +6,43 @@ export interface IframeEmbedProps {
...
+      width: styleWidth,
+      height: styleHeight
+    })
+  }
   return (
     <Box width="100%">
-      <Box position="relative" paddingTop={`${(1 / ratio) * 100}%`} width="100%">
+      <Box
+        position="relative"
+        paddingTop={`${ratio || styleHeight === undefined ? (1 / ratioVal) * 100 + '%' : '0'}`}
```
roosnic1, line 35: I don't understand this logic

packages/editor/src/client/blocks/embeds/iframe.tsx
```diff
@@ -6,23 +6,43 @@ export interface IframeEmbedProps {
...
+        width="100%"
+        minHeight={'45px'}>
```
roosnic1, line 37: Why the `minHeight`?

packages/editor/src/client/panel/embedEditPanel.tsx
```diff
@@ -17,14 +17,13 @@ import {EmbedBlockValue, EmbedType} from '../blocks/types'
...
   readonly value: EmbedBlockValue
-
   onClose(): void
   onConfirm(value: EmbedBlockValue): void
 }
 
 export function EmbedEditPanel({value, onClose, onConfirm}: EmbedEditPanel) {
   const [errorMessage, setErrorMessage] = useState<string>()
-
+  const [ratio, setRatio] = useState<boolean>(true)
```
roosnic1, line 26: shouldn't it get the start value from `value`?

packages/editor/src/client/panel/embedEditPanel.tsx
```diff
@@ -38,7 +37,7 @@ export function EmbedEditPanel({value, onClose, onConfirm}: EmbedEditPanel) {
     const instagramMatch = input.match(/instagram.com\/p\/([0-9a-zA-Z-_]+)/)
     const twitterMatch = input.match(/twitter.com\/([0-9a-zA-Z-_]+)\/status\/([0-9]+)/)
     const vimeoMatch = input.match(/vimeo.com\/([0-9]+)/)
-    const youTubeMatch = input.match(/youtube.com\/watch\?v=([0-9a-zA-Z-_]+)/)
+    const youTubeMatch = input.match(/youtube.com\/watch\?v=([0-9a -zA-Z-_]+)/)
```
roosnic1, line 40: why the space?

packages/editor/src/client/panel/embedEditPanel.tsx
```diff
@@ -66,17 +65,37 @@ export function EmbedEditPanel({value, onClose, onConfirm}: EmbedEditPanel) {
 
         if (iframe) {
           const soundCloudMatch = iframe.src.match(/api.soundcloud.com\/tracks\/([0-9]+)/)
+          const ratioVal = iframe.hasAttribute('style')
```
roosnic1, line 68: I don't understand this and the following code. Maybe we should look at this together.

packages/api/src/db/block.ts
dev7ch: yes indeed the default of this is `true`, currently set as initial state

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: Coz it allow you to just delete the style at the embed input and you get rid of the useless default values.

The same logic should be applied to `height`, `width` and `title`.

The Object is updated on input change.

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: because if height if 0 you are not able tp click the edit button if styles are applied

packages/editor/src/client/panel/embedEditPanel.tsx
dev7ch: oh mhh by linting or mistake

packages/editor/src/client/panel/embedEditPanel.tsx
dev7ch: ideally yes, but this is listening to the input and changes, a state was more reliable due i had some irregular behaviour with the value

packages/editor/src/client/panel/embedEditPanel.tsx
dev7ch: it basically just checks if `style` attribute is in place or not ... if yes the state mentioned above is set to `false`

dev7ch:
the Current emded block has massive limitation to implement an iframe with fixed width and height, especially on mobile.

image

Its not about the difference sizes of the box and the iframe, its about beeing able to modify by `px`, `%`,  `vh` or `vw`

Further the iframe width and height attributes are only pixels, by setting the styles you have more flexibility and are able to overwrite the `width` and `height` attributes. 
In the current implementation you are not able so set a fixed size for the iframe on all screen, this workaround solves the problem by allowing optional adding styles (and disable ratio calc if styles are set)

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: by default `useRatio` or `ratio` is true which means the padding is calculated BUT calc of padding is not always fitting and results to problems

dev7ch:
But yes, there  still some work to do.

dev7ch:
@roosnic1 i remove the `useRatio` and made all attributes except `src`  optional for an custom embed.

This solves as well some some bugs with calced height in the Editor, plz let me know what you think

packages/editor/src/client/panel/embedEditPanel.tsx
dev7ch: ratio was droped

packages/editor/src/client/panel/embedEditPanel.tsx
dev7ch: is removed and more readable and explainable implemented

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: is needed in editor

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: replaced by more explainable vars and logic

packages/api/src/db/block.ts
dev7ch: ratio is dropped

packages/api-db-karma/src/model.ts
dev7ch: reverted

examples/website/src/shared/utility.ts
```diff
@@ -113,3 +113,27 @@ export function createStyleRenderer() {
     plugins: [felaPrefixer(), felaFallbackValue()]
   })
 }
+
+export function transformCssStringToObject(styleCustom: any) {
+  const styleRules = !!styleCustom ? styleCustom.split(';') : []
```
dev7ch, line 118: @roosnic1 this function is exists 2 times, due i assume the website should work independent from Editor and api

dev7ch: Explanation: I tried the approach to store a object first, but changed it back due i see i would be a limitation and type inconvenience to store a string style value as object.

2 main reason:

- if you think beyond React or VueJS implementations, keep the original type
- currently in editor, it would mean to retransform string to object on each input

dev7ch:
@roosnic1 any updates here?

dev7ch:
@roosnic1 👀 ???

packages/editor/src/client/panel/embedEditPanel.tsx
```diff
@@ -119,13 +126,19 @@ export function EmbedEditPanel({value, onClose, onConfirm}: EmbedEditPanel) {
...
+            Embed codes attributes <code>title, src, width, height</code> are validated e.g.
           </Typography>
+          <code>{'<iframe width="560" height="315" src="https://..."></iframe>'}</code>
           <Typography variant="subtitle1" spacing="small">
-            In iframe embed codes the src="", width="" and height="" are validated e.g.
-            {'<iframe src="https://www.youtube.com/embed/id" width="560" height="315"></iframe>'}
+            Set <code>style</code> alternatively as attribute to overwrite the defaults and disable
+            auto ratio size scaling .e.g.
           </Typography>
+          <code>{'<iframe style="height:350px;width:100%" src="https://..."></iframe>'}</code>
```
roosnic1, line 141: missing semicolon in the style attribute at the end.

examples/website/src/shared/blocks/embedBlock.tsx
```diff
@@ -62,21 +63,26 @@ function embedForData(data: EmbedData) {
...
+        noRatio && data.styleCustom !== ''
+          ? transformCssStringToObject(data.styleCustom)
+          : {
+              width: '100%',
+              height: '100%'
+            }
       return (
         <div style={{position: 'relative', width: '100%'}}>
-          <div style={{width: '100%', paddingTop: `${(1 / ratio) * 100}%`}} />
+          <div style={{width: '100%', paddingTop: `${!noRatio ? (1 / ratio) * 100 + '%' : '0'}`}} />
```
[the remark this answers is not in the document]
dev7ch: do you want me to reverse the condition?

packages/editor/src/client/panel/embedEditPanel.tsx
dev7ch: not mandatory on last item, but sure i can add it

packages/editor/src/client/panel/embedEditPanel.tsx
```diff
@@ -66,17 +63,27 @@ export function EmbedEditPanel({value, onClose, onConfirm}: EmbedEditPanel) {
...
-              type: EmbedType.Other,
-              url: iframe.src,
+            // add attributes if set
+            const setEmbedOther = {
               title: iframe.title,
               width: iframe.width ? parseInt(iframe.width) : undefined,
               height: iframe.height ? parseInt(iframe.height) : undefined
+            }
+            // Add styles if set
+            if (!!iframe.style && iframe.hasAttribute('style')) {
```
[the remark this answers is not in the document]
dev7ch: its just a validation and ensure the style attribute really exist, a double validation

examples/website/src/shared/utility.ts
```diff
@@ -113,3 +113,27 @@ export function createStyleRenderer() {
...
+  const styleCustomCss = {}
+  if (styleRules.length > 0) {
+    // Filter css string and transform to k => v styles object
+    const tuples = styleRules
+      .map((rule: any) => {
```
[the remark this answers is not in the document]
dev7ch: yes, but i decided for readability and extendability first.

Can you please provide a proposal how you want this to be handled? In a reduce function.

packages/editor/src/client/panel/embedEditPanel.tsx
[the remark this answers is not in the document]
dev7ch: it applies the style variable only if the are set (please check out the branch and test it your self, i tried to explain this several times)

examples/website/src/shared/blocks/embedBlock.tsx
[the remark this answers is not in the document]
roosnic1: I think it's easier to read `noRatio ? '0' : (1 / ratio) * ....`

examples/website/src/shared/utility.ts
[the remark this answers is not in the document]
roosnic1: ` const cssStyles = styleRules.reduce((styles, currentStyle) => {
    const [key, value] = rule.split(':')
    if (key && value) {
      styles[key.trim()] = value.trim()
    }
    return styles
}, {})
`

packages/editor/src/client/panel/embedEditPanel.tsx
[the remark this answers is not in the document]
roosnic1: I don't think you need the double validation. If `!!iframe.style` is true then it has the attribute.

And the reason that you tried to explain this several times is why we need a comment explaining what is happening here. Otherwise someone else will get confused when they reach this code.

examples/website/src/shared/blocks/embedBlock.tsx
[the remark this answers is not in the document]
dev7ch: ok

packages/editor/src/client/panel/embedEditPanel.tsx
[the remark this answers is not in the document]
dev7ch: ok i ll ad comments and remove double check

examples/website/src/shared/utility.ts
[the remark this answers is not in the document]
dev7ch: thanks :)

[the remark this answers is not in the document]
dev7ch: but there do you take or use `currentStyle` in your function  @roosnic1 ?


```
export function transformCssStringToObject(styleCustom: any) {
  const styleRules = !!styleCustom ? styleCustom.split(';') : []
  const styleCustomCss = {}

  if (styleRules.length > 0) {
    // Filter css string and transform to k => v styles object
      const tuples = styleRules
      .map((rule: any) => {
        let [key, value] = rule.split(':')
        if (key && value) {
          key = key.trim()
          value = value.trim()
          return {[key]: value}
        } else {
          return null
        }
      })
      .filter((x: any) => {
        return x !== null
      })
    Object.assign(styleCustomCss, ...tuples)
  }
  return styleCustomCss
}
```

I still think this is more Developer friendly (as this is important for you regarding your next comments)

examples/website/src/shared/utility.ts
dev7ch: the values are taken from a real iframe html element, ` styleCustom: !!iframe.style && !!iframe.style.cssText ? iframe.style.cssText : ''`

This and the transform function will ensure that only css properties with valid values will be set (e..g you cannot save wron css)

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: due a filter in the editor render FE is done, the object assign can be dropped (no more unexpected behavings on input)

examples/website/src/shared/utility.ts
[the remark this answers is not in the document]
dev7ch: Ok, finally i was able to usr reduce(), but for cleaning the array from whitespaces and empty string in array a map() and filter() are still in use (i added comments to explain why)

packages/editor/src/client/blocks/embeds/iframe.tsx
dev7ch: we did no spoke about that again for a while, i assume this is ok, plz re-resolve if you dont agree.

PR #27, opened 2020-07-29
hotfix iframe styles

dev7ch:
- remove white space removal
- fix condition, prevent undefined


dev7ch:
@roosnic1 we can drop the white space removal, so all valid css should work

roosnic1:
Good. Open a new PR and i'll merge it.

PR #31, opened 2020-08-06
Introduce lerna to the project

roosnic1:
[Lerna](https://github.com/lerna/lerna) helps us with versioning, tagging and releasing of wepublish

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ yarn-error.log
 .tsbuildinfo
 .media
 .env
+.changelog
--- a/lerna.json
+++ b/lerna.json
@@ -22,7 +22,8 @@
       "ignoreChanges": [
         "*.md",
         "*.txt"
-      ]
+      ],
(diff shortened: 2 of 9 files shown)


dev7ch:
What still remains is a relation to github tags to be able to browser the release codebase at github.

Do you already have an idea how to solve that @roosnic1  @0815pascal?

Can be helpful
https://github.community/t/tag-creation-trigger-not-working-with-lerna-publish/17314

dev7ch:
what is the reason of `--no-git-tag-version` in the publish command?

https://github.com/wepublish/wepublish/blob/10bd05fb1c6a37bf87fe7a5bbd448ec6982fba0b/package.json#L22

roosnic1:
@dev7ch git tags will be handled by Lerna and if you check the actual command behind `publish:api-db-mongodb` you'll see that it's the old publish command. The new one will follow in a future PR.

PR #43, opened 2020-09-10
F/wep 77 map leaflet items

0815pascal:
Autocomplete for input-form works fine now


.gitignore
```diff
@@ -12,3 +12,4 @@ yarn-error.log
 .media
 .env
 .changelog
+.vscode/
```
roosnic1, line 15: This belongs into your personal `.gitignore`. I can show you how to do this.

packages/editor/src/client/blocks/mapLeafletBlock.tsx
```diff
@@ -22,148 +25,76 @@ Icon.Default.mergeOptions({
...
-            placeholder="Title"
-            value={title}
-            onChange={e => {
-              const title = e.target.value
-              onChange(value => ({...value, title}))
+      <Box position="relative" width="100%" height="100%">
+        <Map
+          center={[centerLat, centerLng]}
+          zoom={zoom}
+          style={{width: '100%', height: '45vh'}}
```
roosnic1, line 40: why this height?

packages/editor/src/client/blocks/mapLeafletBlock.tsx
```diff
@@ -22,148 +25,76 @@ Icon.Default.mergeOptions({
...
+          onViewportChange={viewport => {
+            if (viewport.center && viewport.zoom) {
```
roosnic1, line 42: Is there a case where it only has either `center` or `zoom`?

packages/editor/src/client/blocks/mapLeafletBlock.tsx
```diff
@@ -22,148 +25,76 @@ Icon.Default.mergeOptions({
...
+              onChange({
+                ...value,
+                centerLat: viewport.center[0],
+                centerLng: viewport.center[1],
+                zoom: viewport.zoom
+              })
+            }
+          }}>
+          <TileLayer
+            url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
```
roosnic1, line 52: This URL and the `attribution` should be configurable. This is something for the settings tab in the editor which we don't have yet. But for now i would put those two into the `.env` file of the editor. But be careful to check for the values and use the current ones as defaults.

packages/editor/src/client/blocks/mapLeafletBlock.tsx
```diff
@@ -22,148 +25,76 @@ Icon.Default.mergeOptions({
...
+            attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
+          />
+          {items.map((item, index) => (
+            <Marker position={[item.value.lat, item.value.lng]} key={index}>
+              <Popup>
+                <b>{item.value.title}</b> <br />
```
roosnic1, line 58: i guess the title and the description should be paragraphs `<p>text</p>`

packages/editor/src/client/blocks/types.ts
```diff
@@ -257,13 +258,16 @@ export function unionMapForBlock(block: BlockValue): BlockInput {
...
-            imageID: image?.id
-          }))
+          items: block.value.items.map(
+            ({value: {address, lat, lng, title, description, image}}) => ({
+              address: address,
+              lat: lat,
+              lng: lng,
+              title,
+              description: description,
+              imageID: image?.id
```
roosnic1, line 268: consistency is key.

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+export function MapLeafletEditPanel({onClose, initialItems}: MapLeafletEditPanel) {
+  const defaultItemValue = {
+    address: 'Brauerstrasse 42, Zürich',
+    lat: 47.3778762,
+    lng: 8.5271078,
+    title: '',
+    description: '',
+    image: null
+  }
+  const [items, setItems] = useState<ListValue<MapLeafletItem>[]>(() =>
```
roosnic1, line 41: why the function?

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+    initialItems.map(item =>
+      item.value.address === ''
```
roosnic1, line 43: why this check?

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+function AddressInputList({
+  isOpen,
+  inputValue,
+  highlightedIndex,
+  getItemProps,
+  getMenuProps
+}: AutocompleteInputListProps) {
+  const [items, setItems] = useState<MarkerPoint[]>([])
+  const api = 'ab48a007c6dd4906adcd2871a5deaebe'
```
roosnic1, line 169: never commit api keys to the repository. The belong to the `.env`

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const fetchURL = `https://api.opencagedata.com/geocode/v1/json?key=${api}&q=${inputValue}&limit=5&pretty=1`
+
+  useEffect(() => {
+    async function fetchData() {
+      if (inputValue !== null && inputValue.length > 1) {
+        try {
+          const request = await axios.get(fetchURL)
```
roosnic1, line 176: this logic should be in `utility.ts`

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
```diff
@@ -0,0 +1,213 @@
...
+          const items = request.data.results.map((res: any) => {
+            return {
+              lat: res.geometry.lat,
+              lng: res.geometry.lng,
+              address: res.formatted
+            }
+          })
+          setItems(items)
+        } catch (error) {
+          console.error(error)
```
roosnic1, line 186: That's not what you do with errors. Either display a message to the user or handle it.

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
roosnic1: And why did you use axios? The native fetch API can do the same.

roosnic1 (changes requested):
I changed your base branch to `f/WEP-77-map-leaflet` because that's where you want those changes.

packages/editor/src/client/blocks/mapLeafletBlock.tsx
0815pascal: In my opinion no: if you create an instance of the mapleaflet you have automatically a default center and zoom. But a question would be if this check would then be necessary at all? But then again it's kind of a absicherung against eventual occurring errors.

packages/editor/src/client/blocks/mapLeafletBlock.tsx
0815pascal: The default height (if I don't set anything) of the map-leaflet is rather narrow and not very convenient for the user. I found that 45vh is a good compromise between having enough space but not filling out too much of the screen. I have chosen vh since it adapts to the viewport of the user's screen and is thus independent of size and resolution.

packages/editor/src/client/blocks/mapLeafletBlock.tsx
0815pascal: done

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
0815pascal: Done. 
I did use axios because I did run into troubles with the native fetch API and with axios it did work. Since I wanted to move forward and since axios seems to have a couple of [advantages over the native fetch API](https://medium.com/@jeffrey.allen.lewis/http-requests-compared-why-axios-is-better-than-node-fetch-more-secure-can-handle-errors-better-39fde869a4a6) I stuck with it.

.gitignore
0815pascal: I think I found [how you do it](https://medium.com/@peter_graham/how-to-create-a-local-gitignore-1b19f083492b)?

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
0815pascal: We made this workaround to avoid the "empty-chip-problem" that we had when we would initialise a markerpoint. We had a discussion about this and decided to add for the moment a default address (Brauerstrasse) until we switch the UI-framework since the problem is related to the karma.run-UI.

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
0815pascal: this logic basically checks if we're creating a new markerpoint (then use the default item values) or if we're dealing with a marker-point that has already been set. This answer is also tied to my previous one (avoiding the "empty-chip-problem).

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
0815pascal: Like [that](https://gist.github.com/fgilio/230ccd514e9381fafa51608fcf137253) for example?

packages/editor/src/client/panel/mapLeafletEditPanel.tsx
0815pascal: Done

.gitignore
roosnic1: 👍🏾

packages/editor/src/client/blocks/mapLeafletBlock.tsx
0815pascal: Done

PR #47, opened 2020-09-19
WEP-86 properties field for articles and pages

roosnic1:


.github/workflows/test.yml
```diff
@@ -10,9 +10,15 @@ jobs:
...
+        with:
+          mongodb-version: '4.0'
       - name: Install deps
         run: yarn install
       - name: Build everything
         run: yarn build:production
       - name: Run tests
         run: yarn test
+        env:
+          TEST_MONGO_URL: 'mongodb://localhost:27017/wepublish_test'
```
dev7ch, line 24: Suggesting to move the up, befoer `steps:`

(btw why Test: is starting with capitals above?)

dev7ch: ... further i am unsure if this is correct here, should this not be in the docker file at `/api`?

packages/api/__tests__/api/private/image.graphql
```diff
@@ -0,0 +1,84 @@
...
+  title
+  description
+  width
+  height
+  ...ImageURLs
+}
+
+fragment FullImage on Image {
+  id
+
```
dev7ch, line 25: Some Fragments have a blank line after id, some not.
is this not handled or cause by linter?

packages/api/__tests__/api/public/image.graphql
```diff
@@ -0,0 +1,48 @@
...
+  title
+  description
+  width
+  height
+  ...ImageURLs
+}
+
+fragment FullImage on Image {
+  id
+
```
dev7ch, line 25: same as above, blank line inconsistence

packages/api/codegen.yml
```diff
@@ -0,0 +1,35 @@
+#schema: ${API_URL}/admin
```
dev7ch, line 1: is this needed, or a case when this comes handy?

Suggesting comment inside the comment if can be handy, if not removal)

packages/api/__tests__/utility.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export async function createGraphQLTestClientWithMongoDB(): Promise<TestClient> {
+  if (!process.env.TEST_MONGO_URL) {
+    throw new Error('TEST_MONGO_URL not defined')
+  }
+  let adminUser
+
+  await MongoDBAdapter.initialize({
+    url: process.env.TEST_MONGO_URL!,
```
dev7ch, line 45: what does the `!` at the end of `url: process.env.TEST_MONGO_URL!`

packages/api/__tests__/utility.ts
```diff
@@ -0,0 +1,129 @@
...
+          roleIDs: [adminUserRoleId]
+        },
+        password: '123'
+      })
+    }
+  })
+
+  const dbAdapter = await MongoDBAdapter.connect({
+    url: process.env.TEST_MONGO_URL!,
+    locale: 'en'
```
dev7ch, line 64: here to, i dont  get, what the `!` does

packages/api/__tests__/api/public/image.graphql
dev7ch: if possible, i d suggest to drop the blank line, or if generated with code gen a manually lint could help

.github/workflows/test.yml
roosnic1: @dev7ch this is how github-actions work. And no the `TEST_MONGO_URL` is not needed in the docker-compose.yaml since that file is not need anywhere but local dev environment.

packages/api/__tests__/api/private/image.graphql
roosnic1: *.graphql files are not yet linted. https://github.com/wepublish/wepublish/issues/54

packages/api/codegen.yml
roosnic1: i removed the line. Thanks

packages/api/__tests__/utility.ts
roosnic1: https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

.github/workflows/test.yml
dev7ch: @roosnic1 yes, thats correct standard way so set env in the jobs (as they are independent), maybe i was not clear enough, my idea/sugesstion was to set the variable on the workflow level, if possible, instead of job level to reuse the var in upcoming tests:

- For readability
- For having only one to define

https://github.community/t/how-to-set-and-access-a-workflow-variable/17335

packages/api/__tests__/utility.ts
dev7ch: thanks

.github/workflows/test.yml
roosnic1: Good point but so far we only need the `TEST_MONGO_URL` var for the tests with mongodb. And the `yarn test` command starts all the tests anyway. But this can/will change as the projects grows. As an old friend of mine always said "Premature optimization is the death of any project" 😄.

dev7ch: 😄

0815pascal (approved):

PR #48, opened 2020-09-23
Await mongoClient calls to avoid memory leaks

roosnic1:
This PR fixes some missing awaits on async calls form mongoClient. Additionally it also closes the additional database connection which gets created during a migration.

### DIFF
--- a/packages/api-db-mongodb/src/migration.ts
+++ b/packages/api-db-mongodb/src/migration.ts
@@ -107,7 +107,7 @@ export const Migrations: Migration[] = [
 
       await userRoles.createIndex({name: 1}, {unique: true})
 
-      userRoles.insertMany([
+      await userRoles.insertMany([
         {
           _id: 'admin',
           createdAt: new Date(),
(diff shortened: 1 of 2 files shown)


packages/api-db-mongodb/src/migration.ts
```diff
@@ -107,7 +107,7 @@ export const Migrations: Migration[] = [
 
       await userRoles.createIndex({name: 1}, {unique: true})
 
-      userRoles.insertMany([
+      await userRoles.insertMany([
```
0815pascal, line 110: Just that I understand it correctly: setting an await here prevents the memory leakage from occuring?

0815pascal (approved):

packages/api-db-mongodb/src/migration.ts
dev7ch: here you are inside a `async` context `async dbMigrate` here, await prefix should be added basically to all functions to ensure, function execution is concatinated correctly.

in general i dont understand why sometime its 
`const migrations = db.collection(CollectionName.Migrations)` or `const migrations = await db.collection(CollectionName.Migrations)`

Just

roosnic1: @0815pascal Not really! Because the method `insertMany` is asynchronous the `await` is needed. The memory leakage happened because the connection to the database during a migration never gets closed.

PR #51, opened 2020-09-26
Multilanguage support for Editor

roosnic1:


packages/editor/src/client/base.tsx
```diff
@@ -36,20 +38,22 @@ export interface BaseProps {
...
+  const {t} = useTranslation()
+
   return (
     <NavigationTemplate
       navigationChildren={
         <>
           <RouteMenuLinkButton
             icon={MaterialIconTextFields}
-            label="Articles"
+            label={t('Articles')}
```
0815pascal, line 49: if I remember it correctly these are the markers-tags for translations, right?

0815pascal (approved):

packages/editor/src/client/blocks/quoteBlock.tsx
```diff
@@ -18,7 +21,7 @@ export function QuoteBlock({value, onChange, autofocus, disabled}: QuoteBlockPro
       <TypographicTextArea
         ref={focusRef}
         variant="h1"
-        placeholder="Quote"
+        placeholder={t('Author')}
```
roosnic1, line 24: @0815pascal careful

packages/editor/src/client/login.tsx
```diff
@@ -147,7 +151,7 @@ export function Login() {
                   flexDirection="column"
                   marginTop={Spacing.Small}
                   alignItems="center">
-                  <p>Or Login with:</p>
+                  <p>{t('Or login with')}:</p>
```
roosnic1, line 154: @0815pascal why did you leave the `:`?

0815pascal: I was thinking that the `:` actually is not something that needs translation; it could also happen that in the process of translation someone deletes it by accident, so I decided to go with that variant.

PR #57, opened 2020-10-07
F/wep 78 editor multi language switch

0815pascal:


packages/editor/src/client/utility.ts
```diff
@@ -127,3 +127,17 @@ export function transformCssStringToObject(styleCustom: string): Record<string,
     return previousValue
   }, {})
 }
+
+export function useStickyState(
```
roosnic1, line 131: Don't put this into `utility.ts` it's used for saving current language to localStorage so you should put it where you need it.

roosnic1 (changes requested):

packages/editor/src/client/languageSwitch.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export function LanguageSwitch() {
+  const css = useStyle()
+  const [uiLanguage, setUILanguage] = useStickyState(
+    {id: 'en', lang: 'en_US', name: 'English'},
+    'savedValues'
+  )
+  const changeLanguage = (lng: string) => {
+    i18n.changeLanguage(lng)
```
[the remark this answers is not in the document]
0815pascal: Are you sure about this? In the documentation it says: 

> While most of the time you only need the t function to translate your content, you can also get the i18n instance (in order to change the language).
> 
> `i18n.changeLanguage('en-US');`

Or am I misunderstand there something?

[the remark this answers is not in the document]
roosnic1: In your code you import the `i18n` object from the file. But you should get it from the hook as described in the documentation.

> `const { t, i18n } = useTranslation();`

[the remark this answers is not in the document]
0815pascal: I'm not really understanding at what you're hinting at. If I use `const [t, i18n] = useTranslation() `within the LanguageSwitch-component and remove the import-statement for `i18n` it causes the app to break after I switch the language. 

Maybe there's something wrong in the code I'm using though: 

```
  const [t, i18n] = useTranslation()

  const changeLanguage = (lng: string) => {
    i18n.changeLanguage(lng)
  }
```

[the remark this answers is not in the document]
roosnic1: What is the error?

0815pascal:
A blank window. In the console I get a warning before the actual crash saying: 

> react_devtools_backend.js:2273 react-i18next:: You will need to pass in an i18next instance by using initReactI18next 

And then when I'm switching the language and the crash occurs: 

> Uncaught TypeError: i18n.changeLanguage is not a function

packages/editor/src/client/languageSwitch.tsx
```diff
@@ -0,0 +1,110 @@
...
+            changeLanguage(values.id)
+          }
+        }}
+      />
+    </div>
+  )
+}
+
+export function LanguageSwitch() {
+  const currentLanguage = JSON.parse(localStorage.getItem('savedValues') || '{}')
```
roosnic1, line 74: I think you're supposed to use the stickyState here.

roosnic1 (changes requested):
Good job but somethings aren't best practices.

packages/editor/src/client/languageSwitch.tsx
```diff
@@ -0,0 +1,110 @@
...
+  )
+
+  const [, i18n] = useTranslation()
+
+  const changeLanguage = (lng: string) => {
+    i18n.changeLanguage(lng)
+  }
+
+  const css = useStyle()
+  const languageSelector = cssRule({
```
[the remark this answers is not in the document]
0815pascal: It prevents the UI element from breaking (see screenshot)
Screenshot-20201022095934-454x200

packages/editor/src/client/languageSwitch.tsx
0815pascal: Actually it came to my mind that I don't need this at all; I just use a key from the translation-file that displays the current language. Otherwise I had difficulties to make this work with stickyState and felt that it would complicate things unnecessarily.

roosnic1: 👍🏾

packages/editor/src/client/languageSwitch.tsx
```diff
@@ -32,17 +32,18 @@ export function useStickyState(
...
+  AVAILABLE_LANG = [
     {id: 'en', lang: 'en_US', name: 'English'},
-    'savedValues'
-  )
+    {id: 'fr', lang: 'fr_FR', name: 'Français'},
+    {id: 'de', lang: 'de_CH', name: 'Deutsch'}
+  ]
 
-  const [, i18n] = useTranslation()
+  const [uiLanguage, setUILanguage] = useStickyState(AVAILABLE_LANG[0], 'localStorageValues')
```
roosnic1, line 44: you are still saving the whole object in the localstorage. And the name is still not very descriptive.

packages/editor/src/client/languageSwitch.tsx
```diff
@@ -32,17 +32,18 @@ export function useStickyState(
   return [value, setValue]
 }
 
+let AVAILABLE_LANG: {id: string; lang: string; name: string}[] = []
```
roosnic1, line 35: Why declaring it here and not also filling it with the data?

roosnic1 (changes requested):

roosnic1 (dismissed):
👍🏾

roosnic1 (dismissed):
👍🏾

roosnic1 (approved):

penk85 (approved):

faebebin (approved):
Looks good. Works all fine locally too.

PR #59, opened 2020-10-09
Page break extend

dev7ch:
Extend PageBreak Block

https://wepublish.atlassian.net/jira/software/projects/BAJ/boards/3?selectedIssue=BAJ-84

- add image field
- add option fields for styling
- add link target option
image

### Todo

- [x] Clearify if we will get a `next` release to for collaborative development or not => no
- [x] how to handle with not MVP feedback? => not clear for Nico
- [ ] Define step to merge Feature extensions fast (max 8h after working PR)

This PR show as fundamental problem of the collaboration and development:

1.) This Pr was never tested in Orbit app (Bajour) and is a best guess depending and the defined Storie in Sprint
2.) Its Mandatory, that the first focus is on functionality only, because used FE UI is not useable in most cases)
3.) **Merge to master and directly release!** 
4.) After Merge and successfully implementation in orbit application, a second PR with basic Style clean up will be done (if needed, should be tracked in WEP Board)

### DIFF
--- a/packages/api/__tests__/api/private/article.graphql
+++ b/packages/api/__tests__/api/private/article.graphql
@@ -25,7 +25,7 @@ fragment MutationArticle on Article {
     }
 }
 
-fragment ArticleRef on Article {
+fragment  ArticleRef on Article {
     id
 
     createdAt
--- a/packages/api/__tests__/api/private/article.graphql
+++ b/packages/api/__tests__/api/private/article.graphql
@@ -25,7 +25,7 @@ fragment MutationArticle on Article {
     }
 }
 
-fragment  ArticleRef on Article {
+fragment ArticleRef on Article {
     id
 
     createdAt
--- a/packages/api/src/graphql/blocks.ts
+++ b/packages/api/src/graphql/blocks.ts
@@ -442,9 +442,13 @@ export const GraphQLLinkPageBreakBlock = new GraphQLObjectType<LinkPageBreakBloc
   name: 'LinkPageBreakBlock',
   fields: {
     text: {type: GraphQLString},
+    richText: {type: GraphQLString},
     linkURL: {type: GraphQLString},
     linkText: {type: GraphQLString},
-    styleOption: {type: GraphQLString}
+    linkTarget: {type: GraphQLString},
+    styleOption: {type: GraphQLString},
+    layoutOption: {type: GraphQLString},
+    imageID: {type: GraphQLID}
   },
   isTypeOf: createProxyingIsTypeOf(value => {
     return value.type === BlockType.LinkPageBreak
--- a/packages/api/src/graphql/blocks.ts
+++ b/packages/api/src/graphql/blocks.ts
@@ -443,7 +443,8 @@ export const GraphQLLinkPageBreakBlock = new GraphQLObjectType<LinkPageBreakBloc
   fields: {
     text: {type: GraphQLString},
     linkURL: {type: GraphQLString},
-    linkText: {type: GraphQLString}
+    linkText: {type: GraphQLString},
+    styleOption: {type: GraphQLString}
   },
   isTypeOf: createProxyingIsTypeOf(value => {
     return value.type === BlockType.LinkPageBreak
@@ -541,7 +542,8 @@ export const GraphQLLinkPageBreakBlockInput = new GraphQLInputObjectType({
   fields: {
     text: {type: GraphQLString},
     linkURL: {type: GraphQLString},
-    linkText: {type: GraphQLString}
+    linkText: {type: GraphQLString},
+    styleOption: {type: GraphQLString}
   }
 })
 
--- a/packages/api/src/graphql/blocks.ts
+++ b/packages/api/src/graphql/blocks.ts
@@ -442,12 +442,13 @@ export const GraphQLLinkPageBreakBlock = new GraphQLObjectType<LinkPageBreakBloc
   name: 'LinkPageBreakBlock',
   fields: {
     text: {type: GraphQLString},
-    richText: {type: GraphQLString},
+    htmlText: {type: GraphQLString},
     linkURL: {type: GraphQLString},
     linkText: {type: GraphQLString},
     linkTarget: {type: GraphQLString},
     styleOption: {type: GraphQLString},
     layoutOption: {type: GraphQLString},
+    image: {type: GraphQLString},
     imageID: {type: GraphQLID}
   },
   isTypeOf: createProxyingIsTypeOf(value => {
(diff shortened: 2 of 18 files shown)


packages/api/__tests__/api/private/article.graphql
```diff
@@ -25,7 +25,7 @@ fragment MutationArticle on Article {
     }
 }
 
-fragment ArticleRef on Article {
+fragment  ArticleRef on Article {
```
roosnic1, line 28: why though?

packages/api/src/graphql/blocks.ts
```diff
@@ -442,8 +442,18 @@ export const GraphQLLinkPageBreakBlock = new GraphQLObjectType<LinkPageBreakBloc
   name: 'LinkPageBreakBlock',
   fields: {
     text: {type: GraphQLString},
+    htmlText: {type: GraphQLString},
```
roosnic1, line 445: We don't save pure HTML in the database. Please use RichTextNode[] => https://github.com/wepublish/wepublish/blob/master/packages/api/src/graphql/richText.ts

In the editor you can then also use the richTextEditor for formatting.

roosnic1: also maybe a different name for the field since it's not clear what's the difference between `text` and `htmlText`

packages/api/src/graphql/blocks.ts
```diff
@@ -442,8 +442,18 @@ export const GraphQLLinkPageBreakBlock = new GraphQLObjectType<LinkPageBreakBloc
...
     linkURL: {type: GraphQLString},
-    linkText: {type: GraphQLString}
+    linkText: {type: GraphQLString},
+    linkTarget: {type: GraphQLString},
```
roosnic1, line 448: Either make it boolean for open in same page or new page. Or do it properly and only accept one of the four options.
[_blank, _self, _parent, _top]

packages/editor/src/client/blocks/linkPageBreakBlock.tsx
```diff
@@ -11,44 +32,174 @@ export function LinkPageBreakBlock({
...
+          height={150}
+          flexGrow={1}
+          marginRight={Spacing.ExtraSmall}>
+          <Box padding={'10'}>
+            <p>Colors and Styles</p>
+            <small>Styles: </small>
+            <select
+              defaultValue={styleOption}
+              onChange={e => onChange({...value, styleOption: e.target.value || ''})}>
+              <option value="default">Default Color</option>
```
roosnic1, line 104: Not happy with that. How should other we.publish instance user expand this.

One option is to add a settings page in the editor where you can add your own Styles with label, Name and description or you change this field to a textinput and just pass a css class name. As i mentioned in the stand up we'll implement a "extend with styles" solution for all blocks soon but having fixed values in the editor is not a suitable option.

packages/editor/src/client/blocks/linkPageBreakBlock.tsx
```diff
@@ -11,44 +32,174 @@ export function LinkPageBreakBlock({
...
+              <option value="light">Light Color Style</option>
+              <option value="dark">Dark Color Style</option>
+            </select>
+          </Box>
+          <Box padding={'10'}>
+            <small>Layouts: </small>
+            <select
+              defaultValue={layoutOption}
+              onChange={e => onChange({...value, layoutOption: e.target.value || ''})}>
+              <option value="default">Default Layout</option>
```
roosnic1, line 114: A little bit the same issue as with the field above. Also i think you copy/pasted a bit too much.

If you want to have that field as a left, center or right aligned field then you should change the type in the API to only allow those values and then present here a Radio Group.

roosnic1 (changes requested):
I left some comments that need changes.

But for a general Feedback. The Page break block as you implemented it doesn't really support the "What you See Is Almost What you Get" approach in the editor. I highly suggest that you move the settings into a separate panel on the side and visualize the placement of the image and text as it would be in the frontpage.

packages/api/src/graphql/blocks.ts
dev7ch: - `text` is plaintext and needs to be untached (backwards compat)
- `htmlText` => is text which can container html (RichText is not matching case, must be saved as pure HTML in DB)

packages/editor/src/client/blocks/linkPageBreakBlock.tsx
dev7ch: MVP Approach, this is the suitable way for now as, this was cleary defined in offer before starting. (MVP Approach)
.we need a release of api to start local development on Bajout FE side ... (rather considering to make a `next` release wich is release at least every day (and fix the meniotoned stuff by you before new versioned is published release)

packages/editor/src/client/blocks/linkPageBreakBlock.tsx
dev7ch: MVP,  out of scope.

This good for now, there will be a final cleanup after merge of this PR => as said, this needs to be merge, that we can start 
to develop.

packages/api/src/graphql/blocks.ts
dev7ch: in Short, rich Text us not suitable.

packages/api/__tests__/api/private/article.graphql
dev7ch: upps

dev7ch:
>The Page break block as you implemented it doesn't really support the "What you See Is Almost What you Get" approach in the editor.

As we spoke before start, MVP Approch in FE ... this means only functional review and comments a valid here.

All other stuff should be treated as a new task at webpublish dev board (Again, we need a merge to start develeopment)

packages/api/src/graphql/blocks.ts
dev7ch: Good point, not at the moment ([_blank, _self, _parent, _top]), but the idea is to keep this as string to be flexible if we need some other values here, not 100% sure yet which values will be stored here

packages/editor/src/client/blocks/linkPageBreakBlock.tsx
dev7ch: MPV, ok for now, can be solved on clean up PR after FE Bajour implementation is done

packages/api/src/graphql/blocks.ts
dev7ch: (richText dont store pure HTML, we need pure HTML)

dev7ch:
Further i d like to had, that i still have not a flawless running locla dev setup, whichs means in general.

- I really the to beat me docker to get MongDB runnnig correclty with correct data (volumes removal, rebuilds etc needed)
- The media server locally works only sometimes in FF (dont work in Chrome for me)
- Mirko was not able to validate the workarounds hack until now, mentioned here https://wepublish.atlassian.net/browse/BAJ-89

PR #69, opened 2020-11-10
F/wep 89 editor rich text options

faebebin:
Add buttons for inserting text (emojis and special characters), and for changing text format to sub/superscript.


packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -320,6 +334,17 @@ export const RichTextBlock = memo(function RichTextBlock({
 
         <LinkFormatButton />
         <RemoveLinkFormatButton />
+
+        <ToolbarDivider />
+
+        {'😄 😁 🤯 🎸 ¿ å etc'.split(' ').map((txt, i) => (
```
roosnic1, line 340: For Emojis: https://github.com/missive/emoji-mart

It looks like there is no "special characters" picker. I suggest we make this into an emoji only picker.

roosnic1 (commented):
Looks good. Once you implemented the emoij picker we're good to go 👍🏾

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Ok, will integrate emoji picker and make an emoji-only picker.

packages/editor/src/client/blocks/emoji-mart-override.css
```diff
@@ -0,0 +1,12 @@
+ .emoji-mart-bar:last-child {
```
roosnic1, line 1: I think the best thing is to name the CSS file the same as the component it belongs to. I suggest you move the emoij-mart-picker to its own component and call the css file the same. 

Also make the css file a less file and import the less file from the component inside that less file.

roosnic1 (changes requested):
looks good. just some tiny clean up

packages/editor/src/client/blocks/emoji-mart-override.css
faebebin: That sounds good, will implement.

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -304,6 +327,62 @@ function FormatButton({icon, format}: SlateBlockButtonProps) {
   )
 }
 
+interface WhisperInstancePopover extends React.Component<WhisperProps, {isOverlayShown?: boolean}> {
```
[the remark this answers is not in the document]
faebebin: I needed a ref with the open() and close() methods. But happy to workout the correct/clean way.

roosnic1 (approved):
👍🏾

packages/editor/src/client/atoms/emojiButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+  readonly iconActive?: IconNames | SVGIcon
+  readonly active?: boolean
+}
+
+export const BaseEmojiButton = forwardRef<HTMLButtonElement, EmojiButtonProps>(
+  function BaseEmojiButton({icon, iconActive, active, ...props}, ref) {
+    icon = iconActive && active ? iconActive : icon
+    return (
+      <button
+        style={{
```
0815pascal, line 22: Is there also a possibility to style this within a css-file instead uf styling it directly in the code?

packages/editor/src/client/atoms/emojiButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+export function EmojiButton({icon, iconActive}: EmojiButtonProps) {
+  const editor = useSlate()
+  const [isEmojiPickerOpen, setIsEmojiPickerOpen] = useState(false)
+
+  const triggerRef = useRef<any>(null)
+
+  const open = () => {
+    if (triggerRef.current) {
+      triggerRef.current!.open()
+      // TODO rather be done with forwardRef ?
```
0815pascal, line 49: Is this TODO still open?

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -71,7 +74,9 @@ const TextTags: any = {
   I: () => ({[TextFormat.Italic]: true}),
```
0815pascal, line 74: Some people also use `<em>` as a way to style italics, similarly to `<b>` and `<strong>`. So in order to stay coherent (you have `<b>` and `<strong>`) I would also include `<em>`.

0815pascal (changes requested):

packages/editor/src/client/atoms/emojiButton.tsx
roosnic1: We want to avoid css files as much as possible. @faebebin already opened a jira issue that addresses this problem.

packages/editor/src/client/atoms/emojiButton.tsx
faebebin: Yes, I think it would be a bit cleaner with forwarding the ref to the BaseEmojiButton and using ref.current.isOpen (or similar) instead of using the isEmojiPickerOpen state variable to have the DOM element of single source of truth here. So maybe we could make a github <good first issue> out of it?

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: ok, thanks, will implement.

penk85 (approved):

packages/editor/src/client/atoms/emojiButton.tsx
faebebin: This is actually something we might have to discuss  regarding conventions. It could easily be moved to the already existing .lee file. But the .less file here is only for overriding npm-package stylesheet. The other /atoms components use inline css mostly. So where should /atoms/components style live?

roosnic1: @faebebin leave it as it is for the moment. We need to clean up the styles anyway and then we can implement a proper style solution that suites all our usecases.

packages/editor/src/client/atoms/emojiButton.tsx
roosnic1: @faebebin yes create a github issue out of this.

packages/editor/src/client/atoms/emojiButton.tsx
faebebin: ok

packages/editor/src/client/atoms/emojiButton.tsx
faebebin: ok

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: italic and strong, both where options I did not actually change anything. But having a look again, they seem to be already quite the way you describe? .. If you see an issue there could you please commit yourself?

0815pascal: ok, I see. No in that case I wouldn't change anything and leave the code as it is :)

0815pascal (approved):

PR #83, opened 2020-11-24
Use E2E Tests for proper integration testing

roosnic1:
Needs to be updated with master and then added to the CI/CD


camilleory (approved):

penk85 (approved):

packages/api/src/server.ts
```diff
@@ -95,7 +95,8 @@ export class WepublishServer {
     })
 
     const corsOptions = {
-      origin: '*',
+      origin: true,
```
bishoymelek, line 98: what does these do?

packages/editor/webpack.config.ts
```diff
@@ -6,8 +6,8 @@ import {AssetListPlugin} from '@karma.run/webpack'
 import ForkTsCheckerWebpackPlugin from 'fork-ts-checker-webpack-plugin'
 import {CleanWebpackPlugin} from 'clean-webpack-plugin'
 
-export default (env: any, {mode}: any) =>
-  ({
+export default (env: any, {mode}: any) => {
```
bishoymelek, line 9: do you want to get it back?

tests/e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+import * as process from "process";
+
+const loginName = Selector('input.username')
+const loginPassword = Selector('input.password')
+const createArticle = Selector('a').withAttribute('href', '/article/create')
+const metadataButton = Selector('button').child('i.rs-icon-newspaper-o')
+const createButton = Selector('button').child('i.rs-icon-save')
+const publishButton = Selector('button').child('i.rs-icon-cloud-upload')
+
+const closeButton = Selector('.rs-drawer-footer').child('button.rs-btn-primary')
```
bishoymelek, line 12: can we have it like `drawerCloseButton`? to differ it from modals for example?

packages/api/src/server.ts
roosnic1: https://github.com/expressjs/cors#cors

packages/editor/src/client/app.tsx
```diff
@@ -139,6 +141,18 @@ export function App() {
     setLang(currentLanguageMap.get(lng))
   })
 
+  const {session} = useContext(AuthContext)
+  const dispatch = useRouteDispatch()
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
roosnic1: We also check in the router but this is the place where we watch the `session` and the session is checked once you focus the browser tab with the editor again.

packages/editor/src/client/route.tsx
```diff
@@ -250,7 +250,7 @@ export function RouteProvider({children}: RouteProviderProps) {
           } else {
             dispatch({
               type: RouteActionType.SetCurrentRoute,
-              route: next
+              route: {...next, hash: ''}
```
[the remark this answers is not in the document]
roosnic1: It's used to clear any URL hash but it's not needed. I removed it.

packages/editor/webpack.config.ts
roosnic1: yes

tests/e2e.ts
roosnic1: This is only a test implementation. @camilleory has in here tests a better naming strategy.

bishoymelek (approved):
🎉

camilleory (approved):

PR #85, opened 2020-11-24
F/wep 90 navigation menu display

0815pascal:


examples/website/src/shared/navigation/navigationBar.tsx
```diff
@@ -261,7 +261,9 @@ export function NavigationBar({itemsCategory, itemsIntern}: NavigationBarProps)
                 <div key={index}>
                   <Link
                     className={css(LinkStyles)}
+                    target={'_blank'}
```
roosnic1, line 264: `target="_blank"` only when href.

examples/website/src/shared/route/baseTemplateContainer.tsx
```diff
@@ -37,6 +49,15 @@ export interface BaseTemplateContainerProps {
   children?: ReactNode
 }
 
+export const {
```
roosnic1, line 52: What is this?

roosnic1 (changes requested):

examples/website/src/shared/navigation/navigationBar.tsx
0815pascal: I made a distinction now between an internal link which has `target="_self"` and an external link which has `target="_blank". Can I leave it with that or should I really only use `target` when we're dealing with an external link?

roosnic1: 👍🏾  `target="self"` is bueno

examples/website/src/shared/route/baseTemplateContainer.tsx
```diff
@@ -64,16 +74,35 @@ export function BaseTemplateContainer({
...
-  return {
-    title: link.label,
-    route: PageRoute.create({slug: link.page?.slug ?? undefined}),
-    isActive: false
+  switch (link.__typename) {
+    case 'PageNavigationLink':
+      return {
+        title: link.label,
+        route: PageRoute.create({
+          slug: link.page?.slug ?? undefined
```
roosnic1, line 83: Here is an example where you should use `!`. TypeScript doesn't know that when `link.__typename` is of type `PageNavigationLink` we have a page for sure. But we developers do. You can rewrite this line to `slug: link.page!.slug`.

examples/website/src/shared/route/baseTemplateContainer.tsx
```diff
@@ -64,16 +74,35 @@ export function BaseTemplateContainer({
...
+        }),
+        isActive: false
+      }
+    case 'ArticleNavigationLink':
+      return {
+        title: link.label,
+        route: ArticleRoute.create({
+          id: link.article?.id ?? undefined
```
roosnic1, line 91: same as above.

roosnic1 (changes requested):
Looks good. One last thing i forgot in the first review.

examples/website/src/shared/route/baseTemplateContainer.tsx
```diff
@@ -80,15 +80,15 @@ function linkToNavigationItem(link: any): NavigationItem {
       return {
         title: link.label,
         route: PageRoute.create({
-          slug: link.page?.slug ?? undefined
+          slug: link.page!.slug ?? undefined
```
roosnic1, line 83: you don't need `?? undefined` when you use `!`

examples/website/src/shared/route/baseTemplateContainer.tsx
```diff
@@ -80,15 +80,15 @@ function linkToNavigationItem(link: any): NavigationItem {
...
         }),
         isActive: false
       }
     case 'ArticleNavigationLink':
       return {
         title: link.label,
         route: ArticleRoute.create({
-          id: link.article?.id ?? undefined
+          id: link.article!.id ?? undefined
```
roosnic1, line 91: you don't need `?? undefined` when you use `!`

roosnic1 (changes requested):

roosnic1 (approved):

faebebin (approved):
Looks good.

PR #88, opened 2020-11-25
F/wep 89 editor rich text table

faebebin:


faebebin:
Currently unsolved: cannot insert whitespace when writing text inside table cell.

packages/editor/src/client/blocks/richTextBlockTable.less
```diff
@@ -0,0 +1,16 @@
+table {
+  white-space: pre-wrap;
+  width: 100%;
+  margin: 10px;
```
0815pascal, line 4: I'm not sure if `px` is the best unit for `margin` and `padding` since it can look ugly with different screen resolutions. `em` and `rem` would be alternatives. What do you think @roosnic1 ?

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -142,6 +180,19 @@ function renderElement({attributes, children, element}: RenderElementProps) {
     case BlockFormat.ListItem:
       return <li {...attributes}>{children}</li>
 
+    case BlockFormat.Table:
+      return (
+        <table>
+          <tbody {...attributes}>{children}</tbody>
```
0815pascal, line 186: Was it a conscious decision to use only `tbody` and to leave out `thead` and `tfoot` ?

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+}
+
+export const BaseTableButton = forwardRef<HTMLButtonElement, TableButtonProps>(
+  function BaseTableButton({icon, iconActive, active, ...props}, ref) {
+    icon = iconActive && active ? iconActive : icon
+    return (
+      <button
+        style={{
+          border: active ? 'blue 1px solid' : '',
+          fontSize: 16,
```
0815pascal, line 502: Don't you need a unit here? If default is `px` I wonder if this is the right way to do it since on high-res screens `px` are not being scaled

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+        }}>
+        <SetRowColNumbers key={'row'} label="rows" num={nrows} setNumber={setNrows} />
+        <SetRowColNumbers key={'col'} label="columns" num={ncols} setNumber={setNcols} />
+        <Button
+          disabled={isFormatActive(editor, BlockFormat.Table)}
+          onClick={() => {
+            Transforms.insertNodes(editor, emptyCellsTable(nrows, ncols))
+            close()
+          }}>
+          {t('insert table')}
```
0815pascal, line 560: We have chosen to use a specific way of formatting translations. Compare with `locales/en.json`: it would be something like `{t('blocks.richTextTable.insertTable')}`

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+        </Button>
+      </div>
+    </Popover>
+  )
+
+  const emptyCellsTable = (nrows: number, ncols: number): SlateElement[] => [
+    {
+      type: BlockFormat.Table,
+      children: Array.from({length: nrows}).map(_ => ({
```
0815pascal, line 569: What's that `_` being used for?

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+            e.preventDefault()
+            isTableSpecsOpen ? close() : open()
+          }}
+        />
+      </Whisper>
+    </>
+  )
+}
+
+interface SetRowColNumbersProps {
```
0815pascal, line 605: I think you should set the `interfaces` at the beginning of the file

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -522,6 +710,50 @@ function withRichText<T extends ReactEditor>(editor: T): T {
...
+    const {selection} = editor
+
+    if (selection && Range.isCollapsed(selection)) {
+      const [cell] = Editor.nodes(editor, {
+        match: n =>
+          !Editor.isEditor(n) && SlateElement.isElement(n) && n.type === BlockFormat.TableCell
+      })
+
+      if (cell) {
+        const [, cellPath] = cell
```
0815pascal, line 723: Do you still need that `,` ?

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -522,6 +710,50 @@ function withRichText<T extends ReactEditor>(editor: T): T {
...
```
0815pascal, line 745: Do you still need that `,` ?

0815pascal (changes requested):
Good job, added some open questions and suggestions

packages/editor/src/client/blocks/richTextBlock.tsx
roosnic1: We already have a ticket for that. https://wepublish.atlassian.net/browse/WEP-118

packages/editor/src/client/blocks/richTextBlock.tsx
roosnic1: 👍🏾

packages/editor/src/client/blocks/richTextBlockTable.less
roosnic1: https://wepublish.atlassian.net/browse/WEP-118

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+            {
+              type: BlockFormat.Paragraph,
+              children: [{text: ''}]
+            }
+          ]
+        }))
+      }))
+    },
+    {
+      type: BlockFormat.Paragraph,
```
roosnic1, line 583: Is that empty text on purpose here? If so please leave a comment why.

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+  setNumber: React.Dispatch<React.SetStateAction<number>>
+}
+
+function SetRowColNumbers({label, num, setNumber}: SetRowColNumbersProps) {
+  const editor = useSlate()
+
+  return (
+    <InputGroup style={{width: '140px'}} disabled={isFormatActive(editor, BlockFormat.Table)}>
+      <InputGroup.Addon style={{width: '60px'}}>{label}</InputGroup.Addon>
+      <Input type="number" value={`${num}`} onChange={val => setNumber(Number(val))} />
```
roosnic1, line 617: Why do you make the number to a string? and then make it a number again in the `onChange`? I'm pretty sure rsuite has a number input component.

roosnic1 (changes requested):

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Was intentionally. As the content can be styled inside the cells, I would leave it to the user to style a row differently (eg toprow bold). I thought of the table as a basic (inline-) structuring element. As @roosnic1 told me, it would be used for tsüri.ch as an element to flexibly combine pictures and text. But, I actually don't know whether not having a thead could imply any problems on some browsers?

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Ok, thanks, will correct.

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Placeholder for an unused argument. Would we have some convention on hoe declaring unused arguments?. I actually used Array.fill(...) first, but that caused key errors in the rendered table.

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Can do that. I read somewhere that interfaces should be placed next to their implementation?

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Think so, because it is array destructuring. This indicates that the second item is catched. Actually this code comes directly out of slatejs examples.

packages/editor/src/client/blocks/richTextBlock.tsx
0815pascal: You might be right, I think I have just always encountered the interfaces on the beginning of the file in our codebase. Maybe @roosnic1 can tell us how we do it at wepublish?

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -431,6 +486,139 @@ function RemoveLinkFormatButton() {
...
+      children: [{text: ''}]
+    }
+  ]
+
+  return (
+    <>
+      <Whisper placement="top" speaker={tableSpecs} ref={triggerRef} trigger="none">
+        <BaseTableButton
+          icon={isTableSpecsOpen ? iconActive || icon : icon}
+          disabled={isFormatActive(editor, BlockFormat.Table)}
```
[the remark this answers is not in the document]
faebebin: Good point. For this to be displayed I will have to count the <tr> and <td> of the currently active table. It just turned out to me,  that dynamically using the current slate nodes tree is not too easy. So if really needed, I could implement this?

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Yes is intentionally. I will add a comment. It enables adding breaklines inside <td>. Actually, all content in a tablecell is wrapped in paragraph. So actually the table looks like <table><tbody><tr><td><p>......</p>. This enables qlot of funcionality, especially as tables are used as 'structuring'-elements too.

packages/editor/src/client/blocks/richTextBlock.tsx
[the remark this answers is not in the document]
faebebin: I think that is not useful information to see how many rows and columns a table has. I will likely use the popup to show further table formatting options when the cursor is inside the table.

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: I do not find specific number input. Little confusing, the docs state that value prop would be type number:
image

faebebin: whereas typescript tells, that value is excluseivly string.

faebebin: I do not see a way to do it more elegant.

packages/editor/src/client/blocks/richTextBlock.tsx
faebebin: Got it: Will use () => ... instead

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -190,7 +194,18 @@ function renderElement({attributes, children, element}: RenderElementProps) {
       return <tr {...attributes}>{children}</tr>
 
     case BlockFormat.TableCell:
-      return <td {...attributes}>{children}</td>
+      // TODO custom borderColor using colorPicker
+      return (
+        <td
+          {...attributes}
+          style={{
+            borderColor: (c => (c === 'transparent' ? `rgb(0, 0, 0, 0.1)` : c))(
```
0815pascal, line 202: What does `c` stand for? Maybe try using more clear variable-names since it makes reading your code easier :)

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -726,45 +792,51 @@ function withRichText<T extends ReactEditor>(editor: T): T {
...
+    if (selection) {
       const [cell] = Editor.nodes(editor, {
         match: n =>
           !Editor.isEditor(n) && SlateElement.isElement(n) && n.type === BlockFormat.TableCell
       })
 
       if (cell) {
         const [, cellPath] = cell
-        const start = Editor.start(editor, cellPath)
+        const loc = Editor[location](editor, cellPath)
```
0815pascal, line 809: Can you use a more meaningful variable-name for `loc`?

packages/editor/src/client/blocks/richTextBlockTable.less
```diff
@@ -5,10 +5,8 @@ table {
   table-layout: fixed;
 }
 
-table,
-td,
-tr {
-  border: 1px solid black;
+td {
```
0815pascal, line 8: I guess you can use only one `td`-declaration instead of two here? (I meant you can include the `padding` declaration in here?)

packages/editor/src/client/blocks/richTextBlock.tsx
```diff
@@ -101,7 +101,11 @@ const ElementTags: any = {
   UL: () => ({type: BlockFormat.UnorderedList}),
   TB: () => ({type: BlockFormat.Table}),
   TR: () => ({type: BlockFormat.TableRow}),
-  TD: () => ({type: BlockFormat.TableCell})
+  // TD: () => ({type: BlockFormat.TableCell}),
+  TD: (el: Element) => ({
```
0815pascal, line 105: Try using a more meaningful name for `el`.

0815pascal (changes requested):

packages/editor/src/client/blocks/richTextBlock.tsx
roosnic1: `el` is a known abbreviation

roosnic1 (changes requested):
Looks very good. 👍🏾 Just some tiny changes.

One question we already talked about but might be worth reconsidering. Is there a way to move all the table related stuff to a separate file?

faebebin:
> Looks very good. 👍🏾 Just some tiny changes.
> 
> One question we already talked about but might be worth reconsidering. Is there a way to move all the table related stuff to a separate file?

It would be nice, but the component uses type BlockFormat.Table which is currently defined and used in the current file. So maybe we refactor the richTextBlock.tsx in a next step?

penk85 (approved):

0815pascal (approved):

bishoymelek (approved):

roosnic1 (approved):
👍🏾

PR #89, opened 2020-11-26
F/wep 110 favicon

0815pascal:


penk85 (dismissed):

packages/editor/src/server/index.tsx
```diff
@@ -42,6 +42,8 @@ async function asyncMain() {
             rel="stylesheet"
           />
           <meta name="viewport" content="width=device-width, initial-scale=1" />
+          <link rel="icon" type="image/png" sizes="64x64" href="../../assets/favicon-128.png" />
```
roosnic1, line 45: that `assets` folder is controlled by webpack and should not contain any other files. But you can create a folder called `static` on that level. Please check `examples/website` for reference.

roosnic1 (changes requested):

faebebin (dismissed):
Looks good.

roosnic1 (changes requested):
As discussed in Slack.

packages/editor/src/server/index.tsx
```diff
@@ -32,6 +32,7 @@ async function asyncMain() {
   const app = express()
 
   app.use('/assets', express.static(joinPath(__dirname, '../../assets'), {index: false}))
+  app.use('/static', express.static(path.resolve(__dirname, '../../static/')))
```
roosnic1, line 35: I suggest you use `joinPath` here as well.

roosnic1 (changes requested):

packages/editor/src/server/index.tsx
0815pascal: ok and `{index: false}` as well?

roosnic1: @0815pascal what does `{index: false}` do?

0815pascal: > By default this module will send "index.html" files in response to a request on a directory. To disable this set false or to supply a new index pass a string or an array in preferred order.


Not sure if I understand this correctly. Does it mean, if I try for example to access the `/assets` directory from my browser, that it will create by default an `/assets/index.html` which shows all the content of `asset`-folder?

roosnic1: 👍🏾  so `{index: false}`seems like a good option.

0815pascal: Ok, will do that!

roosnic1 (approved):

faebebin (approved):
Looks goos.

PR #90, opened 2020-11-30
Issue 78 emojipicker ref

faebebin:
I also work on this issue because I came accross the same topic working on the table button. Problem 'update component (ie. emojibutton) upon change of ref state'. Solution found here: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node . For consistency reasons I also implemented it here.


packages/editor/src/client/atoms/emojiButton.tsx
```diff
@@ -67,17 +33,25 @@ export function EmojiButton({icon, iconActive}: EmojiButtonProps) {
...
+            padding: 2
+          }}
           onMouseDown={e => {
             e.preventDefault()
-            isEmojiPickerOpen ? close() : open()
-          }}
-        />
+            !isPopoverOpen ? triggerRef.current!.open() : triggerRef.current!.close()
+          }}>
+          <Icon icon={icon} element={icon} />
```
roosnic1, line 53: do you need the property `element`?

roosnic1 (commented):
looks good. just one question.

packages/editor/src/client/atoms/emojiButton.tsx
faebebin: No :+1:

packages/editor/src/client/atoms/emojiButton.tsx
```diff
@@ -11,54 +11,20 @@ import {Picker, BaseEmoji} from 'emoji-mart'
...
-    }
-  }
-  const close = () => {
-    if (triggerRef.current) {
-      triggerRef.current!.close()
-      setIsEmojiPickerOpen(false)
-    }
-  }
+  const popOverRef = useCallback((node: any) => {
+    setIsPopoverOpen(!!node)
```
0815pascal, line 23: Do the `!!` have a special role? Otherwise `!!node === node` wouldn't it?

0815pascal (approved):
Looks good, just one question

roosnic1 (approved):

packages/editor/src/client/atoms/emojiButton.tsx
faebebin: I actually found this on this repo. It turns any defined object into boolean, which is needed in this case.

0815pascal: Ah ok, I didn't know that - just learned something new :)

Issue #74, opened 2020-11-13
Author-List doesn't update upon removing an author


0815pascal:
**Expected behaviour:** Upon deleting an author, the author-list should automatically update whereby the recently deleted author shouldn't be displayed anymore. 

**Actual behaviour:** Upon deleting an author the recently deleted author is still being displayed in the list. It only disappears upon manually reloading the page. 

![8ab2935a-e824-4c0e-8db2-895829e3a556](https://user-images.githubusercontent.com/11842633/99073778-ce478000-25b6-11eb-9a38-50392c5959d2.gif)

roosnic1:
Fixed

PR #94, opened 2020-12-08
Bug fix - Author list now updates the UI after an author is removed

persunde:
#74 
Fixed the bug where the author list was not automatically updated when an author was deleted.
Seems to be a bug related to Apollo, so the refetch after a delete might not be needed in the future for newer versions of Apollo for React.


roosnic1 (approved):
👍🏾 . Thanks @persunde

penk85 (approved):

persunde:
Sorry. I added another commit to the wrong branch :(

bishoymelek (approved):

0815pascal (approved):

PR #100, opened 2020-12-09
T/api article tests

penk85:
Tests for Article on api

### DIFF
--- a/packages/api/__tests__/api/private/article.graphql
+++ b/packages/api/__tests__/api/private/article.graphql
@@ -30,10 +30,10 @@ fragment MutationArticle on Article {
         breaking
         tags
         properties {
-        key
-        value
-        public
-      }
+            key
+            value
+            public
+        }
     }
 }
 
--- a/packages/api/__tests__/api/private/article.graphql
+++ b/packages/api/__tests__/api/private/article.graphql
@@ -1,5 +1,6 @@
 fragment MutationArticle on Article {
     id
+    shared
 
     draft {
         publishedAt
@@ -19,9 +20,20 @@ fragment MutationArticle on Article {
     }
 
     latest {
-        publishedAt
+       publishedAt
         updatedAt
         revision
+        title
+        preTitle
+        lead
+        slug
+        breaking
+        tags
+        properties {
+        key
+        value
+        public
+      }
     }
 }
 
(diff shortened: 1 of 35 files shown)


bishoymelek (dismissed):
Good job Penina Thanks

0815pascal (approved):
All good :)

packages/api/__tests__/api/private/article.graphql
```diff
@@ -1,5 +1,6 @@
 fragment MutationArticle on Article {
```
roosnic1, line 1: I would change this to more meaningful name like `StandardArticle` or similar.

packages/api/__tests__/api/private/article.graphql
```diff
@@ -22,6 +23,17 @@ fragment MutationArticle on Article {
...
         updatedAt
         revision
+        title
+        preTitle
+        lead
+        slug
+        breaking
+        tags
+        properties {
+        key
```
roosnic1, line 33: indentation

packages/api/__tests__/specs/article.ts
```diff
@@ -22,7 +31,36 @@ beforeAll(async () => {
 })
 
 describe('Articles', () => {
-  describe('can be created/edited/deleted', () => {
+  describe('can be created/edited/deleted:', () => {
+    let articleIds: string[]
+    articleIds = []
```
roosnic1, line 36: you can write this on the same line and use `const` since you only change the content of the array and not the array istelf.
`const articleIDs: string[] = []`

packages/api/__tests__/specs/article.ts
```diff
@@ -55,6 +93,127 @@ describe('Articles', () => {
...
+      const articles = await query({
+        query: ArticleList,
+        variables: {
+          first: 10
+        }
+      })
+      expect(articles).toMatchSnapshot({
+        data: {
+          articles: {
+            nodes: expect.any(Array),
```
roosnic1, line 108: can you try the following.
`nodes: Array.from({length: articleIDs.length}, () => ({id: expect.any(String)}))`

roosnic1 (changes requested):
looks very good. Just some minor changes.

packages/api/__tests__/specs/article.ts
penk85: Good to know!

bishoymelek (approved):

PR #101, opened 2020-12-09
bug fix - Delete-Dialog not disappearing in pageList

### LINKED ISSUE #73: Delete-Dialog not disappearing in pageList
**Expected behaviour:** upon clicking the confirm-button, the dialog window should disappear and the page-list should be reloaded, not showing the deleted page anymore

**Actual behaviour:** upon clicking the confirmation-button, the dialog window doesn't disappear; it can only be made to disappear by clicking the cancel-button. The page then doesn't reload, the deleted page is still shown. Only by reloading the page manually the page-list doesn't show the deleted page anymore. 

56c2a314-ed0a-49d8-a24d-b50540dfc2d4

persunde:
Fixed #73 Delete-Dialog not disappearing in pageList

The code in `pageList.tsx` tries to query the cache. Since the fetchPolicy was `no-cache` the cache was empty and it threw an error. 

```
// This throws an error, since the cache is empty
const query = cache.readQuery<PageListQuery>({
  query: PageListDocument,
  variables: listVariables
})
```
If you dont want to use a cache, you can use the `refetch` function returned from `usePageListQuery`.
Or you can change the fetchPolicy to something else, like `cache-and-network`. I am not sure what you prefer.

I changed to the fetchPolicy that was the closest to the one originally used `no-cache`, but a fetchPolicy that would still write to cache so to avoid the error. This way the `cache.readQuery<PageListQuery>(...)` function works and returns the latest (cached) results once a page is deleted.

### DIFF
--- a/packages/editor/src/client/routes/pageList.tsx
+++ b/packages/editor/src/client/routes/pageList.tsx
@@ -41,7 +41,7 @@ export function PageList() {
   const listVariables = {filter: filter || undefined, first: PagesPerPage}
   const {data, fetchMore, loading: isLoading} = usePageListQuery({
     variables: {filter: filter || undefined, first: 50},
-    fetchPolicy: 'no-cache'
+    fetchPolicy: 'network-only'
   })
 
   useEffect(() => {


0815pascal (approved):

roosnic1 (approved):

roosnic1:
Thank you very much @persunde. We definitely need a refactoring task to setup those fetch and refetch policies consistent through out the code.

penk85 (approved):

bishoymelek (approved):

PR #105, opened 2020-12-14
T/api author tests

penk85:
Tests for author on api

### DIFF
--- a/packages/api/__tests__/specs/author.ts
+++ b/packages/api/__tests__/specs/author.ts
@@ -36,7 +36,28 @@ describe('Authors', () => {
       const authorInput: AuthorInput = {
         name: 'JRR Tolkien',
         slug: `tolkien-${authorIds.length}`,
-        bio: []
+        links: [
+          {title: 'link 1', url: 'www.link1.ch'},
+          {title: 'link 2', url: 'www.link2.ch'}
+        ],
+        bio: [
+          {
+            type: 'heading-one',
+            children: [
+              {
+                text: 'Test Bio'
+              }
+            ]
+          },
+          {
+            type: 'paragraph',
+            children: [
+              {
+                text: 'bio text 😀'
+              }
+            ]
+          }
+        ]
       }
       const res = await mutate({
         mutation: CreateAuthor,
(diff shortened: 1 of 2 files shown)


packages/api/__tests__/specs/author.ts
```diff
@@ -0,0 +1,160 @@
...
+
+describe('Authors', () => {
+  describe('can be created/edited/deleted:', () => {
+    let authorIds: string[] = []
+    beforeEach(async () => {
+      const {mutate} = testClientPrivate
+      const authorInput: AuthorInput = {
+        name: 'JRR Tolkien',
+        slug: `tolkien-${authorIds.length}`,
+        bio: []
```
roosnic1, line 39: I suggest you add some content to the bio. You can add an author with a bio through the editor and the copy the content from the database.

packages/api/__tests__/specs/author.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
```
roosnic1, line 40: And i would also add some links. They are optional but should be tested too.

roosnic1 (changes requested):
Almost done. Just some more information for the authors.

roosnic1 (approved):
🎉

packages/api/__tests__/specs/author.ts
```diff
@@ -0,0 +1,202 @@
...
+  CreateAuthor,
+  AuthorInput,
+  AuthorList,
+  Author,
+  UpdateAuthor,
+  DeleteAuthor
+} from '../api/private'
+
+let testClientPublic: ApolloServerTestClient
+let testClientPrivate: ApolloServerTestClient
```
0815pascal, line 14: did you use `let` on purpose here instead of `const` ? I assume that the content of `testClientPrivate` etc. won't be changing? But then again, that's just a minor remark :)

0815pascal (approved):

packages/api/__tests__/specs/author.ts
roosnic1: @0815pascal this is just a declaration. The assignment happens 6 line below. It's not possible to use `const` here.

0815pascal: Ok, all good then

bishoymelek (approved):

PR #108, opened 2020-12-15
add placeholder image and use it

bishoymelek:
- Add placeholder Image to `/static` to be used instead of an API Placeholder provider.

- Add styling prop for the Image tag in `chooseEditImage` Atom and start using the placeholder image.

- Use the atom in `EditPeerPanel` with the styling prop and new placeholder image


packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -191,16 +192,14 @@ export function PeerEditPanel({id, onClose, onSave}: ImageEditPanelProps) {
...
-                  profile?.logo?.previewURL ?? 'https://via.placeholder.com/240x240'
-                }`
+            <ChooseEditImage
+              disabled
+              imageStyle={{
+                display: 'block',
+                marginLeft: 'auto',
+                marginRight: 'auto'
               }}
+              image={profile?.logo?.previewURL ?? '/static/placeholder-240x240.png'}
```
roosnic1, line 202: is the `?? '/static/...'` necessary? you check for a valid url inside the component.

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -191,16 +192,14 @@ export function PeerEditPanel({id, onClose, onSave}: ImageEditPanelProps) {
...
-              style={{
-                height: '200px',
-                marginBottom: '20px',
-                backgroundSize: 'cover',
-                backgroundImage: `url(${
-                  profile?.logo?.previewURL ?? 'https://via.placeholder.com/240x240'
-                }`
+            <ChooseEditImage
+              disabled
+              imageStyle={{
```
roosnic1, line 197: I think you can set them inside the component and check a few places where the component is used. I think it's needed everywhere.

roosnic1 (changes requested):
looks good. But you can add the styles for the component in general.

packages/editor/src/client/panel/peerEditPanel.tsx
bishoymelek: I did that at the beginning but I am not sure about it. I will go through other places as well to see it

packages/editor/src/client/panel/peerEditPanel.tsx
bishoymelek: I am sorry I don't get what you mean

0815pascal (dismissed):

roosnic1 (changes requested):
Tested this locally with the demo instance and got the following:
image

Did you test this?

roosnic1 (approved):
👍🏾

0815pascal (approved):

penk85 (approved):

PR #115, opened 2020-12-17
F/wep 128 richtext table border colorpicker

faebebin:
Intruduce Colorpicker UI component. Does not touch the data schema of API. Colorpicker as UI handler to set borderColor to any color value. Previously one could only switch between ['black', 'transparent'].


packages/editor/src/client/atoms/toolbar.tsx
```diff
@@ -149,27 +146,42 @@ export function SubMenuButton({children, icon}: ToolbarButtonProps & ToolbarIcon
...
-        width: '1px',
-        alignSelf: 'stretch',
+  return <Divider style={{height: '1.5em'}} vertical />
+}
 
-        marginLeft: 20,
-        marginRight: 20,
-        marginTop: 2,
-        marginBottom: 2,
+interface FlexHDivProps {
```
roosnic1, line 165: A more descriptive name maybe. What does the `H` stand for?

packages/editor/src/client/atoms/toolbar.tsx
```diff
@@ -149,27 +146,42 @@ export function SubMenuButton({children, icon}: ToolbarButtonProps & ToolbarIcon
...
+  children: ReactNode
+  dividerTop?: boolean
+  dividerBottom?: boolean
+}
 
-        backgroundColor: 'gray'
-      }}
-    />
+export function HBar({children, dividerTop, dividerBottom}: FlexHDivProps) {
```
roosnic1, line 171: A more descriptive name maybe. What does the `H` stand for?

packages/editor/src/client/blocks/richTextBlock/tableMenu.tsx
```diff
@@ -1,34 +1,64 @@
...
   const {t} = useTranslation()
 
-  const isBorderVisible = () => {
-    const [match] = Editor.nodes(editor, {
-      match: node => (node.borderColor && node.borderColor !== 'transparent') as boolean,
-      mode: 'all'
-    })
+  useEffect(() => {
+    // set borderColor in slate node tree on colorPicker change
+    const tablePath = nearestAncestor(editor, BlockFormat.Table)?.path
```
roosnic1, line 28: What does `nearestAncestor` exactly do?

roosnic1 (commented):
Looks very good. Just some questions and tiny naming improvements.

packages/editor/src/client/atoms/emojiPicker.tsx
```diff
@@ -4,14 +4,14 @@ import './emojiPicker.less'
 import {Picker, BaseEmoji} from 'emoji-mart'
 
 interface EmojiPickerProps {
-  doWithEmoji: (emoji: string) => void
+  withEmoji: (emoji: string) => void
```
faebebin, line 7: for consistent naming with colorPicker component

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -99,9 +99,9 @@ export const RichTextBlock = memo(function RichTextBlock({
 
         <ToolbarDivider />
 
-        <EditorSubMenuButton icon="smile-o" editorHasFocus={hasFocus}>
-          <EmojiPicker doWithEmoji={emoji => editor.insertText(emoji)} />
-        </EditorSubMenuButton>
+        <SubMenuButton icon="smile-o">
```
faebebin, line 102: Only renamed the SubmenuButton. No functional changes.

packages/editor/src/client/atoms/colorPicker.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import nanoid from 'nanoid'
+
+interface ColorPickerProps {
+  withColor: (color: string) => void
```
0815pascal, line 5: Do you mean by `withColor` the new color that you pick? The naming is not very clear to me

packages/editor/src/client/blocks/richTextBlock/toolbar/tableMenu.tsx
```diff
@@ -67,50 +91,34 @@ export function TableMenu() {
...
+            <ControlsContainer dividerBottom>
+              <ColorPicker
+                withColor={color => {
+                  setBorderColor(color)
+                }}
+                currentColor={borderColor}
+                label={t('blocks.richTextTable.border')}
+              />
+              <button className="icon-button" onClick={() => setBorderColor('transparent')}>
+                <Icon icon="ban" style={{color: 'red'}} />
```
0815pascal, line 108: somewhere further above you've changed 'black' to HEX '000000'. In order to by consistent you might as well chose to set a HEX value for red? It's #FF0000  and for 'transparent' you could use #00000000 (looks weird but it's black with full transparency). But in the end I guess it's up to you what you use, maybe just use consistently the same way :)

0815pascal (approved):
Looks good, just two little things that you might want to address or not - up to you :)

packages/editor/src/client/atoms/colorPicker.tsx
roosnic1: @0815pascal `withXYZ` is quite a common naming in react. But @faebebin don't you think `setColor` would be more clear here?

packages/editor/src/client/atoms/emojiPicker.tsx
roosnic1: same argument here about `setEmoji`

packages/editor/src/client/atoms/toolbar.tsx
```diff
@@ -166,20 +162,31 @@ export const SubMenuButton = forwardRef<PopoverProps, SubMenuButtonProps>(
...
+  children: ReactNode
+  dividerTop?: boolean
+  dividerBottom?: boolean
+}
 
-        backgroundColor: 'gray'
-      }}
-    />
+export function ControlsContainer({children, dividerTop, dividerBottom}: ControlsContainerProps) {
+  // TODO find rsuite component
```
roosnic1, line 175: Do you want to create a ticket for this?

packages/editor/src/client/blocks/richTextBlock/editor/elements.ts
```diff
@@ -10,7 +10,7 @@ export interface TableCell extends SlateElement {
   borderColor: string
 }
 
-export const defaultBorderColor: TableCell['borderColor'] = 'black'
+export const defaultBorderColor: TableCell['borderColor'] = '#000000'
```
roosnic1, line 13: `export const DEFAULT_BORDER_COLOR`

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
roosnic1: 👍🏾

roosnic1 (changes requested):
looks very good. just some tiny changes.

packages/editor/src/client/blocks/richTextBlock/editor/elements.ts
bishoymelek: @roosnic1  maybe we need a naming conventions' list for different variables.

bishoymelek (approved):
Great work :100: , haven't noticed anything except the mentioned comments, as naming for methods.

packages/editor/src/client/blocks/richTextBlock/toolbar/tableMenu.tsx
faebebin: Agree. Personally I like using names as 'black' or 'red'. Setting the <input value={'black'} ... > caused a console warning, saying something like `expected '#xxyycmzz' string ... ` therefore I switched to hex. Therefore I would use names where possible still. If you prefer consistency I will change?

packages/editor/src/client/atoms/toolbar.tsx
faebebin: Jep, think this would also be a nice first issue ?

packages/editor/src/client/atoms/colorPicker.tsx
faebebin: IMO withColor implies a callback getting passed the color as arg. Setcolor / Setemoji would rather mean passing a value as Prop to the component?

packages/editor/src/client/blocks/richTextBlock/toolbar/tableMenu.tsx
0815pascal: I don't really have a strong opinion on this, maybe @roosnic1 knows better if it matters what we use. But I just did a little research and I came upon this: [Here](http://thenewcode.com/781/Which-CSS-Color-System-To-Use-Where) they advise against using color definitions like `black` and use HEX or other systems instead.

faebebin: Changed it, but now it is more inconsistent with rest of the codebase. Think mainly keyword colors are used in the codebase. And they used 'transparent' everywhere. As mentioned I used '#0000000' for black to avoid a console warning on the html basic input element, as an exception. Maybe there would be better ways getting around this.

0815pascal (approved):
Looks all good now 👌🏼

packages/editor/src/client/atoms/toolbar.tsx
roosnic1: 👍🏾

roosnic1 (approved):
💯

PR #118, opened 2020-12-24
F/wep 159 refactor seperate editor tools modules

faebebin:
Imroved naming for slate editor related modules. This should reflect the terminology used in the Slate docs the discussions found in the web when searching for certain aspects of slate editor.


packages/editor/src/client/blocks/richTextBlock/editor/render.tsx
```diff
@@ -0,0 +1,145 @@
...
+              element.borderColor === 'transparent'
+                ? `rgb(0, 0, 0, 0.1)`
+                : (element.borderColor as string)
+          }}>
+          {children}
+        </td>
+      )
+
+    case InlineFormat.Link:
+      // TODO: Implement custom tooltip
```
0815pascal, line 51: What about this code? Do you still need this for later?

packages/editor/src/client/blocks/richTextBlock/editor/render.tsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  if (leaf[TextFormat.Subscript]) {
+    children = <sub {...attributes}>{children}</sub>
+  }
+
+  return <span {...attributes}>{children}</span>
+}
+
+// TODO: Re-implement via `normalizeNode`
```
0815pascal, line 94: Same question here: do you still need this?

0815pascal (approved):
Looks good in general. I only noticed that you have some uncommented code-fragments and I'm not sure if you're planning to develop this branch further or if you consider it finished. If so I would remove those comments

roosnic1 (approved):
🎉

packages/editor/src/client/blocks/richTextBlock/editor/render.tsx
faebebin: Yes. This will be done in the follow up PR. I read more about it and it turned out to be quite important in our case. It will serve as data validation step important for our table elements which are deeply nested.

packages/editor/src/client/blocks/richTextBlock/editor/render.tsx
faebebin: Yes, the Link node still needs some fixes, I think.

PR #120, opened 2020-12-29
Mail provider for Slack

roosnic1:
This PR introduces a mail provider which posts mails into a slack channel. This is very useful for development. The implementation is in the `examples/api`since it should never be released. It will override the default mailgun mail provider if the env variable `SLACK_DEV_MAIL_WEBHOOK_URL` is set.


penk85 (approved):

faebebin (approved):
Looks good

bishoymelek (approved):
Good One!

0815pascal (approved):
Looking good!

PR #121, opened 2020-12-29
API with hooks

roosnic1:
but for now we call them events 😄


bishoymelek (dismissed):

penk85 (approved):

packages/api/src/server.ts
```diff
@@ -17,6 +20,40 @@ export class WepublishServer {
   constructor(opts: WepublishServerOpts) {
     const app = express()
 
+    const {dbAdapter} = opts
+
+    methodsToProxy.forEach(mtp => {
+      if (mtp.key in dbAdapter) {
+        mtp.methods.forEach(method => {
+          const methodName = `${method}${capitalizeFirstLetter(mtp.key)}`
+          // @ts-ignore
```
0815pascal, line 29: Did you leave those `@ts-ignore` on purpose here?

0815pascal (dismissed):
All good, only one question concerning `@ts-ignore`

packages/api/src/server.ts
roosnic1: Yes. If i leave the out typescript complains and at the moment i have no idea how to type this.

faebebin: Think would have a fix there, will push later.

faebebin (changes requested):

faebebin (dismissed):
Looks good. And all checks pass locally too.

penk85 (approved):

0815pascal (approved):
Looks good

PR #130, opened 2021-01-06
T/api navigations

penk85:
Testing CRUD actions on Navigation.


packages/api/__tests__/specs/navigation.ts
```diff
@@ -0,0 +1,223 @@
...
+        }
+      })
+      expect(res).toMatchSnapshot({
+        data: {
+          updateNavigation: {
+            id: expect.any(String)
+          }
+        }
+      })
+      expect(res.data?.updateNavigation.id).toBe(ids[0])
```
roosnic1, line 196: I would test here if the updated data is updated. for example check if the label says `New Page Label`

roosnic1 (approved):
👍🏾

0815pascal (approved):

PR #134, opened 2021-01-08
B/WEP-175-fix-imageBlock

0815pascal:


roosnic1 (changes requested):
Looks good but the same issue is also in other blocks and panels.

image

penk85 (approved):

packages/editor/src/client/atoms/chooseEditImage.tsx
```diff
@@ -22,20 +22,22 @@ export function ChooseEditImage({
...
   const {t} = useTranslation()
   header = header ?? t('chooseEditImage.header')
+  const panelHeight = image ? (image.height < 240 ? image.height : 240) : 240
   return (
-    <Panel
-      header={header}
-      style={{
-        height: 300
-      }}>
+    <Panel header={header} style={{}}>
```
roosnic1, line 27: why the empty object in style? if you don't have anything you can remove the property.

roosnic1 (changes requested):
Just some tiny changes

bishoymelek (approved):
Looks way better! :muscle:

packages/editor/src/client/atoms/chooseEditImage.tsx
0815pascal: Fixed it

faebebin (approved):
Looks good.

packages/editor/src/client/panel/authorEditPanel.tsx
```diff
@@ -139,62 +140,74 @@ export function AuthorEditPanel({id, onClose, onSave}: AuthorEditPanelProps) {
...
-          <RichTextBlock value={bio} onChange={value => setBio(value)} />
-        </Panel>
+              </InputGroup>
+            </Form>
+          </Panel>
+          <Panel header={t('authors.panels.image')}>
+            <ChooseEditImage
+              image={image}
+              header={''}
+              top={0}
```
bishoymelek, line 166: @0815pascal  Great work, just asking about the usage of top and left values? why not all the sides?

packages/editor/src/client/routes/imageList.tsx
```diff
@@ -127,10 +127,13 @@ export function ImageList() {
...
         <FlexboxGrid justify="space-around" style={{marginTop: '20px'}}>
           {images.map((image, key) => (
-            <FlexboxGrid.Item colspan={7} style={{marginBottom: '20px'}} key={key}>
+            <FlexboxGrid.Item colspan={7} style={{marginBottom: '20px', maxWidth: '300'}} key={key}>
               <Link route={ImageEditRoute.create({id: image.id}, current ?? undefined)}>
-                <Panel shaded bordered bodyFill>
-                  <img src={image.thumbURL || ''} />
+                <Panel shaded bordered bodyFill style={{height: '200', width: 'calc(100% + 2px)'}}>
+                  <img
+                    src={image.mediumURL || ''}
```
bishoymelek, line 134: @0815pascal  I am not sure how thumbnail and medium etc. URLs are being generated for images but if there is a thumbURL? and if not I guess we may show a placeholder image :S what do you think?

packages/editor/src/client/panel/authorEditPanel.tsx
0815pascal: `top` and `left` are here used to position the wrench icon. Because its position is set relative to its parent element, it's sufficient if we only indicate `left` and `top`

packages/editor/src/client/routes/imageList.tsx
0815pascal: Yes it's true that there is also a `thumbURL` but I figured that `thumbURL` causes the picture to be cropped which sometimes results in weird looking images. If you go to `packages/editor/src/client/image.graphql` you can see how the different ImageURLs are generated: 

```
mediumURL: transformURL(input: {width: 300})
thumbURL: transformURL(input: {width: 280, height: 200})
```
The advantage of using `mediumURL` in my mind is that it only reduces *width* and so it's scaled down proportionally. When using `thumbURL` we're forced into using certain *width* and *height* which causes the image to be cropped. 

But then again I might have misunderstood something here...

bishoymelek: Thanks for the explanation, I understand now :+1:

bishoymelek (approved):

roosnic1 (changes requested):
I think you forgot one:

image

roosnic1 (dismissed):
💯

bishoymelek (approved):

faebebin (approved):

PR #136, opened 2021-01-10
B/wep 174 fix focus empty rtb firefox

faebebin:
Update: thanks for your reviews. @roosnic1 's idea to implement a helper as an editor method made me add quite some changes to stay consistent. This new commit is only refactoring and should not change any behaviour you reviewed already. I tried to stick to the Slate docs with implementing this change.

The problem with re-focusing an empty richtextblock, which only occured on Firefox, was related to the placeholder of an empty block. The problem seems popular on the web and not nicely solved. Hopefully there will be a solution in future releases. For now the workaround is just removing the builtin placeholder. I also added a little alternative placeholder div. The placeholder could be left aside completely (blank) if it is more disturbing than helping in your opinions. The Firefox users, please really try the new behaviour and check if improved or some problems presist.


packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -94,13 +98,21 @@ export const RichTextBlock = memo(function RichTextBlock({
...
       <Editable
         readOnly={disabled}
-        placeholder={t('blocks.richText.startWriting')}
+        // placeholder={t('blocks.richText.startWriting')}  # causes focusing problems on firefox !
         renderElement={renderElement}
         renderLeaf={renderLeaf}
         onBlur={() => {
           setLocation(editor.selection)
+          setIsEmpty(
+            JSON.stringify(Array.from(editor.children)) === JSON.stringify(createDefaultValue())
```
roosnic1, line 114: maybe move this to a function inside the editor? `editor.isEmpty()` which returns a boolean. Might be useful in other places.

bishoymelek (dismissed):

0815pascal (approved):

penk85 (approved):
Looks good and works smoothly in Firefox.

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
faebebin: Good idea. Will update a suggestion according to https://docs.slatejs.org/concepts/07-plugins#helper-functions .

roosnic1 (approved):
👍🏾

penk85 (approved):

PR #137, opened 2021-01-11
b/WEP-178 chech ImageTeaserBlock if the image prop exist and has value, or set placeholder image instead

bishoymelek:
Image Teaser Component is expecting the image to has a value but it's not the case if you add an article that has no featured image; So it broke.


examples/website/src/shared/teasers/imageTeaser.tsx
```diff
@@ -125,16 +125,20 @@ export function ImageTeaser({
...
         <div className={css(ImageTeaserImageStyle)}>
           <Image
-            src={image.format === 'gif' ? image.url : image.largeURL}
-            alt={image.description || image.caption}
+            src={
+              image?.format === 'gif'
+                ? image?.url
+                : image?.largeURL || 'https://via.placeholder.com/240x240'
+            }
+            alt={image?.description || image?.caption}
```
faebebin, line 141: Good workaround. Only, to me looks like there is something wrong higher up. The image prop seems not optional and `ImageData` interface requires `url` and `format` to be set. So should not the workaround be set somewhere else?

faebebin (changes requested):

examples/website/src/shared/teasers/imageTeaser.tsx
roosnic1: @faebebin you're right. @bishoymelek will check in the api and editor if the user has selected this teaser style without an image and then throw an error.

@bishoymelek: You might want to convert this PR to a draft until you're ready again.

roosnic1 (changes requested):
as discussed

examples/website/src/shared/teasers/imageTeaser.tsx
bishoymelek: @faebebin  Yes Absolutely correct. discussed it with @roosnic1  and now has validation in the editor and should be on the API level as well afterwords.

packages/editor/src/client/locales/en.json
```diff
@@ -118,7 +118,8 @@
         "hideAuthors": "Hide Authors",
         "socialMediaImage": "Image for Social Media",
         "postImage": "Image for Post",
-        "metadataInfo": "Information from these fields are being used for previews in Social Media Posts. If left blank, information from general-tab will be used instead."
+        "metadataInfo": "Information from these fields are being used for previews in Social Media Posts. If left blank, information from general-tab will be used instead.",
+        "lightStyledWithNoImageError": "Light styled teaser must has an image"
```
roosnic1, line 122: "Light styled teasers must have an image".

Also the location of the translation is wrong. You're in the `article` translations and i think this should go to the `pages` translations.

packages/editor/src/client/panel/teaserEditPanel.tsx
```diff
@@ -36,7 +37,7 @@ export function TeaserEditPanel({
   onClose,
   onConfirm,
   closeLabel = 'Close'
-}: TeaserEditPanelProps) {
+}: TeaserEditPanelProps): JSX.Element {
```
roosnic1, line 40: Why did you add this?

roosnic1 (changes requested):
looks good. Just some minor changes.

packages/editor/src/client/panel/teaserEditPanel.tsx
faebebin: @roosnic1 FYI our linting sometimes (maybe it is just my vscode) requires return type and highlights as warning. In another project we add this return type to every FC to satisfy this constraint. @bishoymelek I think we do not require return type for FC here. But maybe we have to either fix lint rules somewhere or strictly request to have consistency within the repo and the linting.

roosnic1: @faebebin and @bishoymelek you might have to configure your editors to read the eslint config in the `package.json`

faebebin: @roosnic1 yes if you never saw this then will check my config.

examples/website/src/shared/teasers/imageTeaser.tsx
faebebin: @roosnic1 I checked my local settings. I have eslint activated, but that is project setting,no? Just, I did not find some setting to change by now. Don't you have those highlightings requiring return type:
image

packages/editor/src/client/panel/teaserEditPanel.tsx
faebebin: @roosnic1 @roosnic1 I checked my local settings. I have eslint activated, but that is project setting,no? Just, I did not find some setting to change by now. Don't you have those highlightings requiring return type:
image

faebebin: And when runnging: yarn lint:editor

image

faebebin: ? If this is on me I will keep on working on my local settings :)

packages/editor/src/client/locales/en.json
bishoymelek: I guess in both article and page?

packages/editor/src/client/panel/teaserEditPanel.tsx
bishoymelek: @roosnic1 @faebebin Yes I guess it's the project's eslint configuration that always makes these warnings, I actually thought to run linting before git commit so it's always aligned correctly.

packages/editor/src/client/locales/en.json
bishoymelek: @roosnic1 I added it in both article and page panels, tell me what you think

roosnic1 (approved):
🎉

faebebin (approved):

PR #140, opened 2021-01-19
Log all the things

roosnic1:


packages/api/src/server.ts
```diff
@@ -40,21 +51,22 @@ export class WepublishServer {
...
               }
             })
           } else {
-            console.warn(`${methodName} does not exist in dbAdapter[${mtp.key}]`)
+            logger('server').warn('%s does not exist in dbAdapter[%s]', methodName, mtp.key)
           }
         })
       } else {
-        console.warn(`${mtp.key} does not exist in dbAdapter`)
+        logger('server').warn('%s does not exist in dbAdapter')
```
penk85, line 69: Should this have the variable? `('%s does not exist in dbAdapter', mtp.key)`

penk85 (approved):

bishoymelek (approved):
:+1: :tada:

bishoymelek (approved):
Great started using it already in the cron jobs 👌

faebebin:
Will review tomorrow

packages/api/src/server.ts
```diff
@@ -102,6 +121,11 @@ export class WepublishServer {
       cors: corsOptions
     })
 
+    app.use((err: any, req: Express.Request, res: Express.Response, next: NextFunction) => {
```
0815pascal, line 124: Not sure if I misunderstand something here but where do you use `req` and `res` in this function? I only see `err` being used

0815pascal (changes requested):

packages/api/src/server.ts
roosnic1: because `next` is also used

faebebin (approved):
Looks good.

PR #141, opened 2021-01-19
B/wep 189 richtext placeholder nonempty text

faebebin:
Now placeholder only shows up on empty blocks. Staying on a page things work smoothly now. If you save an empty block (respectively the article or page ...) and then reopen, the placeholder will not show up on the block. But I tried couple of things and seems there is some more complexity involved not worth the effort for saving empty blocks as it makes not so much sense IMO.


0815pascal (approved):
All good and also runs locally 👍🏼

bishoymelek (approved):

roosnic1 (approved):
👍🏾

penk85 (approved):

PR #142, opened 2021-01-21
F/wep 195 update comment mutation

bishoymelek:
Added Update comment mutation, add more revisions


0815pascal (approved):
Looks good to me 👍🏼

packages/api/src/graphql/mutation.public.ts
```diff
@@ -99,6 +103,21 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
         })
       }
+    },
+
+    updateComment: {
+      type: GraphQLNonNull(GraphQLPublicComment),
+      args: {
+        input: {type: GraphQLNonNull(GraphQLPublicCommentUpdateInput)}
+      },
+      async resolve(_, {input}, {dbAdapter}) {
```
roosnic1, line 113: You need to check here if the user is authenticated. Also you should check if the user is the same as the comment they are trying to updated. Otherwise return a 403.

roosnic1 (changes requested):
Looks good but needs better security

packages/api/src/graphql/mutation.public.ts
roosnic1: @bishoymelek also as discussed check if the comment is in the `requestForChange` state.

faebebin (approved):

packages/api/src/graphql/mutation.public.ts
```diff
@@ -9,14 +9,16 @@ import {
...
+  UserNotFoundError,
+  NotAuthorisedError as NotAuthorizedError
 } from '../error'
 import {
   GraphQLPublicCommentInput,
   GraphQLPublicCommentUpdateInput,
   GraphQLPublicComment
 } from './comment'
 import {CommentAuthorType, CommentState} from '../db/comment'
+import {UserInputError} from 'apollo-server-express'
```
roosnic1, line 21: What's the difference between `ApolloError` and `UserInputError`. Also could you move the Error definition to `packages/api/src/error.ts` like the other errors.

roosnic1 (changes requested):
looks good

packages/api/src/graphql/mutation.public.ts
bishoymelek: UserInputError is extended from the base `ApolloError` which can provide more details for the user input data error like the invalid arguments passed. 

I haven't added it to `error.ts` and used it as I found no importance to have an empty new class :S  What do you think?

PR #143, opened 2021-01-22
Pagination for tables and API

roosnic1:
Enables pagination on the list queries. Currently only on the user list query. Will be later implemented later for other list queries also.

⚠️ PR will be merged into `f/WEP-23-member-plans`


faebebin (approved):
Looks good. But I am not so deep into API yet ;)

packages/editor/src/client/routes/userList.tsx
```diff
@@ -119,51 +138,90 @@ export function UserList() {
...
+            <Cell dataKey={'name'}>
+              {(rowData: FullUserFragment) => (
+                <Link route={UserEditRoute.create({id: rowData.id})}>
+                  {rowData.name || t('userList.overview.unknown')}
+                </Link>
+              )}
+            </Cell>
+          </Column>
+          <Column width={400} align="left" resizable>
+            <HeaderCell>{t('email')}</HeaderCell>
```
0815pascal, line 181: I would rather use here `userList.overview.email` in order to be consistent with the rest

packages/editor/src/client/routes/userList.tsx
```diff
@@ -119,51 +138,90 @@ export function UserList() {
...
+            <Cell dataKey="email" />
+          </Column>
+          <Column width={100} align="center" fixed="right">
+            <HeaderCell>{t('action')}</HeaderCell>
```
0815pascal, line 185: I would rather use here `userList.overview.action` in order to be consistent with the rest

0815pascal (changes requested):
Looks good, just two tiny things and I noticed that you forgot to add the German and French translation files.

penk85 (approved):

faebebin (approved):
Looks good.

bishoymelek (changes requested):

packages/api/src/db/user.ts
```diff
@@ -58,6 +67,15 @@ export interface UserSubscriptionPeriod {
   readonly invoiceID: string
 }
 
+export interface UserSubscriptionPeriodInput {
```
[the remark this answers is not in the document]
roosnic1: It's the same but they might change in the future. One is to handle the input and the other is the whole representation.

PR #147, opened 2021-01-22
F/wep 155 action on comment mutation

bishoymelek:


packages/api/src/graphql/mutation.private.ts
```diff
@@ -737,6 +740,32 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
         authorise(CanPublishPage, roles)
         return dbAdapter.page.unpublishPage({id})
       }
+    },
+
+    takeActionOnComment: {
```
roosnic1, line 745: I read in an article that this is considered bad practice. Every mutation should have one job and i agree with that philosophy. I suggest you create two mutations called `approveComment` and `rejectComment`. The database adapter method on the other hand should be simple and can be called `updateComment`.

packages/api/src/graphql/mutation.private.ts
```diff
@@ -737,6 +740,32 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+      async resolve(root, {id, input}, {authenticate, dbAdapter}) {
+        const {roles} = authenticate()
+        authorise(CanTakeActionOnComment, roles)
+
+        const {state, rejectionReason} = input
+        if (
+          [CommentState.PendingUserChanges, CommentState.Rejected].includes(state) &&
+          !rejectionReason
+        ) {
+          throw new Error('Rejection Reason is required')
```
roosnic1, line 760: If you throw errors in the GraphQL API they should extend `ApolloError`. Check the file `packages/api/src/error.ts`

roosnic1 (changes requested):

packages/api/src/graphql/mutation.private.ts
bishoymelek: After changing the mutation to multiple ones, we don't need this check anymore.

packages/api/src/graphql/mutation.private.ts
bishoymelek: Agree, We currently have three mutations to approve, reject or request changes from the user.

roosnic1 (approved):
👍🏾

penk85 (approved):

PR #150, opened 2021-01-24
Secure delete of elements

roosnic1:


roosnic1:
PR is too old and will be solved in the future with a refactoring.

PR #164, opened 2021-02-02
F/wpc 148 get public unapproved comments for auth user

bishoymelek:


packages/api/src/graphql/comment.ts
```diff
@@ -172,6 +173,10 @@ export const GraphQLPublicComment: GraphQLObjectType<
 
     text: {type: GraphQLNonNull(GraphQLRichText)},
 
+    state: {type: GraphQLString},
```
roosnic1, line 176: `GraphQLNonNull` or did you leave it out on purpose?

packages/api/src/db/comment.ts
```diff
@@ -106,6 +106,11 @@ export interface GetCommentsArgs {
   readonly order: SortOrder
 }
 
+export interface GetPublicCommentsArgs {
+  readonly id: string
+  readonly userID: string
```
roosnic1, line 111: `userID` should be optional right? because you don't always have the user

roosnic1 (changes requested):
Just some tiny corrections

packages/api/src/db/comment.ts
bishoymelek: My bad

faebebin (approved):

packages/api-db-mongodb/src/db/comment.ts
```diff
@@ -210,14 +213,24 @@ export class MongoDBCommentAdapter implements DBCommentAdapter {
         .match({
           $and: [{itemID: id}, {state: CommentState.Approved}]
         })
+        .toArray(),
+      this.comments
+        .aggregate([{$sort: {'modifiedAt.date': -1}}], {
+          collation: {locale: this.locale, strength: 2}
+        })
+        .match({
+          $and: [{itemID: id}, {userID}, {state: {$ne: CommentState.Approved}}]
```
roosnic1, line 222: i think you need to check if `userID` is undefined. Otherwise you'll get all here all comments that are not approved.

roosnic1 (changes requested):

penk85 (approved):

packages/api-db-mongodb/src/db/comment.ts
bishoymelek: I guess if it has no value "undefined" it won't match with any comments, and I tested it and seems working :S But what do you think?

roosnic1: This is right but we don't know how this will evolve maybe in the future we have anonymous comments with userID undefined. Also performance wise it's faster to not execute the query if we don't have a userID.

bishoymelek: totally agree

roosnic1 (approved):
🎉

PR #167, opened 2021-02-03
WPC-202 rich-text Get characters count method, and get string text method

bishoymelek:
**Characters Count:**
- Get characters' count method
- Show/Hide prop for the count.

**String Text:**
- Get string text method.
- Get string text method as a prop

**Note: it should be reviewed after merging f/WPC-212**


packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -15,14 +15,24 @@ import {withNormalizeNode} from './editor/normalizing'
 import {LinkFormatButton, RemoveLinkFormatButton} from './toolbar/linkButton'
 import {TableMenu} from './toolbar/tableMenu'
 import {WepublishEditor} from './editor/wepublishEditor'
+import {Alert} from 'rsuite'
 
-export type RichTextBlockProps = BlockProps<RichTextBlockValue>
+export type RichTextBlockProps = BlockProps<RichTextBlockValue> & {
+  maxCharsLimit?: number
+  getCharsLength?: any
```
faebebin, line 22: I think the type shoud be: `(text: string) => number` .
But I do not really see why this method is necessary at all?

faebebin, line 22: Could this have a type?

faebebin (commented):
It looks good, but as it is not yet in use anywhere? it is a bit hard to say how it really behaves ...

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -18,14 +18,18 @@ import {WepublishEditor} from './editor/wepublishEditor'
...
 }
 
 export const RichTextBlock = memo(function RichTextBlock({
   value,
   autofocus,
   disabled,
   onChange,
-  displayOnly = false
+  displayOnly = false,
+  showCharCount = true,
```
roosnic1, line 31: default should be `false`.

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -18,14 +18,18 @@ import {WepublishEditor} from './editor/wepublishEditor'
 
 export interface RichTextBlockProps extends BlockProps<RichTextBlockValue> {
   displayOnly?: boolean
+  showCharCount?: boolean
+  getStringText?: (text: string) => void
```
roosnic1, line 22: Not needed.

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -122,6 +133,9 @@ export const RichTextBlock = memo(function RichTextBlock({
           setLocation(editor.selection)
         }}
       />
+      {showCharCount && (
+        <p style={{textAlign: 'right'}}>{`${t('blocks.richText.charsCount')}: ${charCount}`}</p>
```
roosnic1, line 137: https://www.i18next.com/translation-function/interpolation

packages/editor/src/client/blocks/richTextBlock/editor/wepublishEditor.ts
```diff
@@ -58,5 +58,17 @@ export const WepublishEditor = {
...
   isEmpty(editor: Editor) {
     return JSON.stringify(editor.children) === JSON.stringify(this.createDefaultValue())
+  },
+
+  calculateEditorCharCount(editor: Editor) {
+    return this.getTextString(editor).length
+  },
+
+  getTextString(editor: Editor) {
+    return [...Node.texts(editor)].reduce((string, nodePair, index) => {
```
roosnic1, line 68: maybe add a little comment what exactly you are doing here.

packages/api/__tests__/api/public/index.ts
```diff
@@ -228,6 +228,29 @@ export type InstagramPostBlock = {
   postID: Scalars['String']
 }
 
+export type Invoice = {
```
roosnic1, line 231: Why is this file in the PR?

packages/api/__tests__/api/private/index.ts
```diff
@@ -1199,6 +1199,7 @@ export type QueryUsersArgs = {
   before?: Maybe<Scalars['ID']>
   first?: Maybe<Scalars['Int']>
   last?: Maybe<Scalars['Int']>
+  skip?: Maybe<Scalars['Int']>
```
roosnic1, line 1202: why is this file in the PR?

roosnic1 (changes requested):

packages/editor/src/client/blocks/richTextBlock/editor/wepublishEditor.ts
faebebin: This solution looks good. Would be nice to have some unittests on this.

faebebin (changes requested):

packages/editor/src/client/blocks/richTextBlock/editor/wepublishEditor.ts
bishoymelek: @faebebin Do we already have unit tests for this block?

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
bishoymelek: Why?

packages/editor/src/client/blocks/richTextBlock/editor/wepublishEditor.ts
faebebin: No. But would be good to have IMO, there might be more methods in the future.

faebebin: @bishoymelek there are no unittests yet, but you could have a look a the normalizing.test.ts. 
I would create some editor instance with diefferent content and then invoke the method to assert the string.
Would be good to have this as template for future methods also.

Maybe something like:
expect(emptyParagraph).toMatch('')
expect(... text: ' '}).toMatch(' ')
expect(...{ text: 'hello' }).toMatch('hello')
expect(... complex block with tables and lists but not text ).toMatch('')
expect(... complex block with tables and lists but not text ).toMatch('concated text')

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
roosnic1: because we don't need the functionality in the block.

HanaAyman (approved):

0815pascal (approved):

packages/editor/src/client/blocks/richTextBlock/editor/wepublishEditor.ts
bishoymelek: @faebebin Will make separate PR

roosnic1 (approved):

faebebin (approved):

PR #168, opened 2021-02-04
WPC-212 Rich Text in display only mode

bishoymelek:
- Editing is disabled.

- The toolbar is hidden.

### DIFF
--- a/packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
+++ b/packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
@@ -58,58 +58,60 @@ export const RichTextBlock = memo(function RichTextBlock({
         }
       }}>
       {!displayOnly && (
-        <Toolbar
-          fadeOut={!hasFocus}
-          onMouseDown={e => {
-            e.preventDefault()
-            if (!hasFocus && location) focusAtPreviousLocation(location)
-          }}>
-          <FormatButton format={BlockFormat.H1}>
-            <H1Icon />
-          </FormatButton>
-          <FormatButton format={BlockFormat.H2}>
-            <H2Icon />
-          </FormatButton>
-          <FormatButton format={BlockFormat.H3}>
-            <H3Icon />
-          </FormatButton>
+        <>
+          <Toolbar
+            fadeOut={!hasFocus}
+            onMouseDown={e => {
+              e.preventDefault()
+              if (!hasFocus && location) focusAtPreviousLocation(location)
+            }}>
+            <FormatButton format={BlockFormat.H1}>
+              <H1Icon />
+            </FormatButton>
+            <FormatButton format={BlockFormat.H2}>
+              <H2Icon />
+            </FormatButton>
+            <FormatButton format={BlockFormat.H3}>
+              <H3Icon />
+            </FormatButton>
 
-          <ToolbarDivider />
+            <ToolbarDivider />
 
-          <FormatIconButton icon="list-ul" format={BlockFormat.UnorderedList} />
-          <FormatIconButton icon="list-ol" format={BlockFormat.OrderedList} />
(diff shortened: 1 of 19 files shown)


packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -16,13 +16,16 @@ import {LinkFormatButton, RemoveLinkFormatButton} from './toolbar/linkButton'
 import {TableMenu} from './toolbar/tableMenu'
 import {WepublishEditor} from './editor/wepublishEditor'
 
-export type RichTextBlockProps = BlockProps<RichTextBlockValue>
+export interface RichTextBlockProps extends BlockProps<RichTextBlockValue> {
+  displayOnly: boolean
```
roosnic1, line 20: make it optional

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -16,13 +16,16 @@ import {LinkFormatButton, RemoveLinkFormatButton} from './toolbar/linkButton'
...
+}
 
 export const RichTextBlock = memo(function RichTextBlock({
   value,
   autofocus,
   disabled,
-  onChange
+  onChange,
+  displayOnly
```
roosnic1, line 28: add a default value `displayOnly = false`

roosnic1 (changes requested):

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
faebebin: Do we need this additional arg? Couldn't just say if `disabled` then hide toolbar?

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -54,60 +57,62 @@ export const RichTextBlock = memo(function RichTextBlock({
...
+          <SubMenuButton icon="smile-o">
+            <EmojiPicker setEmoji={emoji => editor.insertText(emoji)} />
+          </SubMenuButton>
+        </Toolbar>
+      )}
       {WepublishEditor.isEmpty(editor) && ( // Alternative placeholder
         <div onClick={() => ReactEditor.focus(editor)} style={{color: '#cad5e4'}}>
           {t('blocks.richText.startWriting')}
         </div>
       )}
```
faebebin, line 113: Think the conditional hiding should include the placeholder, as it states 'start writing' ...

faebebin (changes requested):
Do we need `disabled` and `displayOnly` - isn't it same?

roosnic1:
@faebebin i think so. Sometimes (e.g.: connetion lost) we just want to disable the editor and other times we just want to display RichText.

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
bishoymelek: @roosnic1 why add default value false?

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
bishoymelek: Fixed :ok_hand:

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
roosnic1: @bishoymelek because `displayOnly` is a boolean and shouldn't be `undefined`

roosnic1 (changes requested):

faebebin (dismissed):

HanaAyman (approved):

roosnic1 (approved):

HanaAyman (approved):

PR #175, opened 2021-02-10
add displayOnly prop to richText

bishoymelek:
Call to action field is in display-only mode:

Screenshot from 2021-02-10 10-35-03


roosnic1 (approved):

HanaAyman (approved):

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -205,6 +205,7 @@ export function PeerEditPanel({id, onClose, onSave}: ImageEditPanelProps) {
                 {!!profile?.callToActionText && (
                   <RichTextBlock
                     disabled
+                    displayOnly
                     // TODO: remove this
                     onChange={console.log}
```
0815pascal, line 210: What about this `console.log` ? I don't quite see the use of it?

PR #181, opened 2021-02-11
Unique page slug

roosnic1:
Fixes WPC-5. This PR makes the slug field for published pages unique. The `publishPage` mutation now checks if there is another published page with the same slug. If so it returns an error.

### DIFF
--- a/packages/api/src/error.ts
+++ b/packages/api/src/error.ts
@@ -98,7 +98,7 @@ export class UserInputError extends ApolloError {
   }
 }
 
-export class DuplicatePageSlug extends ApolloError {
+export class DuplicatePageSlugError extends ApolloError {
   constructor(publishedPageID: string, slug: string) {
     super(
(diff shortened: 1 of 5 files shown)


penk85 (approved):

0815pascal (approved):

HanaAyman (approved):

packages/api/src/error.ts
```diff
@@ -96,3 +97,12 @@ export class UserInputError extends ApolloError {
     super(`User Input Error: \n${actualError}`, ErrorCode.UserInputError)
   }
 }
+
+export class DuplicatePageSlugError extends ApolloError {
```
bishoymelek, line 101: Is this class will be used a lot afterward?

bishoymelek (approved):

packages/api/src/error.ts
roosnic1: For this kind of error

PR #184, opened 2021-02-15
add ability to display only one line

bishoymelek:
Added displayOneLine prop to rich-text to display only a line depending on the width it has


.github/workflows/main.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-      - f/WEP-153-comments-listing-route
+      - f/display-one-line-rich-text
```
faebebin, line 7: Should this be in the PR?

packages/editor/src/client/routes/commentList.tsx
```diff
@@ -219,6 +219,7 @@ export function CommentList() {
                   {rowData?.revisions?.length ? (
                     <RichTextBlock
                       displayOnly
+                      displayOneLine
                       disabled
                       // TODO: remove this
                       onChange={console.log}
```
faebebin, line 225: Should probably remove this?

faebebin (changes requested):
Looks good. Two things that you might want to remove from PR?

roosnic1 (dismissed):
👍🏾

HanaAyman (dismissed):

packages/editor/src/client/routes/commentList.tsx
bishoymelek: @faebebin We added the option mainly now to use it in the comments list table, Do you suggest we move this line to another PR?

0815pascal: I guess @faebebin meant that you should remove the `onChange={console.log}`  because there's also a TODO which says "remove this"

0815pascal (commented):
Looks good

packages/editor/src/client/routes/commentList.tsx
faebebin: Sorry just noticed that is existing code :)

.github/workflows/main.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - f/WEP-153-comments-listing-route
```
faebebin, line 7: Think you should try: git checkout master .github/workflows/main.yml

faebebin (changes requested):

.github/workflows/main.yml
bishoymelek: @faebebin I guess you are right that no need to have in master another branch will be deployed anytime we merge anything to master

penk85 (approved):

roosnic1 (approved):
👍🏾

faebebin (approved):
Nice

PR #186, opened 2021-02-15
f/wpc-56 & wpc-91 richtext editor link button and URL validation

penk85:
- Link button with popup menu on the richtext editor. 
- Can add link by highlighting text in the editor or by inserting through the menu.
- URLs in links validation.


packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }, [])
+
+  return (
+    <>
+      <Form fluid>
+        <FormGroup>
+          <ControlLabel>{t('blocks.richText.link')}</ControlLabel>
+          <FormControl
+            errorMessage={url && !isValidURL ? 'Invalid Link' : undefined}
```
0815pascal, line 56: You also might want to translate `Invalid Link` if that's possible (I pretty sure it is, but not 100% sure).

0815pascal (changes requested):
Looks good, just a minor remark

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
penk85: Good point!

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const {closeMenu} = useContext(SubMenuContext)
+
+  const [title, setTitle] = useState('')
+  const [url, setURL] = useState('')
+
+  const [selection, setSelection] = useState<Range | null>(null)
+
+  const [isValidURL, setIsValidURL] = useState(false)
+  const isDisabled = !isValidURL
```
roosnic1, line 22: The button shouldn't be disabled if we can't validate the url. It might be that our validation failed or that the URL will only be active at a later date.

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
roosnic1, line 56: I suggest to work with the component https://rsuitejs.com/components/form/#Help%20Block. In my opinion it's less intrusive. Also the message should be more like "Could not validate link. Please check"

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const pattern = new RegExp(
+      '^(https?:\\/\\/)?' + // protocol
+      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
+      '((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
+      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
+      '(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
+        '(\\#[-a-z\\d_]*)?$',
+      'i'
+    )
+    if (!pattern.test(url)) return false
```
roosnic1, line 98: deepin-screen-recorder_Select area_20210217103534

My URL with `https` in the begging was invalid.

roosnic1 (changes requested):
Looks good but needs some improvements.

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
penk85: Should it be disabled if it doesn't pass the regex, or just remove the disabled part completely so there's always an option to add invalid links?

roosnic1: @penk85 I think it should be disable if there is no value in the link or text input. But if the link seems to be invalid it should only show the warning but allow the input.

roosnic1 (approved):
Looks awesome 👍🏾

HanaAyman (approved):

0815pascal (approved):

PR #188, opened 2021-02-16
check first if visitor is authenticated or not

bishoymelek:
For public comments list in an article, add a check first if the visitor is authenticated or not.


penk85 (approved):

0815pascal (approved):
Looks good to me 👍🏼

PR #191, opened 2021-02-16
wpc-220 login with enter

penk85:
Allows user to submit login credentials with enter key.


roosnic1 (approved):
WOW. It was that simple 😄

bishoymelek (approved):
:tada:

HanaAyman (approved):

PR #195, opened 2021-02-24
Wpc 205 character counter

HanaAyman:
There is a character counter that counts title, lead, top title, SEO title, SM title and description for SM.
There is also a maximum number of characters for these fields, if the publisher exceeds it a warning will appear. 
Pre-Title 30
Title 140
Lead 350
SEO-Title: 70 Characters
Title for Social-Media: 100 Characters
Description for Social-Media: 140 Characters
There is also a counter that counts the total number of characters in all the rich text blocks of the article and display the total at the bottom of the page.


packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -131,6 +137,10 @@ export function ArticleMetadataPanel({value, onClose, onChange}: ArticleMetadata
                     onChange?.({...value, socialMediaTitle})
                   }}
                 />
+                <div style={{textAlign: 'right'}}> {`${t("articleEditor.panels.charCount")} ${value.socialMediaTitle?.length}`} </div>
+                { value.socialMediaTitle!.length > socialMediaTitleMax && (
+                <div style={{textAlign: 'right', color: 'gold'}}>{`${t("articleEditor.panels.charCountWarning")} ${socialMediaTitleMax}`}</div>
```
penk85, line 142: Maybe instead of a div, it would be better to set the text corresponding to form input with one of the rsuite form labels or warnings.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -180,6 +180,30 @@ export function ArticleEditor({id}: ArticleEditorProps) {
...
+          blocksCount +
+          block.value.reduce(
+            (total: any, {children}: any) =>
+              total +
+              children.reduce((string: any, nodePair: any, index: any) => {
+                const {text} = nodePair
+                if (index === 0) return `${text}`
+                console.log(children);
+                // if (children){
+                //   return countRichTextBlocksChars(children);
```
roosnic1, line 197: this doesn't work because you expect `blocks` at the beginning and here you only have RichTextNode childs.

I would create two functions. The first one filters all blocks for RichText. Then you reduce those blocks. All those blocks have the value and that value will be the richText. Those nodes you pass to the second function which will then return the character count. But that second function will call itself again if it encounters children. This way you should be able to count all text.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -180,6 +180,30 @@ export function ArticleEditor({id}: ArticleEditorProps) {
     }
   }, [createError, updateError, publishError])
 
+  function countRichTextBlocksChars(blocks: any) {
+    return blocks
+      .filter((value: any) => value.type === BlockType.RichText)
```
roosnic1, line 185: Don't forget to count the Text length of the `BlockType.Title`. The title and lead also count to that number

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -389,6 +415,7 @@ export function ArticleEditor({id}: ArticleEditorProps) {
           }}
         />
       </Modal>
+      <div style={{textAlign: 'right', padding: '20px'}}> {`${t("articleEditor.panels.totalCharCount")} ${totalCharCount}`} </div>
```
roosnic1, line 418: Don't show the character count here. Show it in the metadata panel before the pre title.

roosnic1 (changes requested):
looks good but still needs some improvements

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -180,6 +180,30 @@ export function ArticleEditor({id}: ArticleEditorProps) {
     }
   }, [createError, updateError, publishError])
 
+  function countRichTextBlocksChars(blocks: any) {
```
penk85, line 183: You could probably set types here more specific than any.

packages/editor/src/client/locales/en.json
```diff
@@ -125,7 +125,13 @@
         "postImage": "Image for Post",
         "metadataInfo": "Information from these fields are being used for previews in Social Media Posts. If left blank, information from general-tab will be used instead.",
         "lightStyledWithNoImageError": "Light styled teasers must have an image",
+        "titleHelpBlock": "Please use an SEO-optimised title",
+        "leadHelpBlock": "Please use an SEO-optimised lead",
+        "charCount": "Characters count {{charCount}}",
+        "charCountWarning": "The maximum number of characters should be",
+        "totalCharCount": "Total characters count {{charCount}}",
         "seoTitleHelpBlock": "Please use an SEO-optimised title"
```
penk85, line 133: Can this be removed?

packages/editor/src/client/routes/articleEditor.tsx
faebebin: Yes need to split in 2 functions for iterating blocks and richText seperately. This seems to work also with tables and lists in the block:
```

  function countRichtextChars(blocksCharLength: number, nodes: any) {
    return nodes.reduce((charLength: number, node: any) => {
      if (!node.text && !node.children) return charLength
      // node either has text (leaf node) or children (element node)
      if (!!node.text) {
        return charLength + (node.text as string).length
      }
      return countRichtextChars(charLength, node.children)
    }, blocksCharLength)
  }

  function countRichTextBlocksChars(blocks: BlockValue[]) {
    return blocks.reduce((charLength: number, block: BlockValue) => {
      if (!(block.type === BlockType.RichText)) return charLength
      return countRichtextChars(charLength, block.value)
    }, 0)
  }

```
You still need to test it though with including all kinds of elements (tables, lists, etc) emoji's etc.

Further, @roosnic1  for having this counting while editing the page it seems quite expensive, to iterate all blocks and contents on every letter. Maybe we should rather have a global state.charCount object with memory, where every change does `charCount =+/- 1` ? Maybe some page wide context. Or some package that does it in an optimized way?

faebebin (changes requested):

packages/editor/src/client/locales/en.json
HanaAyman: Why ? It is used in the metaDataPanel of the article.

0815pascal (dismissed):
Looks good to me and also works when testing locally

packages/editor/src/client/routes/articleEditor.tsx
roosnic1: @faebebin you're right about this being very expensive.

@HanaAyman since you only show the characters count inside the metadata you can execute the counting before opening the metadata panel.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -319,7 +320,15 @@ export function ArticleEditor({id}: ArticleEditorProps) {
     }
   }, [isNotFound])
 
-  const totalCharCount = countRichTextBlocksChars(blocks);
+  let totalCharCount = 0;
```
roosnic1, line 323: you don't need this. you can set the count directly in the `useEffect` and don't need to save it first in the variable.

roosnic1 (changes requested):

packages/editor/src/client/locales/de.json
```diff
@@ -117,7 +117,10 @@
         "metadataInfo": "Information from these fields are being used for previews in Social Media Posts. If left blank, information from general-tab will be used instead.",
         "lightStyledWithNoImageError": "Light styled teasers must have an image",
         "titleHelpBlock": "Please use an SEO-optimised title",
-        "leadHelpBlock": "Please use an SEO-optimised lead"
+        "leadHelpBlock": "Please use an SEO-optimised lead",
+        "charCount": "Characters count {{charCount}}",
+        "charCountWarning": "The maximum number of characters should be",
```
bishoymelek, line 122: Maybe we can have a variable here as well?

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -210,6 +245,10 @@ export function ArticleMetadataPanel({value, onClose, onChange}: ArticleMetadata
                   onChange={seoTitle => onChange?.({...value, seoTitle})}
                 />
                 <HelpBlock>{t('articleEditor.panels.seoTitleHelpBlock')}</HelpBlock>
+                <label style={{textAlign: 'right'}}> {`${t("articleEditor.panels.charCount")} ${value.seoTitle.length}`} </label>
```
bishoymelek, line 248: I guess we can have it like it's described here: https://www.i18next.com/translation-function/interpolation
to make use of the variable you added in the locale file ;-)

bishoymelek (changes requested):
Great Work

packages/editor/src/client/panel/articleMetadataPanel.tsx
HanaAyman: I actually removed this part (${t("articleEditor.panels.charCount")}) as we only need to view the number without a text beside it.

bishoymelek: I think it will be helpful? @HanaAyman  @roosnic1 what do you think?

bishoymelek (dismissed):
:tada:

0815pascal (dismissed):

roosnic1 (dismissed):
👍🏾

packages/editor/src/client/panel/articleMetadataPanel.tsx
roosnic1: @bishoymelek it think it's pretty clear what it is and we can omit the text.

roosnic1 (approved):
🎉

roosnic1 (approved):

PR #205, opened 2021-03-04
Add hotkey shortcuts to Richtext Editor

penk85:
Adds keyboard shortcuts for formatting inside the richtext editor. 
ctrl + b (bold)
ctrl + i (italic)
ctrl + u (underlined)
ctrl + shift + x (strikethrough)


packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -141,6 +168,9 @@ export const RichTextBlock = memo(function RichTextBlock({
         onBlur={() => {
           setLocation(editor.selection)
         }}
+        onKeyDown={e => {
+          if (e.ctrlKey) activateHotkey(e)
```
bishoymelek, line 172: Can we have e.preventDefault() here? what do you think?

bishoymelek (changes requested):
Great work, just a small point

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
penk85: Unfortunately not because it would disable typing inside the editor. Even placing it inside the activateHotkey method, but outside of the switch statement, would disable other shortcuts like copy/paste and I haven't thought of any obvious way to avoid the repetitiveness.

bishoymelek (approved):
:tada:

HanaAyman (approved):

roosnic1 (approved):
🎉

PR #206, opened 2021-03-05
B/wpc 245 image selection disabled in edit peer

HanaAyman:
This is a bug fix to remove the plus sign for the image in the edit peer panel


packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -193,7 +193,7 @@ export function PeerEditPanel({id, onClose, onSave}: ImageEditPanelProps) {
         </Panel>
         {!isLoadingPeerProfile && isValidURL && (
           <Panel header={t('peerList.panels.information')}>
-            <ChooseEditImage disabled image={profile?.logo} />
+            {(profile?.logo )?(<ChooseEditImage disabled image={profile?.logo} />):(<ControlLabel>{t('chooseEditImage.header')}<Placeholder.Graph/> </ControlLabel>)}
```
roosnic1, line 196: I like your solution but i think it should be inside the `ChooseEditImage` component. Since it have the `disabled` flag you can show the placeholder whenever `disabled==true` and `image` being undefined.

roosnic1 (changes requested):

roosnic1 (approved):
👍🏾

PR #209, opened 2021-03-10
F/wpc 244 infer link protocol

penk85:
Add dropdown list to select prefix to link inside richtext

### DIFF
--- a/packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
+++ b/packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
@@ -24,8 +24,14 @@ export function LinkMenu() {
   const [title, setTitle] = useState('')
   const [url, setURL] = useState('')
 
-  //TODO define the protocol option types
-  const [protocol, setProtocol] = useState('')
+  enum prefixType {
+    http = 'http://',
+    https = 'https://',
+    mailto = 'mailto://',
+    other = 'other'
+  }
+
+  const [prefix, setPrefix] = useState<prefixType | string>(prefixType.http)
 
   const [selection, setSelection] = useState<Range | null>(null)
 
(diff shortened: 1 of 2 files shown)


packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
```diff
@@ -39,7 +68,22 @@ export function LinkMenu() {
...
+      } else if (nodeUrl.startsWith(prefixType.http)) {
+        setPrefix(prefixType.http)
+        nodeUrl = nodeUrl.replace(prefixType.http, '')
+      } else if (nodeUrl.startsWith(prefixType.mailto)) {
+        setPrefix(prefixType.mailto)
+        nodeUrl = nodeUrl.replace(prefixType.mailto, '')
+      } else {
+        setPrefix(prefixType.other)
+      }
+      setURL((nodeUrl as string) ?? '')
```
roosnic1, line 86: you shouldn't need the code above till line 71 since you're setting the URL and then the `useEffect` above on line 45 should do all the magic.

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
```diff
@@ -51,7 +95,21 @@ export function LinkMenu() {
       <Form fluid>
         <FormGroup>
           <ControlLabel>{t('blocks.richText.link')}</ControlLabel>
-          <FormControl value={url} onChange={url => setURL(url)} />
+          <InputGroup>
+            <select
+              style={{backgroundColor: 'white', border: 'none', boxShadow: 'none'}}
+              value={prefix}
+              onChange={e => {
+                setPrefix(e.target.value !== 'other' ? e.target.value : prefixType.other)
```
roosnic1, line 103: why do you need this check?

roosnic1 (changes requested):
looks good just some little change requests

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
penk85: Actually good point. I originally had it to put in a blank string, but now its doing nothing.

packages/editor/src/client/blocks/richTextBlock/toolbar/linkMenu.tsx
penk85: Ok but I still need to check for it being other because otherwise it defaults to http

roosnic1 (approved):
🎉

bishoymelek (approved):
Great job

HanaAyman (approved):

PR #210, opened 2021-03-12
b/Wpc-229-make-duplicated-articles-appear-immediatly

HanaAyman:
This pull request is to make the duplicated articles and pages appear immediately without needing to refresh the page and also they will appear highlighted for a few seconds.


packages/editor/src/client/routes/theme.less
```diff
@@ -0,0 +1,3 @@
+.duplicated-row {
+    color: #ff4545 ;
```
roosnic1, line 2: I suggest you use `background` and maybe also use something like this `transition: background-color 0.3s linear;` to make it more smooth. I copied this from the `:hover` css.
image

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -76,6 +78,18 @@ export function ArticleList() {
     refetch(articleListVariables)
   }, [filter, page, limit, sortOrder, sortField])
 
+  const [duplicatedDataId, setDuplicatedDataId]: any = useState()
```
roosnic1, line 81: I suggest you rename this to `highlightedRowId`. Pretty sure we can use this for other features. 

You don't need the `any`. But you should set the state `useState<string | null>(null)` this way you can set it to null if there should no highlighted row.

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -76,6 +78,18 @@ export function ArticleList() {
...
+  let output;
```
roosnic1, line 82: Why do you initialize `output` here? You only need it where you actually duplicate the article.

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -122,6 +136,10 @@ export function ArticleList() {
           data={articles}
           sortColumn={sortField}
           sortType={sortOrder}
+          rowClassName={rowData => {
+            if (rowData && rowData.id === duplicatedDataId) return 'duplicated-row'
```
roosnic1, line 140: Another way to write this would be:
`return rowData?.id === duplicatedDataId ? 'duplicated-row' : ''`

packages/editor/src/client/routes/theme.less
```diff
@@ -0,0 +1,3 @@
+.duplicated-row {
```
roosnic1, line 1: I would rename this file to `global.less` and move it a directory up.

packages/editor/src/client/routes/pageList.tsx
```diff
@@ -71,6 +74,20 @@ export function PageList() {
     fetchPolicy: 'network-only'
   })
 
+  const [duplicatedDataId, setDuplicatedDataId]: any = useState()
```
roosnic1, line 77: same as above

packages/editor/src/client/routes/pageList.tsx
```diff
@@ -71,6 +74,20 @@ export function PageList() {
...
+
+  let output
```
roosnic1, line 79: same as above

packages/editor/src/client/routes/pageList.tsx
```diff
@@ -116,6 +133,10 @@ export function PageList() {
           data={pages}
           sortColumn={sortField}
           sortType={sortOrder}
+          rowClassName={rowData => {
+            if (rowData && rowData.id === duplicatedDataId) return 'duplicated-row'
```
roosnic1, line 137: same as above

roosnic1 (changes requested):
looks very good. just some cosmetic changes.

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -16,7 +16,7 @@ import {
 
 import {DescriptionList, DescriptionListItem} from '../atoms/descriptionList'
 
-import './theme.less'
+import '../global.less'
```
roosnic1, line 19: I'm not sure if this import works. The file seems to be one directory up. But it would anyway be better if you import the `global.less` just once in the `app.tsx`

roosnic1, line 19: I'm not sure if this import works. The file seems to be one directory up. But it would anyway be better if you import the `global.less` just once in the `app.tsx`

packages/editor/src/client/global.less
```diff
@@ -0,0 +1,4 @@
+.highlighted-row .rs-table-cell-group .rs-table-cell{
+    background-color: #f2ffff;
+    transition: background-color 0.3s linear;
```
roosnic1, line 3: you don't need the transition here since it's already set through the CSS from rsuite

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -355,7 +353,7 @@ export function ArticleList() {
                   break
 
                 case ConfirmAction.Duplicate:
-                  output = await duplicateArticle({
+                  await duplicateArticle({
```
roosnic1, line 356: i think you don't need the await if you have the structure `duplicateArticle.then()`. The await is needed when you want it in the variable like we did it before.

packages/editor/src/client/routes/pageList.tsx
```diff
@@ -19,7 +19,7 @@ import {FlexboxGrid, Input, InputGroup, Icon, Table, IconButton, Modal, Button}
 import {DescriptionList, DescriptionListItem} from '../atoms/descriptionList'
 import {DEFAULT_TABLE_PAGE_SIZES, mapTableSortTypeToGraphQLSortOrder} from '../utility'
 
-import './theme.less'
+import '../global.less'
```
roosnic1, line 22: same as above

packages/editor/src/client/routes/pageList.tsx
```diff
@@ -316,7 +313,7 @@ export function PageList() {
                   break
 
                 case ConfirmAction.Duplicate:
-                  output = await duplicatePage({
+                  await duplicatePage({
```
roosnic1, line 316: same as above

roosnic1 (changes requested):
just some last changes

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -122,6 +133,9 @@ export function ArticleList() {
           data={articles}
           sortColumn={sortField}
           sortType={sortOrder}
+          rowClassName={rowData => {
+            return rowData?.id === highlightedRowId ? 'highlighted-row' : ''
```
bishoymelek, line 137: It's different code styles but can we remove the return keyword and parentheses? What do you think?

packages/editor/src/client/routes/articleList.tsx
```diff
@@ -337,8 +351,32 @@ export function ArticleList() {
                   break
 
                 case ConfirmAction.Duplicate:
-                  await duplicateArticle({
```
bishoymelek, line 340: I guess the await is important here?

bishoymelek (commented):
Great job, just small points

packages/editor/src/client/routes/articleList.tsx
roosnic1: Not needed anymore because it's a Promise with a then block

packages/editor/src/client/routes/articleList.tsx
HanaAyman: Yes, that will work too 👍

roosnic1 (approved):
🎉

bishoymelek (approved):

PR #212, opened 2021-03-15
fixed links in example website

roosnic1:
WPC-257. Renders links in the example website correctly again.

### DIFF
--- a/examples/website/src/shared/blocks/richTextBlock/editor/render.tsx
+++ b/examples/website/src/shared/blocks/richTextBlock/editor/render.tsx
@@ -53,7 +53,12 @@ export function renderElement({attributes, children, element}: RenderElementProp
       // title={title}
 
       return (
-        <a data-title={element.title} data-href={element.url} {...attributes}>
+        <a
+          target="_blank"
+          rel="noreferrer"
+          href={element.url as string}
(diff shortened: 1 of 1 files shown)


penk85 (approved):

examples/website/src/shared/blocks/richTextBlock/editor/render.tsx
```diff
@@ -53,7 +53,12 @@ export function renderElement({attributes, children, element}: RenderElementProp
       // title={title}
 
       return (
-        <a data-title={element.title} data-href={element.url} {...attributes}>
+        <a
+          target="_blank"
+          rel="noreferrer"
+          href={element.url as string}
```
bishoymelek, line 59: Is that change because of the change in the npm module we use?

bishoymelek (changes requested):

examples/website/src/shared/blocks/richTextBlock/editor/render.tsx
roosnic1: it's because Pascal replaced the original richText render block with the richTextEditor block from the edtior package

bishoymelek (approved):

PR #214, opened 2021-03-17
f/wpc-256-see-the-max-amount-of-characters

HanaAyman:
This pull request is for the publisher to be able to see the maximum amount of characters for each field in the metadata of the article.


packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -137,21 +137,23 @@ export function ArticleMetadataPanel({value, infoData, onClose, onChange}: Artic
               </FormGroup>
               <FormGroup>
                 <ControlLabel>{t('articleEditor.panels.socialMediaTitle')}
-                <label style={{float: 'right'}}> {value.socialMediaTitle?.length}</label>
+                {socialMediaTitle ? <label style={{float: 'right'}}> {value.socialMediaTitle?.length}/{socialMediaTitleMax}</label> :
```
roosnic1, line 140: Another way to write this would be:

`<label style={{float: 'right'}}> {value.socialMediaTitle ? value.socialMediaTitle.length : 0}/{socialMediaTitleMax}</label>`

Makes the whole thing a bit more readable.

roosnic1 (changes requested):

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -206,7 +206,7 @@ export function ArticleMetadataPanel({value, infoData, onClose, onChange}: Artic
             <div style={{paddingBottom: '20px'}}>{t("articleEditor.panels.totalCharCount",{totalCharCount: infoData.charCount})}</div>
               <FormGroup>
                 <ControlLabel>{t('articleEditor.panels.preTitle')}
-                <label style={{float: 'right'}}> {`${value.preTitle.length}`} </label>
+                <label style={{float: 'right'}}> {`${value.preTitle.length}`}/{preTitleMax} </label>
```
bishoymelek, line 209: @HanaAyman Correct me if I am wrong, I guess you are having the length label in different styles, so this one is different than on line 140 and also 164 and 221

bishoymelek (changes requested):
Good Work, Only one point.

roosnic1 (approved):
👍🏾

bishoymelek (approved):
Good Job  :tada:

penk85 (approved):

PR #220, opened 2021-03-19
Remove defunct link from green heart emoji

valioDOTch:
Removing defunct link from heart emoji
Uncertain what the intention for that link was, but it's kinda irritating for a visitor, as it leads to a 404 page


roosnic1:
@valioDOTch Thanks :)

PR #221, opened 2021-03-23
WPC-227 Add unsaved changes dialog to panels

penk85:


roosnic1:
will be tackled in a new PR soon

PR #223, opened 2021-03-24
fix table column width

penk85:
WPC-289: fix column width: In Article List extended the default width of the States column so that content is displayed.


penk85:
Note: Line 173 is the only change. Others are changes by Prettier.

HanaAyman (approved):

bishoymelek (approved):
:tada:

roosnic1 (approved):
🥳

PR #224, opened 2021-03-24
f/wpc-255-character-counter-count-more-blocks

HanaAyman:


bishoymelek (approved):
:tada:

penk85 (approved):

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -191,17 +195,51 @@ export function ArticleEditor({id}: ArticleEditorProps) {
...
+
+    const richTextBlocksCount = richTextBlocks.reduce(
+      (charCount: number, item) => charCount + countRichtextChars(0, item),
+      0
+    )
+
+    return totalTitleChars + richTextBlocksCount
+  }
+
+  function WordCounter(blocks: BlockValue[]): number {
```
roosnic1, line 229: simple functions usually don't start with a capital letter.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -191,17 +195,51 @@ export function ArticleEditor({id}: ArticleEditorProps) {
...
+    return countRichtextChars(0, block.value)
+  }
+
+  function countTitleChars(block: TitleBlockListValue): number {
+    return (block.value.title as string).length + (block.value.lead as string).length
   }
 
-  function countTitle(blocks: BlockValue[]) {
+  function countQuoteChars(block: QuoteBlockListValue) {
+    return (block.value.quote as string).length + (block.value.author as string).length
```
roosnic1, line 207: I think the `as string` should not be necessary here. I think quote and author can be `string | undefined` and you need to check first if they are undefined.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -191,17 +195,51 @@ export function ArticleEditor({id}: ArticleEditorProps) {
...
-    return blocks.reduce((charLength: number, block: BlockValue) => {
-      if (!(block.type === BlockType.RichText)) return charLength
-      return countRichtextChars(charLength, block.value)
-    }, 0)
+  function countRichTextBlocksChars(block: RichTextBlockListValue) {
+    return countRichtextChars(0, block.value)
+  }
+
+  function countTitleChars(block: TitleBlockListValue): number {
+    return (block.value.title as string).length + (block.value.lead as string).length
```
roosnic1, line 203: I think the `as string` should not be necessary here. I think quote and author can be `string | undefined` and you need to check first if they are undefined.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -325,8 +363,10 @@ export function ArticleEditor({id}: ArticleEditorProps) {
   })
 
   useEffect(() => {
-    setInfoData({charCount: countRichTextBlocksChars(blocks) + countTitle(blocks)})
-  },[isMetaDrawerOpen])
+    setInfoData({
+      charCount: 0 + WordCounter(blocks)
```
roosnic1, line 367: why the `0 + `

roosnic1 (changes requested):

packages/editor/src/client/routes/articleEditor.tsx
HanaAyman: I think the quote and author are of type string only so they are never undefined, so I don't think the check is needed.
```
export interface QuoteBlockValue {
  quote: string
  author: string
}
```

roosnic1 (approved):
🎉

penk85 (approved):

PR #227, opened 2021-03-25
f/wpc-266-add-getCommentUrl-in-UrlAdapter-interface

HanaAyman:


packages/api/__tests__/utility.ts
```diff
@@ -37,6 +39,13 @@ class ExampleURLAdapter implements URLAdapter {
   getArticlePreviewURL(token: string): string {
     return `https://demo.wepulish.ch/article/preview/${token}`
   }
+
+  getCommentURL(item: PublicArticle | PublicPage, comment: PublicComment): string {
+    if (comment.itemType === CommentItemType.Article) {
+      return `https://editor.demo.wepublish.media/comments/a/${item.id}/${item.slug}/${comment.id}`
```
roosnic1, line 45: I know this is only a test but the URL would still not be on the editor. It would be more something like `https://demo.wepublish.media/...`

packages/api/__tests__/utility.ts
```diff
@@ -37,6 +39,13 @@ class ExampleURLAdapter implements URLAdapter {
...
+    }
+    return `https://editor.demo.wepublish.media/comments/${item.slug}/${comment.id}`
```
roosnic1, line 47: same here.

roosnic1 (changes requested):
looks good just some tiny changes

roosnic1 (approved):
🦖

bishoymelek (approved):
:tada:

PR #228, opened 2021-03-26
use Visibility API to check if still logged in

bishoymelek:


bishoymelek:
@roosnic1 there is no types package for `react-page-visibility` so I am using // @ts-ignore but linting won't passing it. what do you think?

HanaAyman (approved):

PR #230, opened 2021-03-31
split time and date picker

penk85:
WPC-321: Date picker on publish page and publish article panels split into separate pickers for time and dates.


packages/editor/src/client/locales/en.json
```diff
@@ -55,7 +55,8 @@
         "draft": "Draft",
         "pending": "Pending",
         "published": "Published",
-        "publishDate": "Publish Date",
+        "publishTime": "Publish Time",
```
roosnic1, line 58: why did you rename this? It's still `Publish Date & Time`

packages/editor/src/client/locales/en.json
```diff
@@ -83,7 +84,7 @@
         "confirm": "Confirm",
         "articlePending": "There is already a pending publication scheduled at {{pendingPublishDate, MM/DD/YYYY HH:mm}}. Publishing again will override that publication. ",
         "invalidDate": "Invalid Date",
-        "updateDate": "Update Date",
+        "updateTime": "Update Time",
```
roosnic1, line 87: same here

packages/editor/src/client/locales/en.json
```diff
@@ -241,9 +242,10 @@
         "publishPage": "Publish Page",
         "confirm": "Confirm",
         "pagePending": "There is already a pending publication scheduled at {{pendingPublishDate, MM/DD/YYYY HH:mm}}. Publishing again will override that publication. ",
-        "publishDate": "Publish Date",
+        "publishDate": "Publish Time",
```
roosnic1, line 245: same here

packages/editor/src/client/locales/en.json
```diff
@@ -241,9 +242,10 @@
...
+        "now": "Now",
         "invalidDate": "Invalid Date",
-        "updateDate": "Update Date",
+        "updateTime": "Update Time",
```
roosnic1, line 248: same here

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -45,29 +45,56 @@ export function PublishArticlePanel({
         )}
         <Form fluid={true}>
           <FormGroup>
-            <ControlLabel>{t('articleEditor.panels.publishDate')}</ControlLabel>
+            <ControlLabel>{t('articleEditor.panels.publishTime')}</ControlLabel>
```
roosnic1, line 48: Please set the date field in the first position. Users will more likely change the date then the time

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -45,29 +45,56 @@ export function PublishArticlePanel({
...
+              placement="auto"
+              value={publishDate}
+              format="DD MMM YY"
+              ranges={[
+                {
+                  label: t('articleEditor.panels.now'),
                   value: new Date()
                 }
               ]}
               onChange={publishDate => setPublishDate(publishDate)}
```
roosnic1, line 73: If i change the date after i changed the time the time will be reset to the current time. So you have to improve the `setPublishDate(...)`  function to include the time of the already set date.

This link might help. https://www.digitalocean.com/community/tutorials/understanding-date-and-time-in-javascript

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -45,29 +45,56 @@ export function PublishArticlePanel({
...
-            <ControlLabel>{t('articleEditor.panels.publishDate')}</ControlLabel>
+            <ControlLabel>{t('articleEditor.panels.publishTime')}</ControlLabel>
             <DatePicker
-              block
+              style={{width: 100, marginRight: 8}}
+              placement="auto"
               value={publishDate}
-              format="YYYY-MM-DD HH:mm"
+              format="HH:mm"
               ranges={[
```
roosnic1, line 54: for time please add the following ranges:
- now
- 6 am
- 2 pm

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -45,29 +45,56 @@ export function PublishArticlePanel({
...
+                }
+              ]}
+              onChange={publishDate => setPublishDate(publishDate)}
+            />
+            <DatePicker
+              style={{width: 130}}
+              placement="auto"
+              value={publishDate}
+              format="DD MMM YY"
+              ranges={[
```
roosnic1, line 67: for date please add the following ranges:
- today
- tomorrow
- next saturday

roosnic1 (changes requested):
Looks good but still needs some improvement. Also the two `<DatePicker />` components look like they could be in their own component with all the logic.

packages/editor/src/client/locales/en.json
penk85: Was a visual consideration as well as placing the time picker before the date, but if the more common usage is to set the date then I'll change them both back.

packages/editor/src/client/panel/publishArticlePanel.tsx
penk85: @roosnic1 can you have a look again to see if the time and dates are behaving separately? PR still needs some small changes, but would appreciate if you can check my approach meanwhile.

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+              value: () => {
+                const tomorrow = new Date()
+                tomorrow.setDate(new Date().getDate() + 1)
+                return tomorrow
+              }
+            },
+            {
+              label: t('timeDatePicker.nextSaturday'),
+              value: () => {
+                const nextSaturday = new Date()
```
roosnic1, line 39: i'm not sure if this works. Maybe this article helps: https://codereview.stackexchange.com/questions/33527/find-next-occurring-friday-or-any-dayofweek

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <Form fluid={true}>
+      <FormGroup>
+        <ControlLabel>{label}</ControlLabel>
+        <DatePicker
+          style={{marginRight: 8}}
+          placement="auto"
+          value={dateTime?.getTime() === dateTime?.getTime() ? dateTime : undefined}
+          format="DD MMM YYYY"
+          ranges={[
```
roosnic1, line 23: I would make it possible to add its own ranges in the `DateTimeProps` but make it optional. If no ranges have been provider we use those as defaults. But make sure that the array passed in `DateTimeProps` is also types.

```
export interface DateTimeRange {
  label: string
  value: (): void
}

export interface DateTimeProps {
  ...
  dayRanges?: DateTimeRange[]
  timeRanges?: DatetimeRange[]
 ...
}
```

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react'
+import {DatePicker, FormGroup, ControlLabel, Form} from 'rsuite'
+import {useTranslation} from 'react-i18next'
+
+export interface DateTimeProps {
```
roosnic1, line 5: I would rename it to `DateTimePickerProps`. It's the properties for the `DateTimePicker` component

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const {t} = useTranslation()
+
+  return (
+    <Form fluid={true}>
+      <FormGroup>
+        <ControlLabel>{label}</ControlLabel>
+        <DatePicker
+          style={{marginRight: 8}}
+          placement="auto"
+          value={dateTime?.getTime() === dateTime?.getTime() ? dateTime : undefined}
```
roosnic1, line 21: i don't understand what you do here?

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -30,6 +31,11 @@ export function PublishArticlePanel({
 
   const {t} = useTranslation()
 
+  useEffect(() => {
```
roosnic1, line 34: please remove

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react'
+import {DatePicker, FormGroup, ControlLabel, Form} from 'rsuite'
+import {useTranslation} from 'react-i18next'
+
+export interface DateTimeProps {
+  dateTime?: Date
+  label: string
+  changeDate(publishDate?: Date): void
```
roosnic1, line 8: don't make `publishDate` optional. If you don't have a date you don't call the function `changeDate`

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -43,65 +49,16 @@ export function PublishArticlePanel({
...
-                }
-              ]}
-              onChange={updateDate => setUpdateDate(updateDate)}
-            />
-          </FormGroup>
-        </Form>
+        <DateTimePicker
+          dateTime={publishDate}
+          label={t('articleEditor.panels.publishDate')}
+          changeDate={(date: Date) => setPublishDate(date)}
```
roosnic1, line 55: if you made `publishDate` mandatory above you should not need to type Date here.

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -43,65 +49,16 @@ export function PublishArticlePanel({
...
+        />
+        <DateTimePicker
+          dateTime={updateDate}
+          label={t('articleEditor.panels.updateDate')}
+          changeDate={(date: Date) => setUpdateDate(date)}
```
roosnic1, line 60: if you made `publishDate` mandatory above you should not need to type Date here.

packages/editor/src/client/panel/publishPagePanel.tsx
```diff
@@ -44,65 +45,16 @@ export function PublishPagePanel({
...
-                }
-              ]}
-              onChange={updateDate => setUpdateDate(updateDate)}
-            />
-          </FormGroup>
-        </Form>
+        <DateTimePicker
+          dateTime={publishDate ?? now}
+          label={t('articleEditor.panels.publishDate')}
+          changeDate={(date: Date) => setPublishDate(date)}
```
roosnic1, line 51: if you made `publishDate` mandatory above you should not need to type Date here.

packages/editor/src/client/panel/publishPagePanel.tsx
```diff
@@ -44,65 +45,16 @@ export function PublishPagePanel({
...
+        />
+        <DateTimePicker
+          dateTime={updateDate ?? now}
+          label={t('articleEditor.panels.updateDate')}
+          changeDate={date => setUpdateDate(date)}
```
roosnic1, line 56: if you made `publishDate` mandatory above you should not need to type Date here.

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+              value: () => {
+                const nextSaturday = new Date()
+                const i = 6 - new Date().getDay()
+                nextSaturday.setDate(i ? nextSaturday.getDate() + i : nextSaturday.getDate() + 7)
+                return nextSaturday
+              }
+            }
+          ]}
+          onChange={value => {
+            if (dateTime) {
```
roosnic1, line 47: I think if you click the x inside the date/time picker you get here `value === null` and you don't want to set that. I think you can just remove the x from the pickers but maybe you still need to check here if `value` is defined.

roosnic1 (changes requested):
looks very good. Just needs some tiny improvements

packages/editor/src/client/atoms/dateTimePicker.tsx
penk85: This is an attempt to only display the date/time if it's a valid date. It prevents an invalid date getting set when user clears the picker, but it's not working great.

packages/editor/src/client/atoms/dateTimePicker.tsx
penk85: `i` is remaining days till Saturday. I add them to the current date, unless it's already Saturday and add 7 instead. I think that logic works? I tested with a few different dates in the console, and looks right. I think the solution you linked is essentially the same thing, but with any day besides Saturday you need to handle a negative result for remaining days  so add the `+7` and `%`.

packages/editor/src/client/atoms/dateTimePicker.tsx
penk85: If I check for `value`, when it's null it disables the activity on the x button rather than clearing the form and displaying placeholder like it did before.

packages/editor/src/client/atoms/dateTimePicker.tsx
bishoymelek: @penk85 you mean you are checking if it's a valid date? I suggest we have a util method and use it here
like:
`function isValidDate(d) {
  return d instanceof Date && !isNaN(d);
}
`
and use it like `value={isValidDate(dateTime) ?  dateTime : undefined}` 

But I think that rsuite already handling errors like these?

roosnic1 (approved):
👍🏾

packages/editor/src/client/panel/publishArticlePanel.tsx
bishoymelek: @roosnic1 why?

roosnic1: because if the type is mandatory in the interface TS knows that it can only be a Date. If the property is not mandatory it could be undefined | Date.

packages/editor/src/client/atoms/dateTimePicker.tsx
penk85: @bishoymelek the way the component is set up, it sets only time or date in the value which creates an error in the form. We opted to disable clearing the form since anyhow there's no case here for an optional date.

bishoymelek (approved):
:tada: Great work @penk85

HanaAyman (approved):

PR #235, opened 2021-04-03
F/wpc 158 add font color option to rich text block

bishoymelek:
Added Color picker in Rich text editor where you can select specific text and change the font color
- Note, we had to make changes to API interface as well.

Screenshot:
Screenshot from 2021-04-06 13-47-30

To make use of it in your website: for the `examples/website` will need to make following changes:

examples/website/src/shared/blocks/richTextBlock/editor/render.tsx: line 115

` 
If (leaf[TextFormat.Color]) {
    children = <span style={{color: leaf.color as string}}>{children}</span>
  }
`

examples/website/src/shared/blocks/richTextBlock/editor/formats.ts: line 25

`
Color = 'color'
 `


packages/editor/src/client/blocks/richTextBlock/toolbar/buttons.tsx
```diff
@@ -50,6 +51,18 @@ export function FormatButton({format, children}: FormatBlockButtonProps) {
   )
 }
 
+export function FormatFontColor() {
+  const editor = useSlate()
+
+  return (
+    <ToolbarButton>
```
penk85, line 58: Can you try putting this in a submenu consistent with table, links, and emojis?

penk85 (changes requested):
Looks good and works smoothly. Although I'd tr

packages/api/src/graphql/richText.ts
```diff
@@ -177,7 +179,8 @@ export function parseRichTextNode(value: unknown, path: string[] = []): RichText
       value.underline != undefined ? {underline: value.underline as boolean} : {},
       value.strikethrough != undefined ? {strikethrough: value.strikethrough as boolean} : {},
       value.superscript != undefined ? {superscript: value.superscript as boolean} : {},
-      value.subscript != undefined ? {subscript: value.subscript as boolean} : {}
+      value.subscript != undefined ? {subscript: value.subscript as boolean} : {},
+      value.fontColor != undefined ? {fontColor: value.fontColor as string} : {}
```
roosnic1, line 183: I would suggest to use `color` instead of `fontColor`. I think this is more inline with the CSS property.

packages/editor/src/client/blocks/richTextBlock/toolbar/buttons.tsx
roosnic1: I agree. 👍🏾

roosnic1 (changes requested):

packages/editor/src/client/blocks/richTextBlock/toolbar/buttons.tsx
bishoymelek: @penk85 @roosnic1 Changed :+1: 
Screenshot from 2021-04-06 13-47-30

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -134,6 +133,12 @@ export const RichTextBlock = memo(function RichTextBlock({
 
             <ToolbarDivider />
 
+            <EditorSubMenuButton icon="font" editorHasFocus={hasFocus}>
+              <FormatColor />
```
penk85, line 137: Can you set the whole modal inside the submenu instead of the button?

penk85 (changes requested):
Can you set the whole color change modal inside the submenu?

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
bishoymelek: @penk85 I made some changes, which should let the color picker menu stay open even after clicking on the palette. As well changing the Icon and used references.

PR #238, opened 2021-04-06
F/wpc 323 sorting pagination author list

bishoymelek:
Now we have proper pagination working and sorting with the creation date

### DIFF
--- a/packages/editor/src/client/routes/authorList.tsx
+++ b/packages/editor/src/client/routes/authorList.tsx
@@ -165,7 +165,7 @@ export function AuthorList() {
               )}
             </Cell>
           </Column>
-          <Column width={200} align="left" resizable>
+          <Column width={300} align="left" resizable>
             <HeaderCell>{t('authors.overview.name')}</HeaderCell>
             <Cell>
               {(rowData: FullAuthorFragment) => (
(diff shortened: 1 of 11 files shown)


penk85 (dismissed):

packages/editor/src/client/routes/authorList.tsx
```diff
@@ -52,7 +71,10 @@ export function AuthorList() {
 
   const authorListQueryVariables = {
     filter: filter || undefined,
-    first: 50
+    skip: page - 1,
+    sort: mapColumFieldToGraphQLField(sortField),
+    first: 50,
```
roosnic1, line 76: Pagination doesn't work if you're always getting 50 articles. I think you have to use `limit` here.

roosnic1 (changes requested):
Looks good. Just one little change

packages/editor/src/client/routes/authorList.tsx
bishoymelek: @roosnic1  Can you explain?

roosnic1: Add more than 10 authors and test if pagination works with limit set to 10. If you have a limit of 10 you should only see 10 authors.

roosnic1 (approved):
🎉

penk85 (approved):

PR #240, opened 2021-04-08
f/wpc-320-create-an-author-in-the-article-metadata

HanaAyman:
I added a button that will appear when searching for an author that doesn't exist in the article metadata pannel and it will create the author and make it appear automatically so the publisher can then select it.


packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -380,6 +421,16 @@ export function ArticleMetadataPanel({
                     setAuthorsFilter('')
                   }}
                   block
+                  renderExtraFooter={() =>
+                    authorsFilter &&
+                    data?.authors.nodes.length === 0 && (
```
bishoymelek, line 426: can we have it like `!data?.authors.nodes.length && (` , what do you think?

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -380,6 +421,16 @@ export function ArticleMetadataPanel({
...
+                      <div style={{float: 'right', margin: '20px'}}>
```
bishoymelek, line 427: I guess using `float` has side effects and you will need to clearfix the parent. you can use flex for example what do you think?

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -380,6 +421,16 @@ export function ArticleMetadataPanel({
...
+                        <Button onClick={() => handleCreateAuthor()} appearance={'primary'}>
```
bishoymelek, line 428: `appearance='primary'`

bishoymelek (changes requested):
Amazing work @HanaAyman, just very simple comments

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -137,6 +143,23 @@ export function ArticleMetadataPanel({
     }
   }, [data?.authors, authors, socialMediaAuthors])
 
+  const [createAuthor] = useCreateAuthorMutation({
+    refetchQueries: [getOperationNameFromDocument(AuthorListDocument)]
```
roosnic1, line 147: Why did you add this? What is the purpose of that?

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -137,6 +143,23 @@ export function ArticleMetadataPanel({
...
+  })
+  const [authorName, setAuthorName] = useState('')
+  const [authorSlug, setAuthorSlug] = useState('')
```
roosnic1, line 150: i don't think you need a state for the slug. You can create the slug in `handleCreateAuthor()`

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -211,7 +234,11 @@ export function ArticleMetadataPanel({
                     value: author.id,
                     label: author.name
                   }))}
-                  onSearch={searchKeyword => setAuthorsFilter(searchKeyword)}
+                  onSearch={searchKeyword => {
+                    setAuthorsFilter(searchKeyword)
+                    setAuthorName(searchKeyword)
```
roosnic1, line 239: Do you even need a state for `authorName` you could also take the value from `authorsFilter`.

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -222,6 +249,16 @@ export function ArticleMetadataPanel({
                     setAuthorsFilter('')
                   }}
                   block
+                  renderExtraFooter={() =>
+                    authorsFilter &&
+                    data?.authors.nodes.length === 0 && (
+                      <div style={{float: 'right', margin: '20px'}}>
```
roosnic1, line 255: I would not use float right.

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -222,6 +249,16 @@ export function ArticleMetadataPanel({
...
+                        <Button onClick={() => handleCreateAuthor()} appearance={'primary'}>
+                          {t('authors.panels.createAuthor')}
```
roosnic1, line 257: Please make the label of the button more descriptive. "Create Auhtor Profile for %Name%"

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -380,6 +421,16 @@ export function ArticleMetadataPanel({
                     setAuthorsFilter('')
                   }}
                   block
+                  renderExtraFooter={() =>
```
roosnic1, line 424: you seem to repeat yourself here. It might be a good idea to create a new component callled AuthorCheckPicker and move the logic there.

roosnic1 (changes requested):
Looks very good just some inputs.

packages/editor/src/client/panel/articleMetadataPanel.tsx
HanaAyman: It's not needed anymore since I changed the button's name. 👍

packages/editor/src/client/panel/articleMetadataPanel.tsx
HanaAyman: I use it to make the new author I just created appear automatically in the list after clicking the button so the user can then select it without needing to refresh. Is there a better solution ?

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -222,6 +245,16 @@ export function ArticleMetadataPanel({
...
+                    authorsFilter &&
+                    !data?.authors.nodes.length && (
+                      <div style={{margin: '20px'}}>
+                        <Button onClick={() => handleCreateAuthor()} appearance="primary">
+                          {t('articles.panels.createAuthorProfile', {name: authorsFilter})}
```
penk85, line 253: It's a nice touch to add the user name to the comment, but as a personal preference I find it distracting that the button is dynamically changing size. Maybe the comment can be in a label and the button should have a static text "Create"?

penk85 (changes requested):

packages/editor/src/client/panel/articleMetadataPanel.tsx
roosnic1: ahh. This is awesome. Good job

packages/editor/src/client/panel/articleMetadataPanel.tsx
roosnic1: I think it's good if the name is in the button. Otherwise the users might think it will open the `AuthorEditPanel` and that's not what happens.

roosnic1 (changes requested):
Awesome work. Please remove all the commented code and then we can approve. 🎉

bishoymelek (approved):
:tada: Great work @HanaAyman

penk85 (approved):

roosnic1 (approved):
🥳

PR #244, opened 2021-04-12
Fix escaping regex

bishoymelek:


packages/api-db-mongodb/src/utility.ts
```diff
@@ -25,3 +25,7 @@ export function isNonNull<T>(value: T): value is NonNullable<T> {
 export enum MongoErrorCode {
   DuplicateKey = 11000
 }
+
+export function escapeRegExp(string: string) {
+  return string.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&')
```
roosnic1, line 30: Why the double `\\` in the `\\$&`? I think it would be very good if you add a comment to explain what is happening here.

packages/api-db-mongodb/src/db/memberPlan.ts
```diff
@@ -119,7 +120,7 @@ export class MongoDBMemberPlanAdapter implements DBMemberPlanAdapter {
 
     // TODO: Rename to search
     if (filter?.name !== undefined) {
-      textFilter.$and?.push({name: {$regex: filter.name, $options: 'i'}})
+      textFilter.$and?.push({name: {$regex: escapeRegExp(filter.name), $options: 'i'}})
```
roosnic1, line 123: Nice work. Could you please add the `escapeRegExp()` method to the `textFilter` methods in the other database adapters.

roosnic1 (changes requested):
Very good work.

penk85 (dismissed):

penk85 (approved):

roosnic1 (approved):
💯

PR #252, opened 2021-04-21
Comment list color

penk85:
WPC-340: Highlight color on comment table reflects the status of the comment state.

### DIFF
--- a/packages/editor/src/client/global.less
+++ b/packages/editor/src/client/global.less
@@ -1,9 +1,9 @@
-.highlighted-row,
-.highlighted-row .rs-table-cell-group .rs-table-cell {
-  background-color: #f2faff;
-}
-
 .rs-table {
+  .highlighted-row,
+  .highlighted-row &-cell-group &-cell {
+    background-color: #f2faff;
+  }
+
   &-hover &-row.approved:not(&-row-header):hover,
   &-hover &-row.approved:hover &-cell-group,
   &-hover &-row.approved:hover &-cell {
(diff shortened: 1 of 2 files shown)


HanaAyman (dismissed):

HanaAyman:
👍

packages/editor/src/client/global.less
```diff
@@ -1,4 +1,32 @@
 .highlighted-row,
 .highlighted-row .rs-table-cell-group .rs-table-cell {
-    background-color: #f2faff;
+  background-color: #f2faff;
```
roosnic1, line 3: i think we should do the highlighted-row from the duplicating page and article the same way.

roosnic1 (changes requested):
Looks awesome but i think we should do the same for the highlighted row.

packages/editor/src/client/global.less
```diff
@@ -1,9 +1,9 @@
-.highlighted-row,
-.highlighted-row .rs-table-cell-group .rs-table-cell {
-  background-color: #f2faff;
-}
-
 .rs-table {
+  .highlighted-row,
+  .highlighted-row &-cell-group &-cell {
```
bishoymelek, line 3: Out of curiosity, what does &-cell-group means here? same as `.rs-table-cell-group` ?

bishoymelek (commented):
Great work, small question

packages/editor/src/client/global.less
penk85: @bishoymelek yes

bishoymelek (approved):
🎉

roosnic1 (approved):
looks awesome

PR #254, opened 2021-04-22
add polis embed block

bishoymelek:


examples/website/src/shared/blocks/blockRenderer.tsx
```diff
@@ -30,9 +30,9 @@ export interface BlockRendererProps {
 export function BlockRenderer({blocks, children, ...props}: BlockRendererProps) {
   return (
     <>
-      {blocks.map(block =>
+      {blocks.map((block, index) =>
         block ? (
-          <React.Fragment key={block.key}>
+          <React.Fragment key={index}>
```
penk85, line 35: This gave a no key warning because there was no block.key. Possibly WPC-178.

PR #256, opened 2021-04-27
Color all rows on comment table

penk85:
WPC-340: Update comment list so rows are colored without hover.

### DIFF
--- a/packages/editor/src/client/global.less
+++ b/packages/editor/src/client/global.less
@@ -4,29 +4,36 @@
     background-color: #f2faff;
   }
 
-  &-hover &-row.approved:not(&-row-header):hover,
-  &-hover &-row.approved:hover &-cell-group,
-  &-hover &-row.approved:hover &-cell {
+  &-row.approved:not(&-row-header),
+  &-row.approved &-cell-group,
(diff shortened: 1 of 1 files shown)


packages/editor/src/client/global.less
```diff
@@ -4,29 +4,36 @@
...
-  &-hover &-row.rejected &-cell {
+  &-row.rejected:not(&-row-header),
+  &-row.rejected &-cell-group,
+  &-row.rejected &-cell {
     background: rgb(83, 85, 83);
     color: white;
     text-decoration: line-through;
   }
+
+  &-hover &-row.rejected:not(&-row-header):hover,
```
bishoymelek, line 33: What does this one do?

bishoymelek (commented):
Just one point

packages/editor/src/client/global.less
penk85: @bishoymelek the rejected row is colored gray with white text. On hover the background becomes light blue and the text is no longer visible so I set it black on hover.

bishoymelek (approved):
:tada:

HanaAyman (approved):

PR #262, opened 2021-05-06
calculate subscription amounts differently

bishoymelek:


roosnic1 (changes requested):
@bishoymelek: we'll discuss this in a call.

roosnic1:
@bishoymelek we discussed this with bajour and you can close the PR.

PR #264, opened 2021-05-06
b/wpc-369-make-default-publishing-time-5

HanaAyman:


penk85 (dismissed):

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -88,11 +88,11 @@ export function DateTimePicker({
                 value: new Date()
               },
               {
-                label: t('dateTimePicker.hour', {hour: '6'}),
+                label: t('dateTimePicker.hour', {hour: '5'}),
                 value: () => {
-                  const six = new Date()
-                  six.setHours(6, 0, 0)
-                  return six
+                  const five = new Date()
```
bishoymelek, line 93: Can you for example name it "date" instead?

bishoymelek (approved):

packages/editor/src/client/atoms/dateTimePicker.tsx
penk85: @bishoymelek it's a date object but it's the time, so probably better call it time or timePreset

bishoymelek (approved):

penk85 (approved):
great job :+1:

PR #271, opened 2021-05-27
F/wpc 260 shepherd tour

penk85:
WPC-260: Initial structure for guide tour


packages/editor/src/client/locales/en.json
```diff
@@ -724,6 +724,9 @@
       "nextSaturday": "Next Saturday",
       "hour": "{{hour}}:00"
     },
+    "guideTour": {
```
roosnic1, line 727: We should have all the text in the translations. I would start a new object on the top level and use that for all the guided tour translations.

packages/editor/src/client/index.tsx
```diff
@@ -113,7 +116,9 @@ const onDOMContentLoaded = async () => {
           <FacebookProvider sdkLanguage={'en_US'}>
             <InstagramProvider>
               <TwitterProvider>
-                <HotApp />
+                <ShepherdTour steps={steps} tourOptions={options}>
```
roosnic1, line 119: I suggest you move the whole `<ShepherdTour>` to it's own component. This way you can also remove the imports above from the `index.tsx` which we try to keep as clean as possible.

packages/editor/src/client/base.tsx
```diff
@@ -189,6 +190,13 @@ export function Base({children}: BaseProps) {
                   <DropdownItemLink route={LogoutRoute.create({})}>
                     {t('navbar.logout')}
                   </DropdownItemLink>
+                  <TourMethods>
```
roosnic1, line 193: We also need the possibility to start different tours. If you never started the editor you get the basic introduction tour but if we implemented a new feature we'll want to start that one. I think there should be some options inside shepherd to control this.

packages/editor/src/client/atoms/tourSteps.tsx
```diff
@@ -0,0 +1,96 @@
+import {Step} from 'react-shepherd'
+import {wait} from '../utility'
+
+const navButtons = {
+  exit: {
+    text: 'Exit',
```
roosnic1, line 6: i mention this down bellow but all the strings that are displayed should be in the translation file.

roosnic1 (changes requested):
looks very good but needs some work.

PR #272, opened 2021-05-27
validation for url peering to host

penk85:
WPC-22: Disable peering to host

### DIFF
--- a/packages/editor/src/client/panel/peerEditPanel.tsx
+++ b/packages/editor/src/client/panel/peerEditPanel.tsx
@@ -108,7 +108,7 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
             setErrorMessage(t('peerList.panels.invalidURL'))
           } else if (response?.data?.peerProfile.hostURL !== hostURL) {
             setValidPeer(false)
-            setErrorMessage('peerList.panels.peeredToHost')
+            setErrorMessage(t('peerList.panels.peeredToHost'))
           } else {
             setValidPeer(true)
             setProfile(response.data.peerProfile)
--- a/packages/editor/src/client/panel/peerEditPanel.tsx
+++ b/packages/editor/src/client/panel/peerEditPanel.tsx
@@ -111,6 +111,7 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
             setErrorMessage(t('peerList.panels.peeredToHost'))
           } else {
             setValidPeer(true)
+            setErrorMessage('')
             setProfile(response.data.peerProfile)
           }
         })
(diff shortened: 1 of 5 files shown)


bishoymelek (dismissed):

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -30,12 +30,13 @@ import {RichTextBlock} from '../blocks/richTextBlock/richTextBlock'
 
 export interface ImageEditPanelProps {
   id?: string
+  hostURL?: string
```
roosnic1, line 33: don't make this optional. You should always have a `hostURL`

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -101,7 +102,7 @@ export function PeerEditPanel({id, onClose, onSave}: ImageEditPanelProps) {
           setLoadingPeerProfile(false)
 
           // TODO: Better validation
-          if (response?.data?.peerProfile) {
+          if (response?.data?.peerProfile && response?.data?.peerProfile.hostURL !== hostURL) {
```
roosnic1, line 105: I would make an extra check for this so you can also set a specific error "You can not peer with yourself".

roosnic1 (changes requested):
Awesome work. Just some changes

packages/editor/src/client/panel/peerEditPanel.tsx
penk85: When there's no id?

packages/editor/src/client/panel/peerEditPanel.tsx
penk85: Can I perform the check inside the form control since it's just a boolean here?

HanaAyman (approved):

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -101,27 +103,31 @@ export function PeerEditPanel({id, onClose, onSave}: ImageEditPanelProps) {
...
-            setProfile(response.data.peerProfile)
+          if (!response?.data?.peerProfile) {
+            setValidPeer(false)
+            setErrorMessage(t('peerList.panels.invalidURL'))
+          } else if (response?.data?.peerProfile.hostURL !== hostURL) {
+            setValidPeer(false)
+            setErrorMessage(t('peerList.panels.peeredToHost'))
           } else {
-            setValidURL(false)
+            setValidPeer(true)
```
roosnic1, line 113: i think here you need to set the `ErrorMessage` to `''` otherwise you'll have a valid peer but still show an error.

roosnic1 (changes requested):

roosnic1 (approved):
🎉

bishoymelek (approved):
🎉🎉💯

PR #278, opened 2021-06-08
F/wpc 22 update

penk85:
WPC-22: Fix error caused by missing host url on peer panel.


packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -103,15 +104,15 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
           setLoadingPeerProfile(false)
 
           // TODO: Better validation
-          if (!response?.data?.peerProfile) {
+
+          if (!('peerProfile' in response.data)) {
```
roosnic1, line 108: Could you please change that back to `!response?.data?.peerProfile`? This way its more consistent with the rest of the code.

roosnic1 (changes requested):
just a tiny request

HanaAyman (approved):

roosnic1 (approved):

PR #282, opened 2021-06-12
fix missing media server setup

michael-scheurer:
If we run `yarn run` along the Readme instructions no media server is setup and thus no images can be uploaded in the editor.

### DIFF
--- a/examples/api/.env.example
+++ b/examples/api/.env.example
@@ -3,7 +3,7 @@ MONGO_URL=mongodb://localhost:27017/wepublish
 WEBSITE_URL=http://localhost:5000
 
 MEDIA_SERVER_URL=http://localhost:3004
-MEDIA_SERVER_TOKEN=secret
+MEDIA_SERVER_TOKEN=tPcvBRM2B3uSulyxjXm2ciqH5f1vQy2VDAsf
 
 OAUTH_GOOGLE_DISCOVERY_URL=https://accounts.google.com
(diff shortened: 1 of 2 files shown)


examples/api/.env.example
```diff
@@ -3,7 +3,7 @@ MONGO_URL=mongodb://localhost:27017/wepublish
 WEBSITE_URL=http://localhost:5000
 
 MEDIA_SERVER_URL=http://localhost:3004
-MEDIA_SERVER_TOKEN=secret
+MEDIA_SERVER_TOKEN=tPcvBRM2B3uSulyxjXm2ciqH5f1vQy2VDAsf
```
roosnic1, line 6: `secret` is the right media server token here. In the `examples/media/.env` you'll find the following environment variable `TOKEN=secret`. We could choose a more complex one but it's for local development purposes and therefore okay. If you want to deploy a productive we.publish instance you have to setup a more complex password.

package.json
```diff
@@ -19,7 +19,7 @@
     "publish": "lerna publish from-package --ignore-scripts",
     "watch": "run-p --print-label watch:*",
     "lint": "run-p --print-label lint:*",
-    "start:docker": "docker-compose up mongo mongo-express",
+    "start:docker": "docker-compose up mongo mongo-express media",
```
roosnic1, line 22: The idea is that you run `yarn run dev`. This will run `start:docker` and also the `watch` command which starts all the services like api, media, editor etc. on your local machine. If you want to start the whole environment in docker you can just do that with `docker-compose up`.

roosnic1 (commented):
@michael-scheurer  thanks for the PR. I hope my comments clarify the setup.

PR #283, opened 2021-06-14
F/wpc 165 article status indicator

penk85:
WPC-165: Usability issues: indicate an article or page status


packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -48,6 +48,12 @@ const InitialArticleBlocks: BlockValue[] = [
   {key: '1', type: BlockType.Image, value: {image: null, caption: ''}}
 ]
 
+export enum tagColor {
```
roosnic1, line 51: Enums usually start with an uppercase letter.

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -175,6 +181,32 @@ export function ArticleEditor({id}: ArticleEditorProps) {
     }
   }, [articleData])
 
+  const [borderColor, setBorderColor] = useState('')
```
roosnic1, line 184: You can set the type of a state. Since you only want values of `TagColor` enum you can write is `.... = useState<TagColor>()`. If you do that you're not able to use `''` as the default value. But you could add another entry to the enum `white` which can be used as a default color.

roosnic1: I also suggest to rename `borderColor` to state color. You're using the color for multiple things not just the border.

roosnic1 (changes requested):
looking very good. Just some small improvements.

HanaAyman (dismissed):

HanaAyman:
Nice work 👏

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -175,6 +176,32 @@ export function ArticleEditor({id}: ArticleEditorProps) {
     }
   }, [articleData])
 
+  const [stateColor, setStateColor] = useState<StateColor>(StateColor.none)
+  const [tagTitle, setTagTitle] = useState('')
```
roosnic1, line 180: here you can also set the type `string`

roosnic1 (changes requested):
just one last change :)

roosnic1 (approved):
🎉

bishoymelek (approved):

PR #288, opened 2021-06-16
Allow previews for peered articles

roosnic1:
WPC-17 This PR adds a new method to the `URLAdapter` that returns the URL of a peered article. GraphQL will return this url in the `PeerArticle` type and the editor will display it in the Teaser selection list.

### DIFF
--- a/examples/api/src/index.ts
+++ b/examples/api/src/index.ts
@@ -12,7 +12,8 @@ import {
   StripePaymentProvider,
   URLAdapter,
   WepublishServer,
-  JobType
+  JobType,
+  Peer
 } from '@wepublish/api'
 
 import {KarmaMediaAdapter} from '@wepublish/api-media-karma'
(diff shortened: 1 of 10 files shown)


penk85 (approved):

HanaAyman (approved):

examples/api/src/index.ts
```diff
@@ -43,6 +44,10 @@ class ExampleURLAdapter implements URLAdapter {
     return `${this.websiteURL}/a/${article.id}/${article.slug}`
   }
 
+  getPeeredArticleURL(peer: Peer, article: PublicArticle): string {
+    return `${this.websiteURL}/p/${peer.id}/${article.id}`
```
bishoymelek, line 48: but what if we implemented the pages as well, how would we differ between article and page urls

roosnic1: @bishoymelek pages dont have a `/a/` or `/p/` prefix in the URL path. They just use the slug after the domain.

PR #291, opened 2021-06-23
b/wpc-423-ui-fixes

HanaAyman:
This pull request is to fix some UI bugs. 

1. Language Icon on the bottom left of the editor should only be blue when hovering.
2.  Plus Icon should be behind the actions buttons in the articles and pages.
3. There should be some margin between the blocks and the buttons in the articles and pages.


packages/editor/src/client/atoms/editorTemplate.tsx
```diff
@@ -22,7 +22,7 @@ export function EditorTemplate({children, navigationChildren}: EditorTemplatePro
           display: 'flex',
           position: 'sticky',
           top: 0,
-          zIndex: 3,
+          zIndex: 10,
```
bishoymelek, line 25: Why have you changed this one? :eyes:

packages/editor/src/client/base.tsx
```diff
@@ -23,6 +23,8 @@ import {
 
 import {useTranslation} from 'react-i18next'
 
+import './global.less'
```
bishoymelek, line 26: It's already imported in `app.tsx` which uses base.tsx .. I guess you don't need to import it again in base component

bishoymelek (changes requested):
Helpful points to align styling! :100:  just two points to ask about.

packages/editor/src/client/atoms/editorTemplate.tsx
HanaAyman: This one is to make the plus buttons appear behind the metadata and the save buttons when scrolling.

bishoymelek: You mean disappear?

HanaAyman: yes as it currently passes on top of the buttons.
image

roosnic1 (approved):
👍🏾

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -408,7 +408,7 @@ export function ArticleEditor({id}: ArticleEditorProps) {
             <NavigationBar
               leftChildren={
                 <IconButtonLink
-                  style={{marginTop: '4px'}}
+                  style={{marginTop: '4px', marginBottom: '20px'}}
```
bishoymelek, line 411: Do we need this margin? won't it be affected by the one on line 422?

bishoymelek (commented):
Just very small point

packages/editor/src/client/base.tsx
```diff
@@ -214,6 +214,7 @@ export function Base({children}: BaseProps) {
                   trigger="click"
                   renderTitle={() => (
                     <Icon
+                      className="language-selector"
```
penk85, line 217: What's the advantage of having the icon blue only on hover? For consistency, can we have the same styling on the settings icon?

packages/editor/src/client/atoms/editorTemplate.tsx
penk85, line 25: What browser is it you use where it overlaps? It doesn't overlap on my screen.

HanaAyman: I use both google chrome and Microsoft edge and for me it overlapped on both.

penk85: Oh I see it now. Good catch :+1:

penk85 (approved):
:+1:

bishoymelek (approved):

roosnic1 (approved):
👍🏾

PR #294, opened 2021-07-01
Add properties tab to metadata

penk85:
WPC-422: Add tab in metadata on articles and pages which gives allows setting key value properties.


bishoymelek (dismissed):
:tada:

HanaAyman (approved):

packages/editor/src/client/panel/articleMetadataPanel.tsx
```diff
@@ -384,6 +410,48 @@ export function ArticleMetadataPanel({
             />
           </Panel>
         )
+      case MetaDataType.Properties:
+        return (
+          <Panel>
+            <ControlLabel>{t('articleEditor.panels.properties')}</ControlLabel>
+            <ListInput
```
roosnic1, line 417: I think we should also add a little info text here similar to the social media metadata. It should read something like "Properties can be set per article/page. Public properties are accessible through the public API"

roosnic1 (changes requested):
very good work just some minor changes requested.

roosnic1 (approved):
🎉

HanaAyman (approved):

PR #295, opened 2021-07-01
Handle URLs inside a peer article

bishoymelek:
This PR is about directing all teasers' URLs to external sites in new tabs if the user is in a peer article.


examples/website/src/shared/teasers/TeaserLink.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export interface TeaserLinkProps {
+  readonly url: string
+  readonly route?: Route
+  readonly children: ReactNode
+  readonly className?: string
+  readonly isPeerArticle?: boolean
+}
+
+export const TeaserLink = ({
```
roosnic1, line 12: You don't need this new component. The `<Link>` component already accepts the attributes `href` and with `target` you can control where it should open. If `<Link>` has `route===undefined` it takes the `href` attribute.

roosnic1 (changes requested):
Otherwise it looks good.

roosnic1 (approved):
🥳

penk85 (approved):

PR #296, opened 2021-07-02
Use JS SDK instead of iframe 

roosnic1:
WPC-215. This PR improves the polis implementation and uses the Javascript SDK instead of an iframe.


penk85 (approved):
:+1:

HanaAyman (approved):

packages/editor/src/client/blocks/embeds/polis.tsx
```diff
@@ -1,17 +1,19 @@
...
-      src={`https://pol.is/${conversationID}`}
-      width="100%"
-      height="600"
-      scrolling="yes"
-      frameBorder="0"
-    />
+    <div
+      className="polis"
+      data-page_id={conversationID}
+      data-site_id="polis_site_id_ssFrYaCCx4UpbUX63X"></div>
```
bishoymelek, line 17: what is this id? should be static? and how does this attribute work?

roosnic1: Good catch. I changed it

penk85 (approved):

HanaAyman (approved):

bishoymelek (approved):
:tada:

PR #306, opened 2021-07-09
Public mutation to cancel membership

roosnic1:
WPC-106. This PR adds a public mutation that sets the `deactivatedAt` date of a subscription. The PR also improves how we handle `updateUserSubscription`.


camilleory (approved):

packages/api/src/memberContext.ts
```diff
@@ -116,23 +116,26 @@ export class MemberContext implements MemberContext {
...
         if (periodA.endsAt > periodB.endsAt) return 1
         return 0
       })
       if (
-        periods.length > 0 &&
+        (periods.length > 0 || deactivatedAt === null) &&
         ((paidUntil === null && periods.length > 0) ||
           (paidUntil !== null && periods[periods.length - 1].endsAt > paidUntil))
       ) {
+        console.log('inside periods', periods)
```
penk85, line 130: You keeping this here?

penk85 (dismissed):

packages/api/src/memberContext.ts
```diff
@@ -116,23 +116,25 @@ export class MemberContext implements MemberContext {
...
-      const {periods = [], paidUntil} = userSubscription
+      const {periods = [], paidUntil, deactivatedAt} = userSubscription
       periods.sort((periodA, periodB) => {
         if (periodA.endsAt < periodB.endsAt) return -1
         if (periodA.endsAt > periodB.endsAt) return 1
         return 0
       })
       if (
-        periods.length > 0 &&
+        (periods.length > 0 || deactivatedAt === null) &&
```
bishoymelek, line 126: I think we don't need `> 0` part, what do you think?

roosnic1: @bishoymelek why do you think that?

bishoymelek (approved):

PR #310, opened 2021-07-13
Add image cta in peer profile

penk85:
WPC-434: Add an image Call To Action (image and URL) inside the peer profile.


bishoymelek (approved):
:tada:

HanaAyman (approved):

packages/api/__tests__/specs/__snapshots__/peer.ts.snap
```diff
@@ -135,6 +137,8 @@ exports[`Peers can be created/edited/deleted: can update peer profile 1`] = `
 Object {
   "data": Object {
     "updatePeerProfile": Object {
+      "callToActionImage": null,
+      "callToActionImageURL": null,
```
roosnic1, line 141: here it's `null` and in the test snapshot above it's `""`. I think it should be the same in both.

packages/editor/src/client/panel/peerProfileEditPanel.tsx
```diff
@@ -33,7 +33,10 @@ export function PeerInfoEditPanel({onClose, onSave}: ImageEditPanelProps) {
   const [name, setName] = useState('')
   const [themeColor, setThemeColor] = useState('')
   const [callToActionText, setCallToActionText] = useState<RichTextBlockValue>(createDefaultValue())
-  const [callToActionURL, setCallToActionURL] = useState('')
+  const [callToActionTextURL, setCallToActionTextURL] = useState('')
+  const [callToActionImageID, setCallToActionImageID] = useState<PeerProfileImage>()
```
roosnic1, line 37: I'm not sure if the `PeerProfileImage` is the right type. It's used for the peer profile and you have here just a normal image.

packages/editor/src/client/panel/peerProfileEditPanel.tsx
```diff
@@ -144,15 +204,22 @@ export function PeerInfoEditPanel({onClose, onSave}: ImageEditPanelProps) {
...
+            isLogoChange ? setLogoImage(value) : setCallToActionImageID(value)
           }}
         />
       </Drawer>
 
       <Drawer show={isEditModalOpen} size={'sm'} onHide={() => setEditModalOpen(false)}>
-        {logoImage && (
-          <ImagedEditPanel id={logoImage!.id} onClose={() => setEditModalOpen(false)} />
-        )}
+        {isLogoChange
```
roosnic1, line 213: this is a lot of repetitive code i think you can put must of the logic in side the `id={...}` expression.

roosnic1 (changes requested):

HanaAyman (approved):

packages/api-db-mongodb/src/db/peer.ts
```diff
@@ -30,8 +30,8 @@ export class MongoDBPeerAdapter implements DBPeerAdapter {
         themeColor: '#000000',
         callToActionURL: '',
         callToActionText: [],
-        callToActionImageID: '',
-        callToActionImageURL: ''
+        callToActionImageID: undefined,
```
roosnic1, line 33: they are undefined by default. I think you should set them to `''`

packages/editor/src/client/panel/peerProfileEditPanel.tsx
```diff
@@ -60,7 +62,7 @@ export function PeerInfoEditPanel({onClose, onSave}: ImageEditPanelProps) {
           : createDefaultValue()
       )
       setCallToActionTextURL(data.peerProfile.callToActionURL)
-      setCallToActionImageID(data.peerProfile.callToActionImage)
+      setCallToActionImage(data?.peerProfile?.callToActionImage ?? undefined)
```
roosnic1, line 65: you don't need the `?? undefined` if you use them before the dots.

roosnic1 (changes requested):

HanaAyman (approved):

roosnic1 (dismissed):

bishoymelek (approved):
:tada:

HanaAyman (approved):

roosnic1 (approved):
👍🏾

PR #312, opened 2021-07-19
Updated translations from Phrase

roosnic1:
Exported the most recent locale files from Phrase


bishoymelek (commented):
Great we are having them back. :tada: 
I found that in French we have the navigation bar exactly the same as in English.
And in German, "Authors" has * character, is it normal?
Screenshot from 2021-07-19 12-06-02

PR #313, opened 2021-07-19
Update react to v17

roosnic1:


packages/editor/src/client/atoms/typographicTextArea.tsx
```diff
@@ -1,7 +1,6 @@
 import React, {
```
bishoymelek, line 1: There is an issue in this file now: 
Screenshot from 2021-07-20 12-06-41
You can reproduce it in article edit page => add teasers, choose an article and press on save.

bishoymelek (approved):

packages/editor/src/client/atoms/typographicTextArea.tsx
roosnic1: @bishoymelek i can't reproduce the error. Maybe i'm doing it wrong. What did you mean with "in article edit page"???

packages/editor/src/client/atoms/helpers.ts
```diff
@@ -1,40 +1,4 @@
-import {CSSStyle} from '@karma.run/react'
-
-import {
```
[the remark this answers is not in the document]
roosnic1: @bishoymelek was not needed anymore. The original author implemented something here that he never reused anywhere else. And it doesn't work with React 17 anymore.

packages/editor/src/client/atoms/typographicTextArea.tsx
roosnic1: @bishoymelek fixed it

bishoymelek (approved):
:tada:

PR #326, opened 2021-07-29
wpc-461-add-frame-around-author-bio-info

HanaAyman:
This pull request is to add a thin frame arounf the biographical information in the create author panel.
image


packages/editor/src/client/panel/authorEditPanel.tsx
```diff
@@ -220,7 +220,9 @@ export function AuthorEditPanel({id, onClose, onSave}: AuthorEditPanelProps) {
             </ListInput>
           </Panel>
           <Panel header={t('authors.panels.bioInformation')}>
```
roosnic1, line 222: `<Panel>` has an option for an border. https://rsuitejs.com/components/panel/#%3CPanel%3E

roosnic1 (changes requested):

packages/editor/src/client/panel/authorEditPanel.tsx
HanaAyman: It won't work as it will add the frame around the whole panel including the title and not just the rich text block.

packages/editor/src/client/panel/authorEditPanel.tsx
```diff
@@ -220,7 +220,9 @@ export function AuthorEditPanel({id, onClose, onSave}: AuthorEditPanelProps) {
...
-            <RichTextBlock value={bio} onChange={value => setBio(value)} />
+            <div className="richTextFrame">
```
penk85, line 223: Since this looks like a reusable and practical feature, can you put it inside the RichTextBlock component as an optional prop?

roosnic1: @penk85 we're rebuilding the whole richTextBlock with the dreifuss-wysiwyg-editor and will make the changes once we implement it.

roosnic1 (approved):
👍🏾

penk85 (approved):

PR #329, opened 2021-07-29
wpc-460 add frame around description in member plan panel 

HanaAyman:
This pull request is to add a thin frame around the description in the create member plan panel.
image


packages/editor/src/client/global.less
```diff
@@ -72,3 +72,16 @@
     visibility: hidden;
   }
 }
+
+.frame {
+  padding-bottom: 20px;
```
bishoymelek, line 77: Regarding padding, you can write it as `padding: 0px 20px 20px;`

packages/editor/src/client/global.less
```diff
@@ -72,3 +72,16 @@
...
+  padding-left: 20px;
+  padding-right: 20px;
+  border-style: solid;
```
bishoymelek, line 80: `border: 1px solid #e5e5ea;`

penk85 (approved):

roosnic1 (approved):

Issue #76, opened 2020-11-13
Get rid of @ts-ignore in code


roosnic1:
The PR #64 introduced a new UI framework but also some `// @ts-ignore`. Analyze why typescript complains and fix it.

PR #337, opened 2021-08-04
Get rid of @ts-ignore

gianantoniopini:
This pull request is to get rid of some of the `// @ts-ignore` in the code.
It is somehow related to issue #76 , however there are still some `// @ts-ignore` left.


gianantoniopini:
Hi @bishoymelek @HanaAyman @roosnic1, could you please have a look at this pull request ?

packages/api/src/server.ts
```diff
@@ -39,12 +39,12 @@ export class WepublishServer {
 
     methodsToProxy.forEach(mtp => {
       if (mtp.key in dbAdapter) {
+        const dbAdapterKeyTyped = mtp.key as keyof typeof dbAdapter
```
bishoymelek, line 42: I think this file changes needs more enhancement, as this line isn't satisfying for line 47 for example.

bishoymelek (changes requested):
Thank you @gianantoniopini for the contribution. I just left small comments and will debug it later this week, if still pending. thanks.

gianantoniopini:
Thank you @bishoymelek for the review. I have started to address the requested changes.

packages/api-db-mongodb/src/db/invoice.ts
```diff
@@ -233,4 +232,6 @@ function invoiceDateForSort(invoice: DBInvoice, sort: InvoiceSort): Date {
     case InvoiceSort.PaidAt:
       if (invoice.paidAt) return invoice.paidAt
   }
+
+  return undefined
```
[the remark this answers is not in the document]
gianantoniopini: @bishoymelek Yes, it is a bit more clear with default case. I've just changed it.

packages/api/src/server.ts
gianantoniopini: Yes the `// @ts-ignore` for line 47 is still there as I could not find a simple way to get rid of that. I will investigate further.

gianantoniopini: Not sure how to get rid of the `// @ts-ignore` for line 47, as this one is a bit beyond my current knowledge of TypeScript. I believe it is not straightforward as `methodName` is only known at runtime and it is used as a deep key of the `dbAdapter` object. @bishoymelek Perhaps you have some ideas?

packages/api-db-mongodb/src/db/invoice.ts
```diff
@@ -232,5 +231,9 @@ function invoiceDateForSort(invoice: DBInvoice, sort: InvoiceSort): Date {
 
     case InvoiceSort.PaidAt:
       if (invoice.paidAt) return invoice.paidAt
```
roosnic1, line 233: I'd recommend something like `return invoice.paidAt ? invoice.paidAt : invoice.createdAt` here. With this you don't need the `... | undefined` on line 224.

packages/api/src/server.ts
roosnic1: @gianantoniopini: this whole implementation should get refactored anyway so I am already happy if we can get rid of the `// @ts-ignore` you did 😃 .

roosnic1 (changes requested):
@gianantoniopini thanks so much 🎉 . Looks very good just one little change and then we can merge it.

packages/api-db-mongodb/src/db/invoice.ts
gianantoniopini: Thanks. Just changed it as per your recommendation. I've also changed the `switch` `default` case to `return invoice.createdAt`. Happy to apply more changes if needed.

gianantoniopini:
@roosnic1 thanks a lot for reviewing this 😃 . I've just applied the change as per your recommendation. Let me know in case more tweaks are required.

roosnic1 (approved):
🎉 awesome thank you very much

bishoymelek (approved):
Thanks a lot @gianantoniopini 💯 🎉

penk85 (approved):
:+1:

camilleory (approved):

lakeside763 (approved):

PR #339, opened 2021-08-07
wpc-444-reply to comments only when logged in

HanaAyman:
This pull request is to make the reply button in the comments section appear only if the user is logged in.


penk85 (approved):

bishoymelek (approved):
:tada:

PR #344, opened 2021-08-25
make light teaser option use the original image

HanaAyman:
This pull request is to make the Light teaser option also use the original teaser image and doesn't give an error if the user doesn't select another image


penk85 (approved):

penk85 (dismissed):

bishoymelek (approved):
🎉🎉🎉

roosnic1 (approved):
🎉

bishoymelek (approved):
🎉

penk85 (approved):

PR #358, opened 2021-09-15
F/wpc 482 format date with fns

camilleory:
I replaced moment.js with date fns to format the dates so we only use one library in our application.


packages/editor/package.json
```diff
@@ -56,11 +56,13 @@
     "apollo-upload-client": "^16.0.0",
     "array-move": "^2.2.1",
     "core-js": "^3.4.8",
+    "date-fns": "2.0.0-alpha.23",
     "emoji-mart": "^3.0.0",
     "express": "^4.17.1",
     "graphql": "^14.5.8",
     "i18next": "^19.7.0",
     "i18next-browser-languagedetector": "^6.0.1",
+    "moment": "^2.29.1",
```
penk85, line 65: Do we still need this?

roosnic1: @camilleory this can be removed.

packages/editor/src/client/i18n.ts
```diff
@@ -26,8 +42,4 @@ i18n.use(LanguageDetector).use(initReactI18next).init({
   }
 })
 
-i18n.on('languageChanged', function (lng) {
```
bishoymelek, line 29: why did you remove this?

packages/editor/src/client/locales/de.json
```diff
@@ -11,13 +11,13 @@
         "metadata": "Metadaten",
         "noSlug": "Artikel nicht publiziert. Bitte Slug erfassen.",
         "notFound": "Artikel nicht gefunden",
-        "pending": "Publikation des Artikels geplant am {{date, lll}}",
+        "pending": "Publikation des Artikels geplant am {{date, ccc PP p}}",
```
bishoymelek, line 14: can you explain what's c P p mean?

bishoymelek (commented):
Great work @camilleory , just simple feedback/questions.

packages/editor/src/client/locales/de.json
camilleory: c formats the day (ccc -> Mon), P the date (12/12/2021) and p the time (p for 00:00 and pp for 00:00:00). You can read the documentation [here](https://date-fns.org/v2.23.0/docs/format)

packages/editor/src/client/i18n.ts
```diff
@@ -5,12 +5,27 @@ import LanguageDetector from 'i18next-browser-languagedetector'
...
-import moment from 'moment'
+
+import {format as formatDate, isDate} from 'date-fns'
+import {de as deLocale, enGB as enLocale, fr as frLocale} from 'date-fns/locale'
+
+export interface DateFormatLocale {
+  lng: string
+  locale: Locale
+}
+const DateFormatLocales: DateFormatLocale[] = [
```
roosnic1, line 16: I suggest you use a javascript map here instead of an array. It will be easier to get the locale on line 26.

packages/editor/src/client/i18n.ts
roosnic1: @bishoymelek it's not needed anymore. As you can see on line 27. The locale will be passed to date-fns everytime it being used for a translation rendering.

@camilleory you can remove the code. With source control there is no need for commenting out code :)

roosnic1 (changes requested):
looks good. some minor improvements.
@camilleory why did you add this file `packages/editor/src/client/date-fns.ts`

packages/editor/src/client/i18n.ts
```diff
@@ -26,8 +41,8 @@ i18n.use(LanguageDetector).use(initReactI18next).init({
   }
 })
 
-i18n.on('languageChanged', function (lng) {
-  moment.locale(lng)
-})
+// i18n.on('languageChanged', function (lng) {
+//   moment.locale(lng)
+// })
```
lakeside763, line 47: I think line 44 to 46 can be removed, since it's no longer needed

lakeside763 (commented):
it's look good, except for the comments

packages/editor/src/client/i18n.ts
```diff
@@ -13,18 +13,18 @@ export interface DateFormatLocale {
   lng: string
   locale: Locale
 }
-const DateFormatLocales: DateFormatLocale[] = [
-  {lng: 'de', locale: deLocale},
-  {lng: 'en', locale: enLocale},
-  {lng: 'fr', locale: frLocale}
-]
+const DateFormatLocales = new Map<string, Locale>([
```
roosnic1, line 16: If you use the `new Map` you don't need the `DateFormatLocale` typescript interface on line 12 anymore.

roosnic1: And in general normal javascript variables start with a lowercase letter.

roosnic1 (changes requested):
🎉 all good except for two tiny changes.

bishoymelek (approved):
🎉

roosnic1 (approved):
Good work 🥳

PR #364, opened 2021-09-21
F/wpc 499 add fr de translation where missing

camilleory:


camilleory:
This pulled request got mixed up with older changes, I replaced it with  #365 which only contains the right changes to review

PR #365, opened 2021-09-22
f/499-translation-rsuite-js

camilleory:
Select pickers for tags and authors translated, 
Blocks dropdown menus in article editor and page editor translated,
Phrase "no data found" in all editor sections overviews translated. 
image


packages/editor/src/client/app.tsx
```diff
@@ -24,6 +24,12 @@ import {PaymentMethodList} from './routes/paymentMethodList'
 import {NavigationList} from './routes/navigationList'
 
 import './global.less'
+import {IntlProvider} from 'rsuite'
+import enGB from 'rsuite/lib/IntlProvider/locales/en_GB'
+import fr from './locales/rsuiteFr'
+import en from './locales/rsuiteEn'
```
roosnic1, line 30: I think the `rsuiteEn` is the same as the one you import on line 28 from the `.../lib/IntlProvider/locales/...` you can do it both way but i suggest to use the one provided by rsuite for the english language.

packages/editor/src/client/app.tsx
```diff
@@ -121,5 +125,26 @@ export function App() {
 
   return null
 }
+export function App() {
+  const {current} = useRoute()
 
+  const {i18n} = useTranslation()
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
roosnic1, line 132: you can get rid of the eslint comment by using `Record<string, unknown>` on line 133 instead of `object`.

packages/editor/src/client/app.tsx
```diff
@@ -121,5 +125,26 @@ export function App() {
...
+  const [lng, setLang] = useState<object>(enGB)
+
+  i18n.on('languageChanged', lng => {
+    const currentLanguage: any = new Map()
+
+    if (lng === 'fr') {
```
roosnic1, line 138: if think instead of this if - else if you could also use a `new Map` here.

roosnic1 (changes requested):
Looks good. Just some suggestions

packages/editor/src/client/app.tsx
camilleory: This one is not giving any errors, so I just removed the comment

packages/editor/src/client/app.tsx
```diff
@@ -129,20 +128,15 @@ export function App() {
...
-
-    if (lng === 'fr') {
-      currentLanguage.set(lng, fr)
-    } else if (lng === 'de') {
-      currentLanguage.set(lng, de)
-    } else if (lng === 'en') {
-      currentLanguage.set(lng, en)
-    }
-    setLang(currentLanguage.get(lng))
+    const currentLanguageMap = new Map<string, any>([
```
roosnic1, line 134: You should initialize `const currentLanguageMap` outside the `languageChanged` function. Otherwise you always reset the map on language change.

roosnic1 (changes requested):
one last change

packages/editor/src/client/app.tsx
camilleory: 👍

lakeside763 (commented):
Great job, I have no comment

penk85 (approved):

roosnic1 (approved):
👍🏾

PR #367, opened 2021-09-24
User and Session Database Adapter improvements 

roosnic1:
WPC-449. This PR adds fields and methods to the user and session models. With those it will be possible for users of we.publish to implement their own authentication solution in the website.

### DIFF
--- a/packages/api-db-mongodb/src/db/schema.ts
+++ b/packages/api-db-mongodb/src/db/schema.ts
@@ -16,7 +16,8 @@ import {
   UserSubscription,
   MailLogState,
   PaymentState,
-  UserAddress
+  UserAddress,
+  UserOAuth2Account
 } from '@wepublish/api'
 
 export enum CollectionName {
@@ -100,10 +101,13 @@ export interface DBUser {
   modifiedAt: Date
 
   email: string
+  emailVerified: Date | null
   name: string
   preferredName?: string
   address?: UserAddress
   password: string
+
+  oauth2Accounts: UserOAuth2Account[]
 
   active: boolean
   lastLogin: Date | null
--- a/packages/api-db-mongodb/src/db/session.ts
+++ b/packages/api-db-mongodb/src/db/session.ts
@@ -99,6 +99,60 @@ export class MongoDBSessionAdapter implements DBSessionAdapter {
     return null
   }
 
+  async extendSessionByToken(token: string): Promise<OptionalSession> {
+    const [tokenMatch, session] = await Promise.all([
+      this.tokens.findOne({token}),
+      this.sessions.findOne({token})
+    ])
+
+    if (tokenMatch) {
+      const {value} = await this.tokens.findOneAndUpdate(
+        {_id: tokenMatch._id},
+        {
+          $set: {
+            expiresAt: new Date(Date.now() + this.sessionTTL)
+          }
+        },
+        {returnOriginal: false}
+      )
+      return value
+        ? {
+            type: SessionType.Token,
+            id: value._id,
+            name: value.name,
+            token: value.token,
+            roles: await this.userRole.getNonOptionalUserRolesByID(value.roleIDs)
+          }
+        : null
+    } else if (session) {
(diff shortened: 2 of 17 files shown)


packages/api/src/db/user.ts
```diff
@@ -45,13 +50,6 @@ export interface UserSubscriptionFilter {
   readonly autoRenew?: boolean
 }
 
-export interface UserSubscriptionFilter {
```
roosnic1, line 48: removed the interface because it was duplicated.

packages/api/src/db/user.ts
```diff
@@ -76,14 +74,6 @@ export interface UserSubscriptionPeriodInput {
   readonly invoiceID: string
 }
 
-export interface UserSubscriptionPeriodInput {
```
roosnic1, line 79: removed the interface because it was duplicated.

packages/api-db-mongodb/src/db/schema.ts
```diff
@@ -100,11 +101,14 @@ export interface DBUser {
   modifiedAt: Date
 
   email: string
+  emailVerified: Date | null
```
penk85, line 104: Would it be better to have an indication in the naming that this is a date? `emailVerifiedAt` or something?

packages/api-db-mongodb/src/migration.ts
```diff
@@ -593,6 +593,26 @@ export const Migrations: Migration[] = [
         }
       )
     }
+  },
+  {
+    //  Rename street field in address to address
```
penk85, line 598: I think you meant to change this

penk85 (changes requested):
Left some comments

packages/api-db-mongodb/src/db/session.ts
```diff
@@ -99,6 +99,32 @@ export class MongoDBSessionAdapter implements DBSessionAdapter {
     return null
   }
 
+  async extendUserSessionByToken(token: string): Promise<OptionalUserSession> {
```
bishoymelek, line 102: I think it's so helpful to keep users logged-in depending on specific condition(s) 👍

packages/api-db-mongodb/src/db/user.ts
```diff
@@ -475,6 +520,57 @@ export class MongoDBUserAdapter implements DBUserAdapter {
...
+
+    const {value} = await this.users.findOneAndUpdate(
+      {_id: userID},
+      {
+        $set: {
+          modifiedAt: new Date(),
+          oauth2Accounts: accounts
+        }
+      },
+      {returnOriginal: false}
```
bishoymelek, line 538: I think after a while we have to update this one as it's deprecated [see here](https://stackoverflow.com/a/67909382/7254406)

packages/api-db-mongodb/src/migration.ts
```diff
@@ -593,6 +593,26 @@ export const Migrations: Migration[] = [
...
+          emailVerified: {$exists: false}
+        },
+        {$set: {emailVerified: null}}
+      )
+
+      await users.updateMany(
+        {
+          oauth2Accounts: {$exists: false}
+        },
+        {$set: {oauth2Accounts: []}}
```
bishoymelek, line 613: can't you add this attr to the upper mutation? why two?

bishoymelek (changes requested):
@roosnic1 really helpful feature 🎉 , Just one comment and the `emailVerified` field @penk85 commented

packages/api-db-mongodb/src/db/user.ts
roosnic1: True but that will happen in the general refactoring of core.

packages/api-db-mongodb/src/migration.ts
roosnic1: because it's should only set the field it it doesn't exist.

roosnic1:
Closing PR because the branch name is wrong.

PR #368, opened 2021-09-24
f/wpc-526-theme-color-visible-on-webiste

camilleory:
background-color field rendered on website

I suggest waiting for wpc-524 - (As a publisher, i want to choose theme color with a color panel) to be ready before reviewing and pushing so that the text is always readable


roosnic1 (approved):
👍🏾

lakeside763 (approved):

PR #369, opened 2021-09-24
F/wpc 524 color picker peered articles color

camilleory:
I replaced the theme color in edit peer info by a color picker. 

I also replaced the color field in peer by a div showing the color (instead of the hex number).

image
image

### DIFF
--- a/packages/editor/src/client/panel/peerEditPanel.tsx
+++ b/packages/editor/src/client/panel/peerEditPanel.tsx
@@ -220,13 +220,7 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
                 {profile?.name}
               </DescriptionListItem>
               <DescriptionListItem label={t('peerList.panels.themeColor')}>
-                <div
-                  style={{
-                    backgroundColor: profile?.themeColor,
-                    width: '40px',
-                    height: '20px',
-                    padding: '5px'
-                  }}></div>
+                {profile?.themeColor}
               </DescriptionListItem>
               <DescriptionListItem label={t('peerList.panels.callToActionText')}>
                 {!!profile?.callToActionText && (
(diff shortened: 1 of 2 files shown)


packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -220,7 +220,13 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
                 {profile?.name}
               </DescriptionListItem>
               <DescriptionListItem label={t('peerList.panels.themeColor')}>
-                {profile?.themeColor}
+                <div
```
roosnic1, line 223: I recommend to still display the hex code of the theme color.

roosnic1 (changes requested):
Looks good but keep the hex color code.

penk85 (dismissed):
:+1:  very nice

packages/editor/src/client/panel/peerEditPanel.tsx
camilleory: Ok I'll change it back, can you tell me why it's better with the hex code ?

penk85 (dismissed):

lakeside763 (dismissed):
nice one camille

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -220,7 +220,17 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
...
+                  <div
+                    style={{
+                      backgroundColor: profile?.themeColor,
+                      width: '30px',
+                      height: '20px',
+                      padding: '5px',
+                      marginLeft: '5px'
+                    }}></div>
+                </div>
               </DescriptionListItem>
```
camilleory, line 234: Is this way of adding css a good way in this case ?

roosnic1: for a few css properties yes.

roosnic1 (approved):
👍🏾

bishoymelek (approved):
🎉🎉

PR #370, opened 2021-09-27
user and session improvements

roosnic1:
WPC-449. I had to close #367 because i named the branch wrong. In this PR i just renamed the field `emailVerified` to `emailVerifiedAt`.


packages/api-db-mongodb/src/migration.ts
```diff
@@ -593,6 +593,26 @@ export const Migrations: Migration[] = [
         }
       )
     }
+  },
+  {
+    //  Rename street field in address to address
```
penk85, line 598: This migration comment should probably be changed

packages/api-db-mongodb/src/db/user.ts
```diff
@@ -475,6 +520,57 @@ export class MongoDBUserAdapter implements DBUserAdapter {
...
+
+    const {value} = await this.users.findOneAndUpdate(
+      {_id: userID},
+      {
+        $set: {
+          modifiedAt: new Date(),
+          oauth2Accounts: accounts
+        }
+      },
+      {returnOriginal: false}
```
camilleory, line 538: I don't understand what this line does - `{returnOriginal: false}`. Can you explain ?

packages/api/__tests__/api/private/index.ts
```diff
@@ -1676,6 +1677,7 @@ export type UserFilter = {
 export type UserInput = {
   name: Scalars['String']
   email: Scalars['String']
+  emailVerifiedAt?: Maybe<Scalars['DateTime']>
```
camilleory, line 1680: Why do we need the maybe type here ?

camilleory (approved):

lakeside763 (dismissed):

packages/api-db-mongodb/src/db/user.ts
roosnic1: It means that the operation should return the updated entry rather than the original one. But as @bishoymelek pointed out this option is deprecated and should be replaced with `returnNewDocument`.

packages/api/__tests__/api/private/index.ts
roosnic1: If the email isn't verified yet the property `emailVerifiedAt` will be null. That's why it's maybe.

packages/api-db-mongodb/src/migration.ts
roosnic1: done. Thank you :)

bishoymelek (approved):
🎉🎉👌

packages/api/__tests__/api/private/index.ts
camilleory: Thanks for your explanations

camilleory (approved):

PR #372, opened 2021-09-28
peering between different versions

roosnic1:
WPC-112. This PR changes the way how to add a new peer. Instead of accessing the API of the new peer the editor sends a query to it's own API which will then delegate the request to the other API. This way we can better handle changes in the GraphQL Schema.


packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -1,25 +1,15 @@
...
 
 import {
   PeerListDocument,
   useCreatePeerMutation,
   usePeerQuery,
   useUpdatePeerMutation,
-  PeerProfileDocument,
-  FullPeerProfileFragment
+  FullPeerProfileFragment,
+  useRemotePeerPrfoileQuery
```
penk85, line 12: Little typo here: profile

roosnic1: Good catch. Thank you :)

packages/editor/src/client/api/index.ts
```diff
@@ -1340,6 +1342,12 @@ export type Query = {
 };
 
 
+export type QueryRemotePeerProfileArgs = {
+  hostURL: Scalars['String'];
```
camilleory, line 1345: what's the difference between using `Scalars['String'];` or simply `String`?

packages/editor/src/client/api/index.ts
```diff
@@ -6118,6 +6147,42 @@ export function usePeerProfileLazyQuery(baseOptions?: Apollo.LazyQueryHookOption
...
+ *      hostURL: // value for 'hostURL'
+ *      token: // value for 'token'
+ *   },
+ * });
+ */
+export function useRemotePeerPrfoileQuery(baseOptions: Apollo.QueryHookOptions<RemotePeerPrfoileQuery, RemotePeerPrfoileQueryVariables>) {
+        const options = {...defaultOptions, ...baseOptions}
+        return Apollo.useQuery<RemotePeerPrfoileQuery, RemotePeerPrfoileQueryVariables>(RemotePeerPrfoileDocument, options);
+      }
+export function useRemotePeerPrfoileLazyQuery(baseOptions?: Apollo.LazyQueryHookOptions<RemotePeerPrfoileQuery, RemotePeerPrfoileQueryVariables>) {
```
camilleory: What does this lazy query does ?

camilleory (approved):
Found a spelling mistake and asked a couple of questions

packages/editor/src/client/api/index.ts
```diff
@@ -3196,6 +3208,20 @@ export type PeerProfileQuery = (
   ) }
 );
 
+export type RemotePeerPrfoileQueryVariables = Exact<{
```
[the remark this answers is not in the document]
roosnic1: already fixed :)

packages/editor/src/client/api/index.ts
roosnic1: We use https://www.graphql-code-generator.com to generate this file out of the `*.graphql` files. So you don't have to review it in the PR. I think scalars belongs to something graphql specific.

packages/editor/src/client/api/index.ts
roosnic1: graphql-codegen creates two kind of queries. One to execute directly and another one that can be executed on a click. https://www.apollographql.com/docs/react/data/queries/#manual-execution-with-uselazyquery

bishoymelek (approved):
🎉 🎉

penk85 (approved):

lakeside763 (approved):

PR #376, opened 2021-09-30
F/wpc 522 add font color to peered articles

camilleory:
I added a new field for the font color of peered articles (so that the color can be light when the theme is dark). 
I added the field in the peer info as a colored div as well, so that both theme and font colors get displayed.


packages/api-db-mongodb/src/db/peer.ts
```diff
@@ -28,6 +28,7 @@ export class MongoDBPeerAdapter implements DBPeerAdapter {
       return {
         name: '',
         themeColor: '#000000',
+        themeFontColor: '#d67c15',
```
penk85, line 31: Do you want this color default?

packages/api/src/db/peer.ts
```diff
@@ -3,7 +3,8 @@ import {RichTextNode} from '../graphql/richText'
 export interface PeerProfile {
   name: string
   logoID?: string
-  themeColor: string
+  themeColor?: string
```
penk85, line 6: Do you want it optional here?

packages/editor/src/client/locales/rsuiteEn.ts
```diff
@@ -0,0 +1,65 @@
+// const Calendar = {
```
penk85, line 1: Why did you comment this out here?

camilleory: This file will be deleted, I don't know why it hasn't been when I updated the branch.

packages/api-db-mongodb/src/db/peer.ts
camilleory: Maybe not the nicest color indeed, I'll change it to white

packages/api/src/db/peer.ts
camilleory: I suppose it's more consistant if it's not optional, I changed it

packages/editor/src/client/panel/peerEditPanel.tsx
```diff
@@ -232,6 +232,22 @@ export function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps
...
+                    style={{
+                      backgroundColor: profile?.themeFontColor,
+                      width: '30px',
+                      height: '20px',
+                      padding: '5px',
+                      marginLeft: '5px'
+                    }}></div>
+                </div>
+              </DescriptionListItem>
+              {/* <DescriptionListItem label={t('peerList.panels.themeFontColor')}>
```
bishoymelek, line 248: Do you still want this?

bishoymelek (changes requested):
🎉 good one, just a simple point.

packages/editor/src/client/panel/peerEditPanel.tsx
camilleory: No, small forgetfullness on my side

examples/website/src/shared/route/articleAdapter.ts
```diff
@@ -10,6 +10,7 @@ export function peerAdapter(peer: any): Peer {
     logoURL: peer.profile.logo?.squareURL,
     websiteURL: peer.profile.websiteURL,
     themeColor: peer.profile.themeColor,
+    themefontColor: peer.profile.fontColor,
```
roosnic1, line 13: is this a typo? Or hwy only `fontColor`?

roosnic1 (commented):
Looks good justa question

examples/website/src/shared/route/articleAdapter.ts
camilleory: I thought that would be more precise and prevent confusion if we ever need to use font colors anywhere else in future.

penk85 (approved):
Looks great

roosnic1 (approved):
🥳

bishoymelek (approved):
🎉🎉

PR #377, opened 2021-09-30
make state colors of articles/pages consistent

penk85:
WPC-533: use a single color enum for states across the editor.


packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -409,7 +409,7 @@ export function ArticleEditor({id}: ArticleEditorProps) {
     <>
       <fieldset style={{borderColor: stateColor}}>
         <legend style={{width: 'auto', margin: '0px auto'}}>
-          <Tag color={stateColor}>{tagTitle}</Tag>
+          <Tag style={{backgroundColor: stateColor}}>{tagTitle}</Tag>
```
penk85, line 412: @bishoymelek I think this resolves the Rsuite color warning

camilleory (approved):

packages/editor/src/client/global.less
```diff
@@ -7,19 +7,19 @@
   &-row.approved:not(&-row-header),
   &-row.approved &-cell-group,
   &-row.approved &-cell {
-    background: rgb(225, 248, 222);
+    background: #e1f8de;
```
roosnic1, line 10: are those entries still needed? And if yet where?

roosnic1 (commented):
looks good but i'm not sure about the css in the `global.less` file

lakeside763 (approved):

packages/editor/src/client/global.less
penk85: @roosnic1 this is for the colors on the comment list. I thought it's better if it's the same colors throughout the editor.

roosnic1: 👍🏾

roosnic1 (approved):
🎉

bishoymelek (approved):
🎉  good one

PR #396, opened 2021-10-28
F/wpc 567 pending notification for future publications

camilleory:
- Notification for future publication date in pages and articles editor now show "Article / page pending" instead of "published".
- Test [**here**](https://editor.fwpc567pe.wepublish.dev/)

Oct-28-2021 15-45-43


packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -378,10 +378,16 @@ export function ArticleEditor({id}: ArticleEditorProps) {
       }
 
       setChanged(false)
-      Notification.success({
-        title: t('articleEditor.overview.articlePublished'),
-        duration: 2000
-      })
+
+      publishData === new Date() || publishDate < new Date()
```
roosnic1, line 382: I think you can use smaller or equal for the comparison instead of comparing or smaller. `publishDate <= new Date()`.

Also i think you have a typo. The first argument is publishDat**a** instead of publishDat**e**.

packages/editor/src/client/routes/pageEditor.tsx
```diff
@@ -228,7 +228,9 @@ export function PageEditor({id}: PageEditorProps) {
     }
 
     setChanged(false)
-    Alert.success(t('pageEditor.overview.pagePublished'), 2000)
+    publishData === new Date() || publishDate < new Date()
```
roosnic1, line 231: I think you can use smaller or equal for the comparison instead of comparing or smaller. `publishDate <= new Date()`.

Also i think you have a typo. The first argument is publishDat**a** instead of publishDat**e**.

packages/editor/src/client/routes/pageEditor.tsx
```diff
@@ -228,7 +228,9 @@ export function PageEditor({id}: PageEditorProps) {
...
+      ? Alert.success(t('pageEditor.overview.pagePublished'), 2000)
+      : Alert.success(t('pageEditor.overview.pagePending'), 2000)
```
roosnic1, line 233: You can change here `Alert` also to `Notification`. I think this got forgotten in the past.

roosnic1 (changes requested):
Looks good but needs a few corrections

packages/editor/src/client/routes/pageEditor.tsx
```diff
@@ -160,6 +160,7 @@ export function PageEditor({id}: PageEditorProps) {
     }
   }, [pageData, hasChanged])
 
+  // Change this with notification ?
   useEffect(() => {
     const error = createError?.message ?? updateError?.message ?? publishError?.message
     if (error) Alert.error(error, 0)
```
camilleory, line 166: This is set with `Notification` in articleEditor. Should it be changed here too ?

packages/editor/src/client/routes/pageEditor.tsx
```diff
@@ -228,12 +235,23 @@ export function PageEditor({id}: PageEditorProps) {
...
+        })
   }
 
   useEffect(() => {
     if (isNotFound) {
-      Alert.error(t('pageEditor.overview.pageNotFound'), 0)
+      Notification.error({
+        title: t('pageEditor.overview.pageNotFound'),
+        duration: 0
+      })
```
camilleory, line 254: `Duration` set to `0` here - is that on purpose ?

camilleory (commented):
@roosnic1 I changed the alert to notification in `pagesEditor` - added some comments where I have questions.

packages/editor/src/client/locales/en.json
```diff
@@ -3,6 +3,7 @@
     "articleEditor": {
       "overview": {
         "articlePublished": "Article Published",
+        "articlePending": "Article pending",
```
penk85, line 6: Really small comment, maybe change it to Pending (capital) just so it's consistent with the Article Published format.

camilleory: Done

lakeside763 (commented):
This looks good

packages/editor/src/client/routes/articleEditor.tsx
```diff
@@ -379,7 +379,7 @@ export function ArticleEditor({id}: ArticleEditorProps) {
 
       setChanged(false)
 
-      publishData === new Date() || publishDate < new Date()
+      publishDate <= new Date()
```
roosnic1, line 382: just a little suggestion since the translation key is the only thing changing between the if and else you could also write this like:
```
Notification.success({
  title: t(publishDate <= new Date() ? 'articleEditor.overview.articlePublished' : 'articleEditor.overview.articlePending'),
  duration: 2000
})
```
This way it's a bit more clear that only the message changes

packages/editor/src/client/routes/pageEditor.tsx
roosnic1: @camilleory no. In articleEdtior `Notification` should be replaced with `Alert`.

packages/editor/src/client/routes/pageEditor.tsx
roosnic1: this should also be an `Alert` and the idea is that error alerts should be visible until the user clicks them away

roosnic1 (changes requested):

roosnic1 (approved):
🎉

penk85 (approved):

bishoymelek (approved):
🎉

PR #401, opened 2021-11-04
BAJ-5_upload_size

michael-scheurer:
This features reduces images larger than a certain size (configurable by ENV variable) automatically on client side before uploading to the api. A user hint is provided. This features was asked from Bajour.
Any review is appreciated.


config/create-gke-manifest.js
```diff
@@ -962,7 +962,7 @@ async function applyEditor() {
                 {
                   name: 'PEER_BY_DEFAULT',
                   value: 'true'
-                }
+                },
```
roosnic1: You should add the environment variable here.
```
{
   name: 'IMG_MIN_SIZE_TO_RESIZE',
   value: '10'
}
```

roosnic1: Also maybe rename the variable to `IMG_MIN_SIZE_TO_COMPRESS` for clarity

packages/editor/jest.config.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
     }
   },
   testMatch: ['**/__tests__/specs/**/*.+(ts|tsx|js)'],
+  transformIgnorePatterns: [
```
roosnic1: is this needed?

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -193,6 +195,62 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
...
+   * Resizes an image on client side, if larger than 10 MB, which is the upload max of the api
+   * @param file
+   */
+  async function resizeImage(file: File): Promise<File> {
+    // only resize image if larger than 10 MB
+    if (!willImageResize(file)) {
+      return file // do not resize
+    }
+    const originalFileName: string = file.name
+    if (!originalFileName) {
```
roosnic1, line 208: I would avoid throwing errors here. I'm not sure if it is even possible for a file not to have a name, but if there is no name i would just create a random name e.g. "unnamed_image_%timestamp%". You might need to get the right file extension for the name.

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -193,6 +195,62 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
     }
   }
 
+  /**
+   * Resizes an image on client side, if larger than 10 MB, which is the upload max of the api
+   * @param file
+   */
+  async function resizeImage(file: File): Promise<File> {
+    // only resize image if larger than 10 MB
```
roosnic1, line 203: comment is not right.

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -193,6 +195,62 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
...
+    if (!willImageResize(file)) {
+      return file // do not resize
+    }
+    const originalFileName: string = file.name
+    if (!originalFileName) {
+      throw new Error('No file name provided!')
+    }
+    const options = {
+      targetSize: 10, // the max size in MB, defaults to 2MB
```
roosnic1, line 212: I would compress it to the the size that we defined in the environment variable. If i set the minimum to 50mb and then upload an image of 51mb i don't want it resized to 10mb.

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -238,6 +296,7 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
                 <DescriptionListItem label={t('images.panels.fileSize')}>
                   {prettyBytes(fileSize)}
                 </DescriptionListItem>
+                {resizeImageHint(file!)}
```
roosnic1, line 299: I'm not sure if the the `resizeImageHint` is at the right place. If i were uploading images i want to have the information that my image will be compressed above a certain size before i upload it. I suggest to add the hint below the upload field in `packages/editor/src/client/panel/imageUploadPanel.tsx`. Also include the size after which images will be compressed in the hint. Our i18n translation service allows dynamic parameters.

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -193,6 +195,62 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
...
+  }
+
+  /**
+   * Decide whether an image will be automatically resized or not. The limit is 10 MB which
+   * is the api's max image size
+   * @param file
+   */
+  function willImageResize(file: File) {
+    const originalFileSize: number = file.size / (1024 * 1024)
+    const envMinSize: string | undefined = process.env.IMG_MIN_SIZE_TO_RESIZE
```
roosnic1, line 231: This doesn't work. In the Client/Browser context you don't have access to `process`. You have to pass the variable in the server context `packages/editor/src/server/index.tsx` to the `clientSettings` on line 28. Afterwards you can access the variable like this:
```
const {variableName}: ClientSettings = JSON.parse(
    document.getElementById(ElementID.Settings)!.textContent!
  )
```

packages/editor/package.json
```diff
@@ -55,6 +55,7 @@
     "@wepublish/karma.run-react": "^0.2.0",
     "apollo-upload-client": "^16.0.0",
     "array-move": "^2.2.1",
+    "client-compress": "^2.2.2",
```
roosnic1, line 58: The library looks better maintained but this is a typescript project and i would suggest this library https://www.npmjs.com/package/browser-image-compression

roosnic1 (changes requested):
In an earlier version you also added the environment variable in the `config/create-gke-manifest.js`. Why did you remove that?

PR looks on the right way but still needs some adjustments.

config/create-gke-manifest.js
michael-scheurer: the e2e test failed, when I added it. And passed, when I removed it.

packages/editor/jest.config.js
michael-scheurer: nope. removed it.

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: ok

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: yeah. of course.

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: yeah. true!

config/create-gke-manifest.js
michael-scheurer: done.

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: done

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: I optimized the hint right beneath the image preview before uploading it to the server. Added the expected file size in the translation.

packages/editor/package.json
michael-scheurer: changed the library. good catch.

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -274,6 +276,7 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
                 />
               )}
             </Panel>
+            {resizeImageHint(file!)}
```
roosnic1, line 279: This leads to an error if i click an Image in the "Image Library" and the `imageEditPanel` opens.

Error:
image

My suggestion is to display the hint about image resizing always BUT again not in the `imageEditPanel`. This information is needed at the time of upload which would be in the `imageUploadPanel`. Right next to the `<FileDropInput ... />` component.

packages/editor/src/server/index.tsx
```diff
@@ -27,7 +27,8 @@ async function asyncMain() {
 
   const clientSettings = {
     apiURL: process.env.API_URL,
-    peerByDefault: process.env.PEER_BY_DEFAULT?.toLowerCase() === 'true'
+    peerByDefault: process.env.PEER_BY_DEFAULT?.toLowerCase() === 'true',
+    imgMinSizeToCompress: process.env.IMG_MIN_SIZE_TO_COMPRESS
```
roosnic1, line 31: To avoid `imgMinSizeToCompress` to be `string | undefined` i would add a default value here if the environment variable is undefined. Also i suggest to use `number` instead of `string`. Since we're passing those settings via JSON to the client we can already use the proper type.

So this line could look like:
`imgMinSizeToCompress: parseInt(process.env.IMG_MIN_SIZE_TO_COMPRESS) || 10`

roosnic1 (changes requested):
Looks much better. Just some final changes.

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: The time of upload is not in the imageUploadPanel ;) But I moved it anyway to the imageUploadPanel. Looks good for me as well.

packages/editor/src/server/index.tsx
michael-scheurer: yeah. I improved it with an ternary operator since parseInt() doesn't like undefined that much ;)

roosnic1 (dismissed):
👍🏾

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -193,6 +195,37 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
...
   }
 
+  /**
+   * Resizes an image on client side, if larger than the IMG_MIN_SIZE_TO_COMPRESS env variable
+   * @param file
+   */
+  async function resizeImage(file: File): Promise<File> {
+    const imgMinSizeToCompress: number = getImgMinSizeToCompress()
+    // only resize image if larger than IMG_MIN_SIZE_TO_COMPRESS env variable
+    // ATTENTION: Media server must be configured to accept the IMG_MIN_SIZE_TO_COMPRESS
```
tomaszdurka, line 205: Hi Michael! 

Regarding this comment - why do we nee that on the Media server? To clarify intention: I understand this code is fully browser-based and mentioned env var needs to be set on the Editor app, right?

camilleory (dismissed):
One question: would it not be relevant to add the message info about the image size in the article and page image upload panel too ? 
image

packages/editor/src/client/panel/imageEditPanel.tsx
michael-scheurer: Your absolutely right. This code is fully browser based. If I'm remembering right, this comment is Nico's. And his intention is probably the following: If you configure the IMG_MIN_SIZE_TO_COMPRESS larger than the upload max on the server side, the server will refuse the upload and throw an error. So theses settings should be in sync for the best.

penk85 (approved):

camilleory (approved):

lakeside763 (approved):

tomaszdurka (approved):

PR #402, opened 2021-11-04
changed paymentProviderCustomer to an array

roosnic1:


penk85 (dismissed):
:+1:

packages/api-db-mongodb/src/migration.ts
```diff
@@ -613,6 +614,33 @@ export const Migrations: Migration[] = [
...
+      const users = await db.collection(CollectionName.Users)
+
+      const allUsers: DBUser[] = await users.find({}).toArray()
+
+      for (const user of allUsers) {
+        const paymentProvidersCustomersArray: PaymentProviderCustomer[] = []
+        const paymentProviderCustomers = Object.keys(user.paymentProviderCustomers)
+        paymentProviderCustomers.forEach(ppc => {
+          // @ts-ignore
+          const temp = user.paymentProviderCustomers[ppc]
```
camilleory, line 631: the name of the `temp `variable is not very descriptive, could it be more self-explanatory or can you explain what it describes?

packages/api/src/db/user.ts
```diff
@@ -117,8 +117,8 @@ export interface DeleteUserSubscriptionArgs {
 }
 
 export interface PaymentProviderCustomer {
-  readonly id: string
-  readonly createdAt: Date
```
camilleory, line 121: Is `createdAt` not relevant anymore ?

camilleory (changes requested):

packages/api/src/db/user.ts
roosnic1: Yeah i added it in the past but realized now that it was premature optimization and we don't need it.

camilleory (approved):

penk85 (approved):

lakeside763 (approved):

bishoymelek (commented):
Does this has anything to do with the Editor or Public API?

PR #409, opened 2021-11-10
F/wpc-574 rename author and source in img library

camilleory:
In this branch I changed author to source and source to link in the image edit panel in the 3 languages.

image

test **[here](https://editor.fwpc574im.wepublish.dev/)**


camilleory:
Only the translation should be updated, not the variable names (renaming provokes problems, we would need to make sure all old fields still work). I'll open a new pull request and update the translation update.

PR #413, opened 2021-11-11
f/wpc-576: highlight slug field and disable publish button when no slug

camilleory:
The slug field gets a red highlight and the publish button is disabled when the article doesn't have a slug
The test has been adapted.

image


packages/editor/src/client/api/index.ts
```diff
@@ -1731,6 +1737,7 @@ export type User = {
   properties: Array<Properties>;
   roles: Array<UserRole>;
   subscription?: Maybe<UserSubscription>;
+  paymentProviderCustomers: Array<PaymentProviderCustomer>;
```
penk85, line 1740: What's this from?

camilleory: I think it comes from Nico's branch (the one he updated the payment type). Not sure why it comes in my code. @roosnic1 is it problematic ?

lakeside763 (approved):

packages/editor/src/client/api/index.ts
roosnic1: @penk85 and @camilleory  Yeah i forgot to commit that file. Is fixed here https://github.com/wepublish/wepublish/pull/420 please review and once merged this should be gone.

penk85 (approved):
Looks great

roosnic1 (changes requested):
I recommend to wait with this PR and reuse the component from https://github.com/wepublish/wepublish/pull/418

PR #417, opened 2021-11-16
F/wpc 578 highlight missing field article metadata

camilleory:
- Fields in article publish panel get highlighted in orange when empty.
- Slug field highlighted in red when empty
- Publish button disabled when slug is missing
- Test updated

image


packages/editor/src/client/locales/en.json
```diff
@@ -56,6 +56,15 @@
         "emptyLead": "Leave empty to use original lead.",
         "emptyPreTitle": "Leave empty to use original pre-title",
         "emptyTitle": "Leave empty to use original title",
+        "enterImage": "Please add an image",
+        "enterTitle": "Please enter a title",
+        "enterSeoTitle": "Please enter a seo title",
+        "enterSocialMediaDescription": "Please enter a description for social media",
+        "enterSocialMediaAuthors": "Please enter an author for social media",
```
bishoymelek, line 63: Add/Select an author?

packages/editor/src/client/locales/en.json
```diff
@@ -56,6 +56,15 @@
...
+        "enterSocialMediaTitle": "Please enter a title for social media",
+        "enterSocialMediaImage": "Please enter an image for social media",
```
bishoymelek, line 65: same as above I guess.

packages/editor/src/client/panel/publishArticlePanel.tsx
```diff
@@ -64,20 +70,52 @@ export function PublishArticlePanel({
...
+            )}
           </DescriptionListItem>
+
           <DescriptionListItem label={t('articleEditor.panels.slug')}>
             {metadata.slug || '-'}
           </DescriptionListItem>
+
           <DescriptionListItem label={t('articleEditor.panels.tags')}>
-            {metadata.tags.join(', ') || '-'}
+            {!metadata.tags.join(', ') ? (
```
bishoymelek, line 104: Do you mean checking if the array has values and not empty?
```
{!metadata.tags.length ? (
```

bishoymelek (commented):
Great job @camilleory 🎉 , I would suggest in this case you check first if values exist and not other way around like:
```
{metadata.title ? (
   metadata.slug
) : (
   <div style={missingInfoStyle}>{t('articleEditor.panels.enterTitle')}</div>
)}
```
And for the arrays,If i understand correctly .join actually do mapping, so instead of "checking" using it in an if statement, you check if the array has values `.length`

packages/editor/src/client/locales/en.json
```diff
@@ -56,6 +56,15 @@
         "emptyLead": "Leave empty to use original lead.",
         "emptyPreTitle": "Leave empty to use original pre-title",
         "emptyTitle": "Leave empty to use original title",
+        "enterImage": "Please add an image",
+        "enterTitle": "Please enter a title",
+        "enterSeoTitle": "Please enter a seo title",
+        "enterSocialMediaDescription": "Please enter a description for social media",
+        "enterSocialMediaAuthors": "Please add an author for social media",
+        "enterSocialMediaTitle": "Please enter a title for social media",
+        "enterSocialMediaImage": "Please enter an image for social media",
```
lakeside763, line 65: @camilleory How about "Please add an image for social media"

lakeside763 (dismissed):
This looks good

bishoymelek (changes requested):
Almost, have same feedback as @lakeside763

packages/editor/src/client/locales/en.json
camilleory: Done

packages/editor/src/client/atoms/descriptionList.tsx
```diff
@@ -44,7 +45,16 @@ export function DescriptionListItem({label, children}: DescriptionListItemProps)
...
+        {children || (
+          <div
+            style={{
+              borderRadius: '8px',
+              padding: '6px',
+              backgroundColor: '#FFEBCD'
+            }}>
+            {message}
+          </div>
+        )}
```
lakeside763, line 57: @camilleory From line 48 - 57 May be it can be structured as follow
{message 
  ?     (<div
            style={{
              borderRadius: '8px',
              padding: '6px',
              backgroundColor: '#FFEBCD'
          }}>
            {message}
          </div>) 
     : (<span>{children}<span>)}

lakeside763 (commented):
Just a point to look at

packages/editor/src/client/atoms/descriptionList.tsx
```diff
@@ -44,7 +45,16 @@ export function DescriptionListItem({label, children}: DescriptionListItemProps)
         style={{
           marginLeft: 20
         }}>
-        {children}
+        {children || (
```
roosnic1, line 48: what happens if `children` and `message` are empty? Also you need a way to change the background color if you want to use the same component for error messages.

I suggest to create a new component that reuses the `<DescriptionListItem>`

packages/editor/src/client/atoms/descriptionList.tsx
```diff
@@ -44,7 +45,16 @@ export function DescriptionListItem({label, children}: DescriptionListItemProps)
...
+          <div
+            style={{
+              borderRadius: '8px',
+              padding: '6px',
+              backgroundColor: '#FFEBCD'
```
roosnic1, line 53: Since we use rsuite.js i recommend to use those components. Maybe https://rsuitejs.com/components/message/ could be used here. If not we should at least reuse the colors for warnings and errors.

roosnic1 (changes requested):
Looks good but does need some improvements.

packages/editor/src/client/atoms/descriptionList.tsx
camilleory: I made an attempt with a new component, look forward to having your review

packages/editor/src/client/atoms/descriptionList.tsx
camilleory: I didn't use the rsuite component because the message was huge and didn't look good in the publish panel but I used its colors

packages/editor/src/client/atoms/infoMessage.tsx
```diff
@@ -0,0 +1,30 @@
+import React, {ReactNode} from 'react'
+
+export interface InfoMessageProps {
+  type?: ReactNode
+  message: ReactNode
+}
+
+export function InfoMessage({type, message}: InfoMessageProps) {
+  if (type === 'warning') {
+    type = '#fffaf2'
```
bishoymelek, line 10: I think this is not recommended, you are mutating the prop(param), I think you can have 
```
let msgBgColor = '#ffffff'
``` 
for example and assign values to it.

-------------
You can also have colors as enum, if doesn't exist somewhere else, like:
```
export enum MessageColor {
  warning = '#fffaf2'
  error = '#fff2f2'
}
```

packages/editor/src/client/atoms/infoMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+  } else if (type === 'error') {
+    type = '#fff2f2'
+  } else {
+    type = '#ffffff'
+  }
+
+  return (
+    <div>
```
bishoymelek, line 18: do you need two divs?

bishoymelek (changes requested):
Great one 👍 just tiny comments.

packages/editor/src/client/atoms/descriptionList.tsx
```diff
@@ -20,9 +21,11 @@ export function DescriptionList({children}: DescriptionListItemProps) {
 export interface DescriptionListItemProps {
   label?: ReactNode
   children?: ReactNode
+  message?: ReactNode
+  type?: ReactNode
```
penk85, line 25: Can you name this `messageType`? Just so it's clear what's being given a type here.

packages/editor/src/client/atoms/descriptionList.tsx
```diff
@@ -44,7 +47,11 @@ export function DescriptionListItem({label, children}: DescriptionListItemProps)
         style={{
           marginLeft: 20
         }}>
-        {children}
+        {message ? (
```
penk85, line 50: Is the warning message only intended to be used to highlight empty property fields? Would there ever be a use case of children being displayed with a warning message (invalid input maybe)? Because then I'd recommend to put both message and the children conditionally without the else statement.

camilleory: It's supposed to get highlighted when information is missing only, at least that's how it's described in the Jira ticket

penk85 (dismissed):
:+1:

bishoymelek (approved):
👌

packages/editor/src/client/atoms/descriptionListwithMessage.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function DescriptionListItemWithMessage({
+  label,
+  children,
+  message,
+  messageType
+}: DescriptionListWithMessageItemProps) {
+  return (
+    <div
```
roosnic1, line 18: instead of copying the html from `DescriptionListItem` component you could reuse it here.

```
<DescriptionListItem label={label}>
  {children || <InfoMessage messageType={messageType} message={message}></InfoMessage>}
</DescriptionListItem>
```
Also you'll not need the `'-'` anymore.

roosnic1 (changes requested):
looks better just one last change.

packages/editor/src/client/atoms/infoMessage.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export function InfoMessage({messageType = InfoColor.white, message}: InfoMessageProps) {
+  return (
+    <div
+      style={{
+        borderRadius: '8px',
+        padding: '0px 6px',
+        backgroundColor: `${messageType}`
+      }}>
+      {/* {' '} */}
```
roosnic1, line 22: For what do you need that?

roosnic1 (commented):
Looks very good. Just one question.

packages/editor/src/client/atoms/infoMessage.tsx
camilleory: I cleaned it up

packages/editor/src/client/atoms/descriptionListwithMessage.tsx
```diff
@@ -0,0 +1,23 @@
+import React, {ReactNode} from 'react'
+import {InfoMessage, InfoColor} from './infoMessage'
+import {DescriptionListItem} from './descriptionList'
+
+export interface DescriptionListWithMessageItemProps {
+  label?: ReactNode
+  children?: ReactNode
+  message?: ReactNode
```
penk85, line 8: If this component is now meant specifically for messages, I think you should make the message required.

camilleory: That makes sense 👍

bishoymelek (approved):
🎉

penk85 (approved):
:+1:

roosnic1 (approved):
🎉

lakeside763 (approved):
This looks good

PR #423, opened 2021-11-22
Update of stripePaymentProvider

roosnic1:


packages/api/src/payments/stripePaymentProvider.ts
```diff
@@ -83,36 +86,53 @@ export class StripePaymentProvider extends BasePaymentProvider {
...
       ),
-      ...(props.customerID
+      ...(customerID && paymentMethod // only add customer if we have customer and paymentMethod
         ? {
             confirm: true,
-            customer: props.customerID,
+            customer: customerID,
             off_session: true,
-            payment_method: paymentMethod?.id,
+            payment_method: typeof paymentMethod === 'string' ? paymentMethod : paymentMethod?.id,
```
tomaszdurka, line 126: Took me a while you expect this multi-type paymentMethod - that's historical then. If this is always an `string` `id`, I would keep it as `string`/`null` only.

tomaszdurka: Plus even rename this to `paymentMethodId`.

tomaszdurka (changes requested):

penk85 (approved):

bishoymelek (approved):

camilleory (approved):

tomaszdurka (approved):

PR #429, opened 2021-11-26
manual checkIntentStatus for Stripe and Payrexx

roosnic1:
WPC-105. This PR implements the `checkIntentStatus` method for the Stripe and Payrexx paymentProvider. In addition to that a new public GraphQL mutation was added that allows to check if a specific invoice has been paid. Together with the `checkIntentStatus` this can now be checked. This is need if some one doesn't want to work/rely on webhooks or if the user needs to know the information immediately.


packages/api/src/graphql/mutation.public.ts
```diff
@@ -536,6 +538,49 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+      args: {
+        id: {type: GraphQLNonNull(GraphQLID)}
+      },
+      description:
+        'This mutation will check the invoice status and update with information from the paymentProvider',
+      async resolve(root, {id}, context) {
+        const {authenticateUser, dbAdapter, paymentProviders} = context
+        const {user} = authenticateUser()
+
+        const invoices = await dbAdapter.invoice.getInvoicesByUserID(user.id)
```
tomaszdurka, line 554: I've noticed you check presence of item in those arrays returned from database. Do we expect null/undefined here?

Same on line `562`, 565`, `570`.

packages/api/src/graphql/mutation.public.ts
```diff
@@ -536,6 +538,49 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+          const paymentProvider = paymentProviders.find(
+            pp => pp && pp.id === paymentMethod.paymentProviderID
+          )
+          if (!paymentProvider) continue // TODO: what happens if we don't find a paymentProvider
+
+          const intentState = await paymentProvider.checkIntentStatus({intentID: payment.intentID})
+          await paymentProvider.updatePaymentWithIntentState({intentState, context})
+        }
+
+        // FIXME: We need to implement a way to wait for all the database
```
tomaszdurka, line 578: I understand this is async operation from db and we expect invoice there after that? Shall we just add a loop/longpolling here?

packages/api/src/payments/paymentProvider.ts
```diff
@@ -94,6 +101,53 @@ export abstract class BasePaymentProvider implements PaymentProvider {
...
+        paymentData: intentState.paymentData,
+        intentData: payment.intentData,
+        intentSecret: payment.intentSecret,
+        intentID: payment.intentID,
+        invoiceID: payment.invoiceID,
+        paymentMethodID: payment.paymentMethodID
+      }
+    })
+
+    if (!updatedPayment) throw new Error('Error while updating Payment')
```
tomaszdurka, line 126: General question: Should these errors not come out of `updatePayment` method?

packages/api/src/payments/paymentProvider.ts
```diff
@@ -94,6 +101,53 @@ export abstract class BasePaymentProvider implements PaymentProvider {
...
+
+    if (intentState.customerID && payment.invoiceID) {
+      const invoice = await context.loaders.invoicesByID.load(payment.invoiceID)
+      if (!invoice?.userID)
+        throw new Error(`Invoice with ID ${payment.invoiceID} does not have a userID`)
+
+      const user = await context.dbAdapter.user.getUserByID(invoice.userID)
```
tomaszdurka, line 133: Shouldn't `get` always return value otherwise throw? Wdyt about convention `get` returns object/throws, `find` returns object/null.

packages/api/src/payments/payrexxPaymentProvider.ts
```diff
@@ -111,18 +111,46 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
...
+        method: 'GET'
+      }
+    )
+    if (res.status !== 200) throw new Error(`Payrexx response is NOK with status ${res.status}`)
+
+    const payrexxResponse = await res.json()
+    const [gateway] = payrexxResponse.data
+    if (!gateway) throw new Error(`Payrexx didn't return a gateway`)
+
+    const state = mapPayrexxEventToPaymentStatue(gateway.status)
```
tomaszdurka, line 129: Typo? `Statue` :)

packages/api/src/graphql/mutation.public.ts
```diff
@@ -536,6 +538,49 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+        id: {type: GraphQLNonNull(GraphQLID)}
+      },
+      description:
+        'This mutation will check the invoice status and update with information from the paymentProvider',
+      async resolve(root, {id}, context) {
+        const {authenticateUser, dbAdapter, paymentProviders} = context
+        const {user} = authenticateUser()
+
+        const invoices = await dbAdapter.invoice.getInvoicesByUserID(user.id)
+        const invoice = invoices.find(invoice => invoice !== null && invoice.id === id)
```
bishoymelek, line 555: What do you think about: 
```
const invoice = invoices.find(invoice => invoice?.id === id)
```

packages/api/src/graphql/mutation.public.ts
```diff
@@ -536,6 +538,49 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+        if (!invoice) throw new NotFound('Invoice', id)
+
+        const payments = await dbAdapter.payment.getPaymentsByInvoiceID(invoice.id)
+        const paymentMethods = await dbAdapter.paymentMethod.getActivePaymentMethods()
+
+        for (const payment of payments) {
+          if (!payment || !payment.intentID) continue
```
bishoymelek, line 562: Maybe:
```
if (!payment?.intentID) continue
```

packages/api/src/payments/stripePaymentProvider.ts
```diff
@@ -150,18 +158,43 @@ export class StripePaymentProvider extends BasePaymentProvider {
...
+      throw new Error('empty paymentID')
+    }
+
+    let customerID
+    if (
+      intent.setup_future_usage === 'off_session' &&
+      intent.customer === null &&
+      intent.payment_method !== null
+    ) {
+      customerID = await this.createStripeCustomer({intent})
```
tomaszdurka, line 190: Would be great to talk it over. I understand you `create customer` within a `check` method?

packages/api/src/graphql/mutation.public.ts
```diff
@@ -536,6 +538,49 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+
+          const paymentMethod = paymentMethods.find(
+            pm => pm && payment && pm.id === payment.paymentMethodID
```
bishoymelek, line 565: I think this is enough for this phase? 
```
pm => pm?.id === payment?.paymentMethodID
```

packages/api/src/graphql/mutation.public.ts
```diff
@@ -536,6 +538,49 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+          )
+          if (!paymentMethod) continue // TODO: what happens if we don't find a paymentMethod
+
+          const paymentProvider = paymentProviders.find(
+            pp => pp && pp.id === paymentMethod.paymentProviderID
```
bishoymelek, line 570: same

tomaszdurka (approved):
I approve with small comments on conventions as I assume it's tested and urgent. The best would be to talk this over in-person.

bishoymelek (commented):
🎉 just left simple suggestion

packages/api/src/graphql/mutation.public.ts
roosnic1: In this case we don't. But there are methods that accept a list of ids and then return the found models. There it could be that some of the items could be `null`. In any case we use the same type `OptionalInvoice` and that's why i have to check.

packages/api/src/graphql/mutation.public.ts
roosnic1: It's not the database operation. It's the event hook afterwards that updates other models for which we wait and i think we should have a better solution then just a loop.

packages/api/src/payments/paymentProvider.ts
roosnic1: Good question. Might be something to address in the new database abstraction layer.

packages/api/src/payments/paymentProvider.ts
roosnic1: I like that. Same comment as above. Something for the database abstraction layer discussion.

packages/api/src/payments/stripePaymentProvider.ts
roosnic1: Agree. Something we can discuss for refactoring.

packages/api/src/graphql/mutation.public.ts
roosnic1: it's not the same here. Here i'm checking for undefined

packages/api/src/payments/payrexxPaymentProvider.ts
```diff
@@ -111,18 +111,46 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
...
+  async checkIntentStatus({intentID}: CheckIntentProps): Promise<IntentState> {
+    const signature = crypto.createHmac('sha256', this.instanceAPISecret).digest('base64')
+
+    const res = await fetch(
+      `https://api.payrexx.com/v1.0/Gateway/${intentID}/?instance=${this.instanceName}&ApiSignature=${signature}`,
+      {
+        method: 'GET'
+      }
+    )
+    if (res.status !== 200) throw new Error(`Payrexx response is NOK with status ${res.status}`)
```
lakeside763, line 123: Could this be
```
`Payrexx response is NOT with status ${res.status}
```

lakeside763 (approved):

camilleory (approved):

penk85 (approved):

packages/api/src/graphql/mutation.public.ts
tomaszdurka: So you ok with single 100ms delay for now? If so let's go. Otherwise would go for loop/delay maybe.

packages/api/src/payments/paymentProvider.ts
tomaszdurka: OK.

packages/api/src/payments/payrexxPaymentProvider.ts
tomaszdurka: Nico, I understood you meant `NOK = NOT OK`, right? Moses you ok with that?

tomaszdurka (approved):

packages/api/src/payments/payrexxPaymentProvider.ts
roosnic1: yes

packages/api/src/graphql/mutation.public.ts
roosnic1: For now it should be enough. If it wouldn't be enough then calling the mutation again would return the correct result.

PR #433, opened 2021-11-30
Use mailproviders with templates

roosnic1:
This PR implements the template function of mailchimp and mailgun.


penk85 (approved):

packages/api/__tests__/specs/mailchimpMailProvider.ts
```diff
@@ -39,10 +50,48 @@ describe('Mailchimp Mail Provider', () => {
...
+  expect(mockSendTemplate).toHaveBeenCalledTimes(1)
+  expect(mockSendTemplate).toHaveBeenCalledWith(
+    {
+      template_name: 'test-mail',
+      template_content: [],
       message: {
         html: undefined,
-        text: 'hello Test',
+        text: undefined,
         subject: 'test subject',
```
camilleory, line 95: Why undefined here ?

camilleory (approved):

packages/api/__tests__/specs/mailchimpMailProvider.ts
roosnic1: If i send a mail with a template from Mailchimp i don't need any text. All the text will be in the template from mailchimp

PR #436, opened 2021-12-01
payment method with slugs

roosnic1:
This PR adds a slug field to the paymentMethods. It also allows two GraphQL mutation to accept the slug instead of just the paymentMethodID.


lakeside763 (approved):

penk85 (approved):

camilleory (approved):

tomaszdurka (approved):
Looks good. I understand you still support ids additionally to slugs because backwards compatibility, right?

PR #449, opened 2021-12-07
Some subscription changes

roosnic1:
This PR moves the mutation `checkInvoiceStatus` to GraphQL query. This PR also creates a new memberContext method called handleSubscriptionChange` which handles all the necessary steps after a subscription is changed. This methods gets called in the public and private `updateSubscription`.


camilleory (approved):

packages/api/src/utility.ts
```diff
@@ -166,6 +166,7 @@ export const MAX_COMMENT_LENGTH = 1000
 export const ONE_HOUR_IN_MILLISECONDS = 60 * 60 * 1000
 export const ONE_DAY_IN_MILLISECONDS = 24 * ONE_HOUR_IN_MILLISECONDS
 export const FIFTEEN_MINUTES_IN_MILLISECONDS = 900000
+export const ONE_MONTH_IN_MILLISECONDS = 31 * ONE_DAY_IN_MILLISECONDS
```
penk85, line 169: Won't monthly billing be inconsistent with a static 31 days, or is this how it's meant to be?

roosnic1: @penk85 I just need it to compare if roughly a month has passed. I went for 31 because that means it's a month for sure.

penk85 (approved):

PR #451, opened 2021-12-07
catch stripe errors and send mail

roosnic1:
This PR catches errors during `off_session` payments (currently on possible with the `StripePaymentProvider`) and sends a mail with the error code to the user. The PR also catches `off_session` payments where the user doesn't have a stripe customer.


examples/api/templates/emails/memberSubscriptionPayment/offSessionPaymentFailed/html.pug
```diff
@@ -0,0 +1,7 @@
+extends ../../base.pug
+
+block content
+    h1 Hi #{user.name}
+
+    p We tried to bill your cc with #{paymentProviderID} and received the following error: "#{errorCode}"
```
tomaszdurka, line 6: This looks very technical with `#{paymentProviderId}`. Is provider sending email as well?

packages/api/src/memberContext.ts
```diff
@@ -293,11 +294,22 @@ export class MemberContext implements MemberContext {
...
           user.id
         )
+        await this.mailContext.sendMail({
+          type: SendMailType.MemberSubscriptionOffSessionFailed,
+          recipient: invoice.mail,
+          data: {
+            user,
+            invoice,
+            paymentProviderID: paymentMethod.paymentProviderID,
+            errorCode: 'cutomer_missing'
```
camilleory, line 304: Typo here ? `customer_missing` ?

tomaszdurka (dismissed):
I approve, but I feel like you could use a few more consts/functions which describe intention.

E.g.
```
if (
        intent.setup_future_usage === 'off_session' &&
        intent.customer === null &&
        intent.payment_method !== null
      )
```
Could say if it's because customer not exists yet, has no payment or what exactly. Current way force you to really deep dive imo.

lakeside763 (dismissed):

packages/api/src/payments/stripePaymentProvider.ts
```diff
@@ -119,29 +121,51 @@ export class StripePaymentProvider extends BasePaymentProvider {
...
+          : {}),
+        currency: 'chf',
+        // description: props.invoice.description, TODO: convert to text
+        ...(saveCustomer ? {setup_future_usage: 'off_session'} : {}),
+        metadata: {
+          paymentID: paymentID,
+          mail: invoice.mail
+        }
+      })
+    } catch (error) {
```
lakeside763, line 150: Maybe from line 127 to 148 can be broken into different variable for improved maintainability

examples/api/templates/emails/memberSubscriptionPayment/offSessionPaymentFailed/html.pug
roosnic1: @tomaszdurka this is just as an example anyway. Actual we.publish user could do an `if` depending on the `paymentProviderId`

camilleory (approved):

packages/api/src/payments/stripePaymentProvider.ts
```diff
@@ -119,29 +121,51 @@ export class StripePaymentProvider extends BasePaymentProvider {
...
+      logger('stripePaymentProvider').error(
+        error,
+        'Error while creating Stripe Intent %s for paymentProvider %s',
+        this.id
```
penk85, line 154: This is a syntax question, do you need a second string here?

roosnic1: Yeah. My mistake didn't change the message.

penk85 (approved):

camilleory (approved):

PR #458, opened 2021-12-09
better logging

roosnic1:


lakeside763 (approved):

packages/api/src/payments/payrexxPaymentProvider.ts
```diff
@@ -125,7 +125,15 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
         method: 'GET'
       }
     )
-    if (res.status !== 200) throw new Error(`Payrexx response is NOK with status ${res.status}`)
+    if (res.status !== 200) {
+      logger('payrexxPaymentProvider').error(
+        res,
```
tomaszdurka, line 130: Is logging whole `res` is actually enough ? I believe body will be `readableStream` and it's not gonna log it's content but rather object. Please confirm you've checked it.
Otherwise you can write `status`, `statusText` and possibly `body()`.

penk85 (approved):

packages/api/src/payments/payrexxPaymentProvider.ts
```diff
@@ -125,7 +125,15 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
...
+        'Payrexx response for intent %s is NOK with status %s',
+        intentID,
```
tomaszdurka, line 132: What do you think about logging object/hash instead of flat values. 
```
logger('payrexxPaymentProvider').error({
  res,
  message: 'Payrexx response for intent %s is NOK with status %s',
  intentID,
  status: res.status
})
```
I did that in previous project and usually what logging viewers do with that is they present message plus you can expand json payload when needed.

tomaszdurka: Still not sure if that's a place for this change - maybe a general rework of logging in another PR. @roosnic1 please share your thoughts.

tomaszdurka (changes requested):

tomaszdurka (approved):

PR #460, opened 2021-12-13
F/wpc-585-pagination-image-library

camilleory:
Pagination added to image library
- `imageList` is now organised as a table (instead of FlexGrid), as it wasn't possible to implement pagination on the image library on flexgrid with the version of rsuite we're currently using. 

image


penk85 (dismissed):
Looks really nice! :+1:

packages/editor/src/client/routes/imageList.tsx
```diff
@@ -32,36 +32,44 @@ import {
...
   const {current} = useRoute()
   const dispatch = useRouteDispatch()
+  const [images, setImages] = useState<FullImageFragment[]>([])
+
+  const [filter, setFilter] = useState('')
+
+  const [isConfirmationDialogOpen, setConfirmationDialogOpen] = useState(false)
+  const [currentImage, setCurrentImage] = useState<ImageRefFragment>()
+
+  const [activePage, setActivePage] = React.useState(1)
```
penk85, line 56: Its in the imports so you shouldn't need the `React.` here I think.

camilleory: Good point thanks

penk85 (approved):

tomaszdurka (approved):

PR #469, opened 2021-12-15
Member context improvements

roosnic1:
Added new job `dailyInvoiceChecker`  that checks the intent state of all payments on open invoices.
 
The job `chargeOpenInvoices` now charges open invoices only if the user is active.

The job `dailyInvoiceReminder` only sends mail if the user still exists and is active.

This PR also corrects the calculation for the end date of periods when renewing subscriptions.

### DIFF
--- a/examples/api/src/index.ts
+++ b/examples/api/src/index.ts
@@ -432,6 +432,17 @@ async function asyncMain() {
         await server.runJob(JobType.DailyInvoiceCharger, {})
         process.exit(0)
       }
+    )
+    .command(
+      'checkdic',
+      'check open invoices',
+      () => {
+        /* do nothing */
+      },
+      async () => {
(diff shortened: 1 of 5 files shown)


penk85 (approved):

camilleory (approved):

examples/api/src/index.ts
```diff
@@ -432,6 +432,17 @@ async function asyncMain() {
         await server.runJob(JobType.DailyInvoiceCharger, {})
         process.exit(0)
       }
+    )
+    .command(
+      'checkdic',
+      'check open invoices',
+      () => {
+        /* do nothing */
```
tomaszdurka, line 440: Cool

tomaszdurka (approved):

PR #474, opened 2021-12-22
f/wpc-606 image edit panel layout

camilleory:
- Description section moved to panel bottom
- Fields order in  information section changed 
- Bigger field for image description
- Save and close button instead of two buttons (save, close)
Dec-23-2021 09-42-11

### DIFF
--- a/packages/editor/src/client/panel/imageEditPanel.tsx
+++ b/packages/editor/src/client/panel/imageEditPanel.tsx
@@ -1,6 +1,5 @@
 import React, {useState, useEffect} from 'react'
 import prettyBytes from 'pretty-bytes'
-import '../global.less'
 
 import {
   useUploadImageMutation,
--- a/packages/editor/src/client/panel/imageEditPanel.tsx
+++ b/packages/editor/src/client/panel/imageEditPanel.tsx
@@ -240,7 +240,7 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
       <Drawer.Body>
         {!isLoading && (
           <>
-            <Panel style={{backgroundColor: 'dark'}} className={'noYPadding'}>
+            <Panel style={{backgroundColor: 'dark'}}>
               {imageURL && imageWidth && imageHeight && (
                 <FocalPointInput
                   imageURL={imageURL}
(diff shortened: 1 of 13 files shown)


packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -357,9 +363,6 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
         <Button appearance={'primary'} disabled={isDisabled} onClick={() => handleSave()}>
           {isUpload ? t('images.panels.upload') : t('images.panels.save')}
         </Button>
-        <Button appearance={'subtle'} onClick={() => onClose?.()}>
```
roosnic1, line 360: I would keep this button. Otherwise people only have the chance of the little x in the top right corner to cancel their changes. I recommend to change the appearance to red.

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -249,48 +251,9 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
                 />
               )}
             </Panel>
-            <Panel header={t('images.panels.description')}>
```
roosnic1, line 252: Instead of moving it down you can could also use the collapse function of the panels. https://rsuitejs.com/components/panel/#collapsible-panel-group

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -302,11 +265,22 @@ export function ImagedEditPanel({id, file, onClose, onSave}: ImageEditPanelProps
...
                   <FormControl
+                    rows={5}
+                    name="textarea"
+                    componentClass="textarea"
                     value={description}
                     disabled={isDisabled}
                     onChange={value => setDescription(value)}
                   />
                 </FormGroup>
+                <FormGroup>
```
roosnic1, line 276: Is this a request that users can change file names? I think it's rather confusing having an image name and a file name. I'd prefer it if we don't allow this to change and also don't display it so prominently.

roosnic1 (commented):
Some ideas

tomaszdurka:
Hey @camilleory, what's the story with this one?
I propose to close it and reopen if we focus back on wpc-606.

Issue #110, opened 2020-12-16
Make login form submit via enter


roosnic1:
**Is your feature request related to a problem? Please describe.**
Pressing enter while the username or password field have the focus does not submit the login form.

**Describe the solution you'd like**
If either the username or password field have focus and the user presses enter the login form should be submitted.

Itrulia:
Hey hey Nico! I just went through the "good first issue" issues and found this one. It seems like this is already the case now, can this be closed or is there still anything missing?

PR #476, opened 2021-12-28
F/submit via enter

treflxyz:
#110
Hi @tomaszdurka  @roosnic1, could you have a look at this pull request ?

### DIFF
--- a/packages/editor/src/client/login.tsx
+++ b/packages/editor/src/client/login.tsx
@@ -211,7 +211,7 @@ export function Login() {
                 className={'username'}
                 value={email}
                 autoComplete={'username'}
-                onChange={email => {console.log('em', email); setEmail(email)}}
+                onChange={(email: string) => setEmail(email)}
                 inputRef={(ref: HTMLInputElement): void => {
                   emailRef.current = ref
                 }}
@@ -224,7 +224,7 @@ export function Login() {
                 type="password"
                 value={password}
                 autoComplete={'currentPassword'}
-                onChange={password => {console.log('pas', password); setPassword(password)}}
+                onChange={(password: string)  => setPassword(password)}
                 inputRef={(ref: HTMLInputElement): void => {
                   passwordRef.current = ref
                 }}
--- a/packages/editor/src/client/login.tsx
+++ b/packages/editor/src/client/login.tsx
@@ -174,7 +174,7 @@ export function Login() {
     }
   }, []);
 
-  const listener = (e): void => {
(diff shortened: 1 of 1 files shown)


DamianNode:
I add missing type for event in listener functions.

roosnic1:
@DamianNode you should run `yarn build` before you commit.

DamianNode:
Now build is good.

packages/editor/src/client/login.tsx
```diff
@@ -193,7 +225,10 @@ export function Login() {
                 type="password"
                 value={password}
                 autoComplete={'currentPassword'}
-                onChange={password => setPassword(password)}
+                onChange={(password: string) => setPassword(password)}
+                inputRef={(ref: HTMLInputElement): void => {
```
roosnic1, line 229: Instead of `inputRef` i suggest to use `onKeyDown`. Makes the whole thing a bit simpler and you don't require a document event listener.

roosnic1 (changes requested):
Thank you very much for the contribution. I added a suggestion.

packages/editor/src/client/login.tsx
DamianNode: It's good suggestion 👍 I try it before I implement this using ref but doesn't work 😄 now it's good 🤣 I change this

DamianNode: Sorry for so many commits but I forget when i change to onKeyDown to leave refs, because they are needed to checking focus.

tomaszdurka: Hey @treflxyz,

Thanks for the contribution to wepublish. I took over Github project maintenance from Nico so I'll be responsible for accepting external changes. I agree with Nico that we should make it as simple as possible with the simple `onKeyDown` per element. 

Or can you explain why you reverted back to refs?
Why you need to check focus? `onKeyDown` triggers event only when actually pressed on that certain element.

tomaszdurka: Hey @treflxyz,

Thanks for the contribution to wepublish. I took over Github project maintenance from Nico so I'll be responsible for accepting external changes. I agree with Nico that we should make it as simple as possible with the simple `onKeyDown` per element. 

Or can you explain why you reverted back to refs?
Why you need to check focus? `onKeyDown` triggers event only when actually pressed on that certain element.

tomaszdurka:
@treflxyz let me know if you are willing to adjust PR to the comments. Otherwise I will close it for now and possibly later finish up.

Thanks, Tomasz

PR #484, opened 2022-01-13
export csv for users with subscriptions

bishoymelek:
Jira ticket [here](https://wepublish.atlassian.net/browse/WPC-612)

- Editor: https://editor.fwpc612cs.wepublish.dev
- Private API: https://api.fwpc612cs.wepublish.dev/admin

you can copy/paste the result [here](https://www.convertcsv.com/csv-viewer-editor.htm) to view it easily


packages/api/src/graphql/query.private.ts
```diff
@@ -290,6 +293,57 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+        const users = await dbAdapter.user.getUsers({
+          cursor: InputCursor(),
+          filter: {subscription: {}},
+          limit: Limit(1000),
+          sort: UserSort.ModifiedAt,
+          order: SortOrder.Descending
+        })
+
+        if (users?.nodes.length) {
+          const list: UserWithSubscription[] = users.nodes.map(
```
roosnic1, line 313: Here it looks like your doing the work twice. First you map through the users and then pass the array to the `parse` function which i assume will do another loop through all the objects. 

I suggest you get rid of the `parse` function and remove the dependency from the package.json. Instead of `map` i suggest you use `reduce` and for every users you add a new CSV line. Something like:
```
users.reduce((previousValue, currentValue) => {
  previousValue = `${previousValue}\n${currentValue.id},${currentValue.createdAt},....`
  return previousValue
}, ""}
```
Also you don't need the new type `UserWithSubscription`

packages/api/src/graphql/query.private.ts
```diff
@@ -290,6 +293,57 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
       }
     },
 
+    userAndSubscriptionBulkData: {
```
roosnic1, line 296: I suggest a more descriptive name like `subscriptionAsCsv`

packages/editor/src/client/api/user.graphql
```diff
@@ -1,106 +1,125 @@
 fragment FullUserSubscription on UserSubscription {
-    memberPlan {
```
roosnic1, line 2: Are these changes from prettier or your code editor?

packages/editor/src/client/app.tsx
```diff
@@ -70,6 +70,7 @@ export function contentForRoute(route: Route) {
     case RouteType.UserList:
     case RouteType.UserCreate:
     case RouteType.UserEdit:
+    case RouteType.UserExport:
```
roosnic1, line 73: there is no need for a new route. You can just open a new modal in the userlist. Similar to the preview link modal in the articlesList.

packages/editor/src/client/panel/userExportPanel.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  {t('userList.panels.exportSubscription')}
+                </Button>
+              </div>
+            </FormGroup>
+          </Form>
+          <Divider>{t('userList.panels.result')}</Divider>
+          {isLoading ? (
+            <Paragraph rows={6}>
+              {/* eslint-disable-next-line i18next/no-literal-string */}
+              <Loader center content="Loading" />
```
roosnic1, line 67: translations

roosnic1 (changes requested):
Looks good but needs some improvement

packages/api/src/graphql/query.private.ts
bishoymelek: can you check again?

packages/editor/src/client/api/user.graphql
bishoymelek: @roosnic1  Code editor, is that correct?

bishoymelek: @roosnic1 code editor but i changed that to just have my addition. I am not sure what's the right? 
If it's needed, it should have been done by husky right?

packages/api/src/graphql/query.private.ts
```diff
@@ -290,6 +296,40 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+        const {roles} = authenticate()
+        authorise(CanGetUsers, roles)
+
+        const subscriptionsList: User[] = []
+
+        const getAllUsersWithSubscriptions = async ({endCursor}: any = {}) => {
+          const listResult = await dbAdapter.user.getUsers({
+            cursor: InputCursor(endCursor),
+            filter: {subscription: {}},
+            limit: Limit(1),
```
roosnic1, line 312: `Limit(1)` You only get one user in every call? Why not the max? 100?

packages/api/src/graphql/query.private.ts
```diff
@@ -290,6 +296,40 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+            sort: UserSort.ModifiedAt,
+            order: SortOrder.Descending
+          })
+
+          subscriptionsList.push(...listResult.nodes)
+
+          if (listResult.totalCount > subscriptionsList.length) {
```
roosnic1, line 319: `listResult` has the property `pageInfo` which contains the flag `hasNextPage`. I suggest you use that.

packages/editor/src/client/routes/userList.tsx
```diff
@@ -263,6 +300,38 @@ export function UserList() {
         />
       </Drawer>
 
+      <Modal show={isExportModalOpen} onHide={() => setExportModalOpen(false)}>
```
roosnic1, line 303: Please move the whole logic for exporting in its own component.

roosnic1 (changes requested):
looking much better.

packages/api/src/utility.ts
```diff
@@ -4,6 +4,81 @@ import {delegateToSchema, IDelegateToSchemaOptions, Transform, ExecutionResult}
...
+      subscription!.autoRenew +
+      ',' +
+      new Date(subscription!.startsAt).toISOString() +
+      ',' +
+      (subscription?.paidUntil || '') +
+      ',' +
+      subscription!.paymentMethodID +
+      ',' +
+      new Date(subscription!.deactivatedAt).toISOString() +
+      '\r\n'
```
tomaszdurka, line 78: For readability I would go:
```
[value, value, value].join(',') + '/r/n'
```
as with `jsonFields`

packages/editor/src/client/locales/en.json
```diff
@@ -779,6 +779,7 @@
         "modifiedAtDate": "{{modifiedAtDate, ccc PP p}}",
         "name": "Name",
         "newUser": "New User",
+        "exportSubscription": "Export",
```
tomaszdurka, line 782: Use plural for this

tomaszdurka (changes requested):
Adding some small comments on top of Nico's.

packages/api/src/graphql/query.private.ts
```diff
@@ -290,6 +296,40 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+    subscriptionAsCsv: {
+      type: GraphQLString,
+      args: {},
+      async resolve(root, args, {dbAdapter, authenticate}) {
+        const {roles} = authenticate()
+        authorise(CanGetUsers, roles)
+
+        const subscriptionsList: User[] = []
+
+        const getAllUsersWithSubscriptions = async ({endCursor}: any = {}) => {
```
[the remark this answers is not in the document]
bishoymelek: @roosnic1 please check again

packages/api/src/graphql/query.private.ts
```diff
@@ -290,6 +296,40 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+          const listResult = await dbAdapter.user.getUsers({
+            cursor: InputCursor(endCursor),
+            filter: {subscription: {}},
```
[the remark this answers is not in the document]
bishoymelek: @roosnic1 Yes It does.

packages/api/src/utility.ts
bishoymelek: @tomaszdurka  please re-check.

packages/api/src/graphql/query.private.ts
```diff
@@ -296,35 +296,35 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
       args: {},
       async resolve(root, args, {dbAdapter, authenticate}) {
         const {roles} = authenticate()
         authorise(CanGetUsers, roles)
 
-        const subscriptionsList: User[] = []
+        const allSubscriptionsList: User[] = []
 
-        const getAllUsersWithSubscriptions = async ({endCursor}: any = {}) => {
+        const getAllUsersWithSubscriptions = async (after?: string | null) => {
```
roosnic1, line 308: `after` is optional and can be `string` and `null`? I think that's one type to many.

packages/editor/src/client/api/user.graphql
```diff
@@ -72,8 +71,8 @@ query User($id: ID!) {
     }
 }
 
-query SubscriptionAsCsv {
-  subscriptionAsCsv
+query SubscriptionsAsCsv {
+  csv: subscriptionsAsCsv
```
roosnic1, line 75: why the renaming? Seems useless.

packages/editor/src/client/routes/userList.tsx
```diff
@@ -300,37 +279,7 @@ export function UserList() {
...
-          </div>
-        </Modal.Body>
-
-        <Modal.Footer>
-          <Button onClick={() => setExportModalOpen(false)} appearance="default">
-            {t('userList.panels.close')}
-          </Button>
-        </Modal.Footer>
-      </Modal>
+      <SubscriptionAsCsvModal onHide={() => setExportModalOpen(false)} isOpen={isExportModalOpen} />
```
roosnic1, line 282: I suggest you implement the `<subscriptionAsCsvModal>` component the same way as the the `<ResetUserPasswordPanel>` below.

roosnic1 (dismissed):
Looks good but i left some comments.

packages/editor/src/client/routes/userList.tsx
bishoymelek: done

packages/api/src/graphql/query.private.ts
bishoymelek: Can you explain more what you mean?

packages/editor/src/client/api/user.graphql
bishoymelek: I think it's good to have it like `subscriptionsCsvData.csv` instead, what do you think?

tomaszdurka (approved):

penk85 (approved):
:+1:

camilleory (approved):

PR #494, opened 2022-01-20
duplicate pending article/page

penk85:
WPC-616: Fixed duplicate article/page mutation to include pending.

### DIFF
--- a/packages/api/src/graphql/mutation.private.ts
+++ b/packages/api/src/graphql/mutation.private.ts
@@ -775,7 +775,7 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
         if (!article) throw new NotFound('article', id)
 
         const articleRevision = Object.assign(
-          article.published ?? article.pending ?? article.draft,
+          article.draft ?? article.pending ?? article.published,
           {
             slug: '',
             publishedAt: undefined,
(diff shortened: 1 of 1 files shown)


packages/api/src/graphql/mutation.private.ts
```diff
@@ -774,11 +774,14 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
 
         if (!article) throw new NotFound('article', id)
 
-        const articleRevision = Object.assign(article.published ?? article.draft, {
-          slug: '',
-          publishedAt: undefined,
-          updatedAt: undefined
-        })
+        const articleRevision = Object.assign(
+          article.published ?? article.pending ?? article.draft,
```
roosnic1, line 778: I actually would suggest you change the order. First check if there is a `draft` if not check for `pending` and only then take the `published`. I think this is more intuitively for the users.

roosnic1 (commented):
Left a comment what do you think?

packages/api/src/graphql/mutation.private.ts
penk85: @roosnic1 good point

roosnic1 (approved):

camilleory (approved):
👍

PR #497, opened 2022-01-26
karma media additional url

penk85:
WPC-610: Access media adapter from internal URL


packages/api-media-karma/src/index.ts
```diff
@@ -81,7 +83,7 @@ export class KarmaMediaAdapter implements MediaAdapter {
   }
 
   async deleteImage(id: string): Promise<boolean> {
-    const response = await fetch(`${this.url}${id}`, {
+    const response = await fetch(`${this.internalURL}${id}`, {
```
roosnic1, line 86: what happens here if `internalURL` is undefined?

packages/api-media-karma/src/index.ts
```diff
@@ -19,10 +19,12 @@ export class MediaServerError extends Error {
 export class KarmaMediaAdapter implements MediaAdapter {
   readonly url: URL
   readonly token: string
+  readonly internalURL?: URL
 
-  constructor(url: URL, token: string) {
+  constructor(url: URL, token: string, internalURL: URL = url) {
     this.url = url
     this.token = token
+    this.internalURL = internalURL
```
roosnic1, line 27: I suggest you set `this.internalURL` to `internalURL ?? url` and then use `this.internaURL` for `_uploadImage()` and `deleteImage`.

docker-compose.yml
```diff
@@ -42,6 +43,7 @@ services:
     image: ghcr.io/wepublish/karma-media-server:latest
     ports:
       - 4100:4100
+      - 4101:4100
```
roosnic1, line 46: why the additional port?

roosnic1 (changes requested):

packages/api-media-karma/src/index.ts
penk85: It has a default value set in the constructor.

roosnic1: but on line 36 you use `this.internalURL ?? this.url`. Both it okay but you should stay consistent.

packages/api-media-karma/src/index.ts
penk85: `internalURL` already has `url` as its default value so I am not sure that would change anything. In `_uploadImage` I needed the ` ?? this.url`, for typescript reasons, but I don't think that's even necessary because I can't see how it would be undefined.

docker-compose.yml
penk85: @tomaszdurka suggestion.

packages/api-media-karma/src/index.ts
roosnic1: Ahhh now i see. you're using `url` as the default paramter. Okay but then you can remove the `?` from line 22 and then you shouldn't need `??` on line 36.

camilleory (approved):

tomaszdurka (approved):

packages/api/__tests__/utility.ts
```diff
@@ -96,6 +96,7 @@ export async function createGraphQLTestClientWithMongoDB(): Promise<TestClient>
   const mediaAdapter: KarmaMediaAdapter = {
     url: new URL('https://fakeurl.com'),
     token: 'fake',
+    internalURL: new URL('https://internalurl.com'),
```
roosnic1, line 99: I don't think this is needed. `KarmaMediaAdapter` should work well without an internal URL.

packages/api-media-karma/__tests__/specs/index.ts
```diff
@@ -49,132 +50,153 @@ let arrayBufferPromise: Promise<ArrayBufferUpload>
...
+})
+
+describe('Karma Media Adapter with Internal URL', () => {
+  const karmaMediaAdapterInternalURL = new KarmaMediaAdapter(
+    new URL(TEST_URL),
+    TEST_TOKEN,
+    new URL(TEST_INTERNAL_URL)
+  )
+
+  test('should accept UploadImage', async () => {
```
roosnic1, line 192: I would also test that when you delete an image the internal url is used. And also test that when you get an image that the normal url is used.

roosnic1 (changes requested):
looks good. just some additional tests needed

packages/api/__tests__/utility.ts
penk85: It shows an error when I remove it.

roosnic1 (approved):
🎉

tomaszdurka (approved):

roosnic1 (approved):
👍🏾

camilleory (approved):

tomaszdurka (approved):

PR #493, opened 2022-01-20
f/wpc-611-improve-publication-behavior

camilleory:
- `publishAt` variable added to article and page publish panel 
- Renaming of date variables so they match the field names (`initialPublishDate` --> `publishedAt`, `initialUpdateDate` --> `updatedAt`)
- The `publishAt` date is equal to `publishedAt` if the user doesn't set it. If the user sets it, it will have its own value. 
- As soon as the checkbox is disabled, `publishAt` will be set back to `publishedAt`

- For notification and info messages, check this [PR](https://github.com/wepublish/wepublish/pull/498) (based on this one)

image

PR #498, opened 2022-01-26
F/wpc 611 publish panel notifications

camilleory:
- This PR completes [this one](https://github.com/wepublish/wepublish/pull/493)
- It adds a warning message when the update date is older than the publication date
- It disables the publish panel button when the update date is older than the publication date
- it adds a info message about what the  "available online from" (`publishAt`) does

Jan-27-2022 09-21-37


packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -77,7 +88,24 @@ export function DateTimePicker({
   }
   return (
     <>
-      <ControlLabel style={{display: 'block', marginTop: '5px'}}>{label}</ControlLabel>
+      {!helpInfo ? (
+        <ControlLabel style={{display: 'block', marginTop: '5px'}}>{label}</ControlLabel>
```
tomaszdurka, line 92: Would be nice to not repeat the control element and just add the Whisper.

tomaszdurka (changes requested):

packages/editor/src/client/atoms/dateTimePicker.tsx
```diff
@@ -19,6 +28,7 @@ export interface DateTimePickerProps {
 
   dateRanges?: DateTimePreset[]
   timeRanges?: DateTimePreset[]
+  helpInfo?: string
```
penk85, line 31: Can this be added external to the datePicker component? I'm not sure if it's necessary to be part of this, or would it make more sense to have the info label as part of the publish panel? Alternatively maybe the label can be set as children to allow to pass the label text with any custom icons inside. Not sure if this is possible.

packages/editor/src/client/locales/en.json
```diff
@@ -618,6 +620,7 @@
         "close": "Close",
         "confirm": "Confirm",
         "createdAt": "{{createdAt, ccc PP p}}",
+        "dateExplanationPopOver": "This message will explain precisely what the 'available online from' date does.",
```
penk85, line 623: Can we change this? :smile:

camilleory: I'm waiting for Nina's update, she'll write a non-technical sentence about this

penk85, line 623: You still need to update this translation from pages. Otherwise looks great. :+1:

camilleory: Done 👍

penk85 (approved):

roosnic1 (approved):
👍🏾

tomaszdurka (approved):

PR #500, opened 2022-01-27
generate loginURL for mails

roosnic1:
This PR removes the `userPaymentURL` from the `sendReminderForInvoices` method and adds a loginURL for the subscription reminder mails.

### DIFF
--- a/examples/api/src/index.ts
+++ b/examples/api/src/index.ts
@@ -74,7 +74,7 @@ class ExampleURLAdapter implements URLAdapter {
   }
 
   getLoginURL(token: string): string {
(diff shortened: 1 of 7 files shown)


camilleory (approved):

penk85 (approved):
:+1:

tomaszdurka (commented):
@roosnic1 I've looked through it (didn't deep dive) and have a question.

1. We have special login url, and then we pass extra invoice query parameter which is processed by login action on the website?
2. I understand the idea is to send the url for invoice payment with jwt? Isn't it common behaviour to send link and then if it requires auth you ask to login? Let me know how your pov.

PR #505, opened 2022-02-07
Save flex teaser of page and peer

penk85:
WPC-650: Flex teasers with page or peer as a teaser weren't saving. This is because in the input type it wasn't set in a teaser.

### DIFF
--- a/packages/editor/src/client/blocks/types.ts
+++ b/packages/editor/src/client/blocks/types.ts
@@ -446,14 +446,16 @@ export function unionMapForBlock(block: BlockValue): BlockInput {
 
               case TeaserType.PeerArticle:
                 return {
-                  peerArticle: {
-                    style: flexTeaser.teaser.style,
-                    imageID: flexTeaser.teaser.image?.id,
-                    preTitle: flexTeaser.teaser.preTitle || undefined,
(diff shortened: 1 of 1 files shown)


packages/editor/src/client/blocks/types.ts
```diff
@@ -467,13 +469,15 @@ export function unionMapForBlock(block: BlockValue): BlockInput {
...
                 return {
-                  page: {
-                    style: flexTeaser.teaser.style,
-                    imageID: flexTeaser.teaser.image?.id,
-                    preTitle: flexTeaser.teaser.preTitle || undefined,
-                    title: flexTeaser.teaser.title || undefined,
-                    lead: flexTeaser.teaser.lead || undefined,
-                    pageID: flexTeaser.teaser.page.id
+                  teaser: {
+                    page: {
```
roosnic1, line 473: what about the `teaser: { article: ...}}` did you already implement that?

penk85: Yes that was already in place.

camilleory (approved):

roosnic1 (approved):
👍🏾

tomaszdurka (approved):

PR #518, opened 2022-02-18
send next invoice reminder fix

roosnic1:
This PR fixes a bug in the `memberContext`. If `parseInt returns `NaN` it should take the right side value. This works with `||` but not with `??`.


camilleory (approved):

penk85 (approved):

tomaszdurka (approved):
lgtm

I think you could even 'simplify' that with `parseInt(process.env.INVOICE_REMINDER_FREQ) || 3`

PR #523, opened 2022-02-24
F/wpc 620 peered articles list

camilleory:
- Peered articles list filterable by title and by peer
- The filter parameter now accepts an `object` (not a `string` anymore), which will enable other filter options in future. This also affects the filtering of peered articles in the select teasers panel, where the filter input needed to be adapted. 
image


packages/api/src/graphql/query.private.ts
```diff
@@ -763,7 +773,8 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
         )
 
         const peerArticles = articles.flatMap<PeerArticle & {article: any}>((result, index) => {
-          const peer = peers[index]
+          const peer = peers.filter(peer => (peerFilter ? peer.name === peerFilter : peer))[index]
```
roosnic1, line 776: here you should also return `true` instead of peer.

packages/editor/src/client/route.tsx
```diff
@@ -157,6 +159,8 @@ export const PaymentMethodCreateRoute = route(
   routePath`/paymentmethod/create`
 )
 
+export const PeerArticleListRoute = route(RouteType.PeerArticleList, routePath`/peeredarticles`)
```
roosnic1, line 162: I suggest to change the path to `/peerarticles`

packages/editor/src/client/routes/peerArticleList.tsx
```diff
@@ -0,0 +1,273 @@
...
+    loading: isLoading
+  } = usePeerArticleListQuery({
+    variables: listVariables,
+    fetchPolicy: 'network-only'
+  })
+
+  // fetch all peers
+  const {data: peerListData} = usePeerListQuery({
+    fetchPolicy: 'network-only',
+    errorPolicy: 'ignore'
```
roosnic1, line 73: why?

packages/editor/src/client/routes/peerArticleList.tsx
```diff
@@ -0,0 +1,273 @@
...
+    refetch(listVariables)
+  }, [filter, page, limit, sortOrder, sortField, peerFilter])
+
+  const {Column, HeaderCell, Cell, Pagination} = Table
+
+  useEffect(() => {
+    if (peerArticleListError) {
+      Notification.error({
+        title: peerArticleListError!.message,
+        duration: 5000
```
roosnic1, line 100: error messages should not be auto closed.

packages/editor/src/client/routes/peerArticleList.tsx
```diff
@@ -0,0 +1,273 @@
...
+          <InputGroup>
+            <Input value={filter.title || ''} onChange={value => setFilter({title: value})} />
+            <InputGroup.Addon>
+              <Icon icon="search" />
+            </InputGroup.Addon>
+          </InputGroup>
+        </FlexboxGrid.Item>
+      </FlexboxGrid>
+
+      {allPeers.length >= 2 ? (
```
roosnic1, line 122: I would always display the `<selectPicker>`. It makes the UI more consistent.

roosnic1 (changes requested):
looks very good i left some suggestions

tomaszdurka (approved):
lgtm

packages/api/src/graphql/query.private.ts
```diff
@@ -763,7 +773,8 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
         )
 
         const peerArticles = articles.flatMap<PeerArticle & {article: any}>((result, index) => {
-          const peer = peers[index]
+          const peer = peers.filter(peer => (peerFilter ? peer.name === peerFilter : true))[index]
```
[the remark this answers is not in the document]
roosnic1: True. @camilleory i guess it makes sense to filter the `peers` on line 661. This way we only have to do it once.

roosnic1 (changes requested):
Looks very good. Just one last thing.

packages/api/src/graphql/query.private.ts
```diff
@@ -659,107 +659,107 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
         after = after ? JSON.parse(base64Decode(after)) : null
 
         const peers = await dbAdapter.peer.getPeers()
+        peers.filter(peer => (peerFilter ? peer.name === peerFilter : true))
```
roosnic1, line 662: This does not work. `filter()` returns a new array with the filtered elements. But you can write it all on one line.

`const peers = await dbAdapter.peer.getPeers().filter(peer => (peerFilter ? peer.name === peerFilter : true))`

roosnic1 (changes requested):

penk85 (approved):
This looks and works great

roosnic1 (approved):
👍🏾

penk85 (approved):

tomaszdurka (approved):
lgtm

PR #525, opened 2022-02-28
display action buttons

camilleory:
All action buttons are always displayed and disabled when necessary

image


roosnic1 (approved):
🎉

tomaszdurka (approved):
lgtm

PR #533, opened 2022-03-15
added env variables to .env.example files

camilleory:
env variables added to `.env.example` file


examples/api/.env.example
```diff
@@ -1,9 +1,13 @@
 HOST_URL=http://localhost:4000
 MONGO_URL=mongodb://localhost:27017/wepublish
+MONGO_LOCALE=en
 WEBSITE_URL=http://localhost:5000
+# PORT
```
tomaszdurka, line 5: If you comment something which you can uncomment, you should not add extra space. 

Make it
```
#PORT=
```

Same goes for others.

tomaszdurka: Space you usually add when you want to add a pernament comment e.g.
```
# Here is the beloved Stripe configuration
```

tomaszdurka (changes requested):

tomaszdurka (approved):

tomaszdurka (approved):

tomaszdurka:
As discussed this is good start for improving documentation.
Let's merge as is and we can improve docu and setup itself in other prs.

penk85 (approved):

PR #536, opened 2022-03-16
bug fix in memberplans

camilleory:
- removed nonNull array in graphQL so tags can be empty
-  panel edit to be able to read empty tags


penk85 (approved):
:+1:

tomaszdurka (approved):

PR #539, opened 2022-03-16
Payload too large

penk85:
WPC-609: Increase payload limit size, and send clearer error message.


camilleory (approved):

packages/api/src/server.ts
```diff
@@ -1,12 +1,12 @@
...
+import express, {Application, NextFunction, Request, Response} from 'express'
 
 import {ApolloServer} from 'apollo-server-express'
 
 import {contextFromRequest, ContextOptions} from './context'
 import {GraphQLWepublishSchema, GraphQLWepublishPublicSchema} from './graphql/schema'
 import {MAIL_WEBHOOK_PATH_PREFIX, setupMailProvider} from './mails/mailProvider'
 import {setupPaymentProvider, PAYMENT_WEBHOOK_PATH_PREFIX} from './payments/paymentProvider'
-import {capitalizeFirstLetter} from './utility'
+import {capitalizeFirstLetter, MAX_PAYLOAD_SIZE} from './utility'
```
tomaszdurka, line 9: This is now not customizable, but I am ok with that tbh.

tomaszdurka (changes requested):
Pleae let me know if the message is now presented correct way. And if we really need to intercept this response as it seems self explanatory to me.

packages/api/src/server.ts
```diff
@@ -130,9 +131,13 @@ export class WepublishServer {
...
     })
 
-    app.use((err: any, req: Express.Request, res: Express.Response, next: NextFunction) => {
+    app.use((err: any, req: Request, res: Response, next: NextFunction) => {
       logger('server').error(err)
-      return next(err)
+
+      if (err.status === 413) {
+        res.status(413).send({error: `Payload cannot exceed ${MAX_PAYLOAD_SIZE}`})
+        return next(err)
```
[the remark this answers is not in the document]
penk85: It logs to the console. The error in the editor is "Response not successful: Received status code 413" instead of the previous error about '<'.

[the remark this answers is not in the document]
penk85: @tomaszdurka done and looking forward to feedback. I left out the 413 custom message because the generic message in console contains more information. Editor error message is the same either way.

camilleory (dismissed):

tomaszdurka (changes requested):

tomaszdurka (approved):

packages/api/src/utility.ts
```diff
@@ -220,6 +220,7 @@ export function countRichtextChars(blocksCharLength: number, nodes: any) {
 }
 
 export const MAX_COMMENT_LENGTH = 1000
+export const MAX_PAYLOAD_SIZE = '1MB'
```
michael-scheurer, line 223: what is the default? Isn't 1MB very low? And about what kind of payload are we talking here?

elias-summermatter: @michael-scheurer since its the text size its not to low ;-) Or how many text documents you have that are over 1M ;-)

elias-summermatter (approved):
Its fine for me

camilleory (approved):

PR #542, opened 2022-03-21
WPC-608: As an user i want to comment without authentication

elias-summermatter:


packages/editor/src/client/routes/commentList.tsx
```diff
@@ -183,6 +183,9 @@ export function CommentList() {
     })
   }
 
+  let printUsername = ` ${currentComment?.anonymousName} (anonymous)`
+  if (currentComment?.user) printUsername = `${currentComment?.user.name}`
```
camilleory, line 187: Translation for `anonymous`

packages/api/src/error.ts
```diff
@@ -121,6 +122,24 @@ export class CommentLengthError extends ApolloError {
...
+
+export class AnonymousCommentsDisabledError extends ApolloError {
+  constructor() {
+    super(`Anonymous comments are disabled!`)
+  }
+}
+
+export class AnonymousCommentError extends ApolloError {
+  constructor() {
+    super(`You need to give an anonymousname if your not authenticated`)
```
penk85, line 139: typo here: "anonymous name", "you're".

packages/api/src/graphql/mutation.public.ts
```diff
@@ -144,18 +147,37 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
+            throw new AnonymousCommentsDisabledError()
+
+          if (!input.anonymousName) throw new AnonymousCommentError()
+          const captchaResult = await challenge.validateChallenge({
+            challengeID: input.challengeID,
+            solve: input.challengeSolution
+          })
+          if (!captchaResult.valid) throw new CommentAuthenticationError(captchaResult.msg)
+        } else {
+          input.anonymousName = ''
```
penk85, line 170: Can you explain why `anonymousName` gets cleared here?

packages/editor/src/client/panel/memberPlanEditPanel.tsx
```diff
@@ -106,6 +106,7 @@ export function MemberPlanEditPanel({id, onClose, onSave}: MemberPlanEditPanelPr
 
   useEffect(() => {
     if (data?.memberPlan) {
+      if (!data.memberPlan.tags) data.memberPlan.tags = ['']
```
penk85, line 109: Why did you add this?

packages/editor/src/client/routes/commentList.tsx
penk85: Maybe better to display as 'unregistered user' if there's a display name?

penk85: For simplicity would a ternary assignment work better here?

packages/editor/src/client/panel/memberPlanEditPanel.tsx
elias-summermatter: My IDE does weird stuff... it changed something else which made this change required have fixed this now

packages/api/src/error.ts
```diff
@@ -19,7 +19,8 @@ export enum ErrorCode {
   CommentLengthError = 'COMMENT_LENGTH_ERROR',
   PeerTokenInvalid = 'PEER_TOKEN_INVALID',
   InternalError = 'InternalError',
-  UserSubscriptionAlreadyDeactivated = 'USER_SUBSCRIPTION_ALREADY_DEACTIVATED'
+  UserSubscriptionAlreadyDeactivated = 'USER_SUBSCRIPTION_ALREADY_DEACTIVATED',
+  InvalidCaptcha = 'InvalidCaptcha'
```
tomaszdurka, line 23: Again "ChallengeFailed" ?

tomaszdurka (changes requested):

packages/api/src/context.ts
```diff
@@ -105,12 +106,14 @@ export interface Context {
...
   readonly paymentProviders: PaymentProvider[]
   readonly hooks?: Hooks
+  readonly challenge: ChallengeProvider
 
   getOauth2Clients(): Promise<OAuth2Clients[]>
 
   authenticate(): Session
   authenticateToken(): TokenSession
   authenticateUser(): UserSession
+  optionalAuthenticateUser(): UserSession | null
```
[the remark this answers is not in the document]
elias-summermatter: Why rename? Its an additional middleware.

packages/api/src/graphql/query.public.ts
```diff
@@ -426,6 +427,22 @@ export const GraphQLPublicQuery = new GraphQLObjectType<undefined, Context>({
         const updatedInvoices = await dbAdapter.invoice.getInvoicesByUserID(user.id)
         return updatedInvoices.find(invoice => invoice !== null && invoice.id === id)
       }
+    },
+
+    // Challenge
```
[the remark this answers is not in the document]
elias-summermatter: You mean all comments or just "Challenge"?

packages/api/src/challenges/challengeProvider.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+export type ChallengeValidationProps = {
+  challengeID: string
+  solve: number
+}
+
+export type ChallengeValidationReturn = {
+  result: string
+  msg: string
+  valid: boolean
```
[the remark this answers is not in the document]
elias-summermatter: msg or new message is human readable message while result is a string to react on a specific reason why it failed. We don't use this for now but it can be interesting in the future. I can remove this but i guess at some point we will be happy to have this.

michael-scheurer (dismissed):

packages/api/src/graphql/comment.ts
```diff
@@ -97,7 +97,9 @@ export const GraphQLPublicCommentInput = new GraphQLInputObjectType({
   name: 'CommentInput',
   fields: {
     parentID: {type: GraphQLID},
-
+    anonymousName: {type: GraphQLString},
+    challengeID: {type: GraphQLString},
```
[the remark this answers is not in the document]
elias-summermatter: Did this

elias-summermatter:
@tomaszdurka can you please specify what is needed that this pr is ready to merge?

tomaszdurka (changes requested):
So let's wrap it up again:
1. As discussed before this doesn't hold with the `authorType: VerifiedUser` as it's not really a user. Add here `GuestUser` value.
2. Rename `anonymousName` to `username`, keep it empty for comments with user object as you suggested.

Going academic:
Actually I think this whole structure should be within field `author` which would be an object with field `authorType: VerifiedUser | GuestUser | Team | Author` which would then discriminate other properties like `user` or username`.
But that's breaking a lot and not worth hassle.

elias-summermatter:
> So let's wrap it up again:
> 
>     1. As discussed before this doesn't hold with the `authorType: VerifiedUser` as it's not really a user. Add here `GuestUser` value.
> 
>     2. Rename `anonymousName` to `username`, keep it empty for comments with user object as you suggested.
> 
> 
> Going academic: Actually I think this whole structure should be within field `author` which would be an object with field `authorType: VerifiedUser | GuestUser | Team | Author` which would then discriminate other properties like `user` or username`. But that's breaking a lot and not worth hassle.

I changed it this way is that fine like this... I Named it guestUsername to  make it clear and get not mixed up with other usernames

tomaszdurka (approved):

michael-scheurer (approved):

PR #544, opened 2022-03-22
WIP: fix bugs in multiple subscriptions

michael-scheurer:
fixes some bugs in multiple subscriptions:
- error when calling cancelUserSubscription mutation
- creates invoice when adding a new user subscription via editor.

### DIFF
--- a/packages/api-db-mongodb/src/db/user.ts
+++ b/packages/api-db-mongodb/src/db/user.ts
@@ -14,6 +14,7 @@ import {
   OptionalUser,
   ResetUserPasswordArgs,
   SortOrder,
+  TempUser,
   UpdatePaymentProviderCustomerArgs,
   UpdateUserArgs,
   User,
@@ -28,6 +29,7 @@ import {CollectionName, DBUser} from './schema'
 import {escapeRegExp, MongoErrorCode} from '../utility'
 import {MaxResultsPerPage} from './defaults'
 import {Cursor} from './cursor'
+import * as crypto from 'crypto'
 
 export class MongoDBUserAdapter implements DBUserAdapter {
   private users: Collection<DBUser>
@@ -68,6 +70,23 @@ export class MongoDBUserAdapter implements DBUserAdapter {
 
       throw err
     }
+  }
+
+  public async createUserFromTempUser(tempUser: TempUser): Promise<OptionalUser> {
+    const newUser = await this.createUser({
+      input: {
+        email: tempUser.email,
+        name: tempUser.name,
+        address: tempUser.address,
+        preferredName: tempUser.preferredName,
+        active: true,
+        roleIDs: [],
+        properties: [],
+        emailVerifiedAt: null
+      },
+      password: crypto.randomBytes(48).toString('hex')
+    })
+    return newUser
   }
 
   async getUser(email: string): Promise<OptionalUser> {
--- a/packages/api/src/graphql/invoice.ts
+++ b/packages/api/src/graphql/invoice.ts
@@ -68,6 +68,7 @@ export const GraphQLPublicInvoice = new GraphQLObjectType<Invoice, Context>({
     description: {type: GraphQLString},
     paidAt: {type: GraphQLDateTime},
     items: {type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLInvoiceItem)))},
+    subscriptionID: {type: GraphQLNonNull(GraphQLID)},
     total: {
       type: GraphQLNonNull(GraphQLInt),
       resolve: createProxyingResolver(({items}) => {
--- a/packages/api/src/graphql/subscription.ts
+++ b/packages/api/src/graphql/subscription.ts
@@ -50,7 +50,7 @@ export const GraphQLSubscription = new GraphQLObjectType<Subscription, Context>(
     createdAt: {type: GraphQLNonNull(GraphQLDateTime)},
     modifiedAt: {type: GraphQLNonNull(GraphQLDateTime)},
     user: {
-      type: GraphQLNonNull(GraphQLUser),
+      type: GraphQLUser,
       async resolve({userID}, args, {dbAdapter}) {
         if (isTempUser(userID)) {
           const tempUser = await dbAdapter.tempUser.getTempUserByID(removePrefixTempUser(userID))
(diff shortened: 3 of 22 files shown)


packages/api/src/payments/payrexxPaymentProvider.ts
```diff
@@ -110,6 +109,10 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
       payrexxResponse.data?.[0].id,
       this.id
     )
+    // in case Payrexx throws an error
+    if (payrexxResponse.status === 'error') {
```
tomaszdurka, line 113: Could we put here a PayrexxResponse structure/type?

packages/api/src/graphql/invoice.ts
```diff
@@ -68,6 +68,7 @@ export const GraphQLPublicInvoice = new GraphQLObjectType<Invoice, Context>({
     description: {type: GraphQLString},
     paidAt: {type: GraphQLDateTime},
     items: {type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLInvoiceItem)))},
+    subscriptionID: {type: GraphQLNonNull(GraphQLID)},
```
tomaszdurka, line 71: 1. Should this be only on public invoice?
2. Shouldn't this be connected with the invoice-item?

tomaszdurka (changes requested):

packages/api/src/graphql/invoice.ts
tomaszdurka: I understood that if we have the invoice that can have multiple items, actually the items may be connected with different subscriptions.

packages/api-db-mongodb/src/db/user.ts
```diff
@@ -70,6 +72,23 @@ export class MongoDBUserAdapter implements DBUserAdapter {
...
+        email: tempUser.email,
+        name: tempUser.name,
+        address: tempUser.address,
+        preferredName: tempUser.preferredName,
+        active: true,
+        roleIDs: [],
+        properties: [],
+        emailVerifiedAt: null
+      },
+      password: crypto.randomBytes(48).toString('hex')
```
elias-summermatter, line 87: Why to hex? Shouldn't it be to .toString('base64') or .toString('utf8')? The Problem is if you use hex the complexity is reduced to 16 possibilities per char. This would not matter if we would use an modern hash function but since we use bcrypt and bcrypt cuts of after 72 bytes. Which means we loose some generated bytes ;-) Probably not security relevant.

packages/api/src/graphql/subscription.ts
```diff
@@ -50,7 +50,7 @@ export const GraphQLSubscription = new GraphQLObjectType<Subscription, Context>(
     createdAt: {type: GraphQLNonNull(GraphQLDateTime)},
     modifiedAt: {type: GraphQLNonNull(GraphQLDateTime)},
     user: {
-      type: GraphQLNonNull(GraphQLUser),
```
elias-summermatter, line 53: Why is here not Null constraint removed. Doesn't a subscription always belong to a user?

elias-summermatter (changes requested):

packages/api/src/payments/paymentProvider.ts
```diff
@@ -129,32 +131,94 @@ export abstract class BasePaymentProvider implements PaymentProvider {
...
+  ): Promise<OptionalSubscription> {
+    // create non-temp user id
+    userID = removePrefixTempUser(userID)
+    const tempUser = await dbAdapter.tempUser.getTempUserByID(userID)
+    if (!tempUser) {
+      throw new Error('User not found while updating payment with intent state.')
+    }
+
+    // creating the new user out of the existing temp_user
+    const user = await dbAdapter.user.createUserFromTempUser(tempUser)
```
tomaszdurka, line 206: Is it possible that we run this subscription activation for 2 subscriptions with the same temp user and will try to create user twice? And actually succeed as the email uniqueness validation is only in the mutation.

packages/api/src/graphql/subscription.ts
michael-scheurer: Once an user gets deleted, the subscription is null. otherwise the app fails. I opened a ticket, so that no user gets hardly deleted from the database but are gonna be flagged.

tomaszdurka (changes requested):
@michael-scheurer please let me know about the comments / questions.

packages/api-db-mongodb/src/db/user.ts
michael-scheurer: I copy pasted the snippet from Nicos code. But toString() is limited to base 16, which is hex ;)

packages/api/src/payments/paymentProvider.ts
```diff
@@ -129,32 +131,94 @@ export abstract class BasePaymentProvider implements PaymentProvider {
...
+
+    if (!user) {
+      throw new Error('User could not be created from temp user.')
+    }
+
+    if (!subscription) {
+      throw new Error('Subscription is empty within activateTempUserAndSubscription method.')
```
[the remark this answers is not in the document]
michael-scheurer: we're not updating the user but the subscription. so if we don't have any subscription id, we cannot update the subscription ;)

packages/api-db-mongodb/src/db/user.ts
elias-summermatter: .toString('base64') is surly not limited to base 16 ;-) its base 64 ;-) And .toString('utf8') even base 256 ;-)

packages/api/src/payments/paymentProvider.ts
michael-scheurer: Not possible. The mutation only allows to create one subscription. If an user would try to register an new subscription it creates a new temp user. And it's not possible to add a subscription for a temp user via the editor. We should be fine ;)

packages/api-db-mongodb/src/db/user.ts
michael-scheurer: changed it. it's buffer encoding which allows "ascii" | "utf8" | "utf-8" | "utf16le" | "ucs2" | "ucs-2" | "base64" | "latin1" | "binary" | "hex"

PR #547, opened 2022-03-28
added sandbox property to iframe

elias-summermatter:


michael-scheurer (dismissed):

packages/editor/src/client/utility.ts
```diff
@@ -207,6 +207,14 @@ export function validateURL(url: string) {
   return false
 }
 
+export function flattenArray(array: unknown[] | DOMTokenList) {
```
tomaszdurka, line 210: Not sure why we need this? This works exactly like `array.join(" ")`, let's use that in tsx.

tomaszdurka (changes requested):

packages/editor/src/client/utility.ts
elias-summermatter: Since its a DOMTokenList there is no join("") function.....

elias-summermatter:
> If you want hotfix on 2.15, we can't merge to master.

Sorry what we want is to hotfix 3.0.0-alpha.1 or we just make a normal release since we have not done one since 3 weeks

packages/editor/src/client/utility.ts
tomaszdurka: OK!

tomaszdurka (changes requested):

tomaszdurka:
If that's the case we can merge to master sure.

packages/editor/package.json
```diff
@@ -37,7 +37,7 @@
     "watch:client": "webpack-dev-server --mode development --hot --inline",
     "watch:server": "run-p watch:server:*",
     "watch:server:start": "nodemon -r dotenv/config --inspect=0.0.0.0:9230 ./dist/server/index.js",
-    "watch:server:build": "tsc -b ./src/shared ./src/server --watch",
```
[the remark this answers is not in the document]
elias-summermatter: Was just fo developing

penk85 (approved):

camilleory (approved):

tomaszdurka (approved):

PR #550, opened 2022-03-29
fix layout for iframe embed

penk85:
WPC-572: Fix sizing of iframe embeds. 
[Editor](https://editor.bwpc572sp.wepublish.dev/)
[Website](https://bwpc572sp.wepublish.dev/a/zk4jiLO6G6NFbU0T/spotify-embeds)

WhatsApp Image 2022-03-29 at 20 17 39
WhatsApp Image 2022-03-29 at 20 18 47


camilleory (approved):

examples/website/src/shared/utility.ts
```diff
@@ -125,6 +125,16 @@ export function transformCssStringToObject(styleCustom: string): object {
...
+      const index = key.indexOf('-')
+      if (index !== -1) {
+        let camelCaseKey = key.substring(index)
+        camelCaseKey = camelCaseKey.slice(1).replace(camelCaseKey[1], camelCaseKey[1].toUpperCase())
+
+        return Object.assign(previousValue, {
+          [(key.substring(0, index) + camelCaseKey).trim()]: value.trim()
+        })
+      }
+
```
tomaszdurka, line 137: This seems to cover only single dash cases - I think there might be more.

Wdyt about using something more generic for the `key`?
```
key.toLowerCase().replace(/\-(.)/gm, ($0, $1) => $1.toUpperCase())
```

tomaszdurka (changes requested):

examples/website/src/shared/utility.ts
penk85: @tomaszdurka I'm not sure I understand the second half of this, but it indeed works.

examples/website/src/shared/blocks/embedBlock.tsx
```diff
@@ -100,7 +100,7 @@ function embedForData(data: EmbedData) {
             allowFullScreen
             title={data.title}
             src={data.url}
-            width={data.width}
+            width="100%"
```
michael-scheurer, line 103: In my opinion it should overtake data.width per default and if data.width is not set a.k. undefined use 100%
Like so:
`width={data.wdith || "100%"}`

examples/website/src/shared/utility.ts
```diff
@@ -125,7 +125,12 @@ export function transformCssStringToObject(styleCustom: string): object {
   return styleRules.reduce((previousValue: object, currentValue: string) => {
     const [key, value] = currentValue.split(':')
     if (key && value) {
-      return Object.assign(previousValue, {[key.trim()]: value.trim()})
+      return Object.assign(previousValue, {
```
michael-scheurer, line 128: what function! don't get, why the heck it is necessary to write such functions. but if it works still for other iframes, it's ok for me.

packages/editor/src/client/blocks/embeds/iframe.tsx
```diff
@@ -41,7 +40,9 @@ export function IframeEmbed({url, title, width, height, styleCustom, sandbox}: I
           scrolling="no"
           frameBorder="0"
           allowFullScreen
-          sandbox={sandbox}
+          width="100%"
+          height={height}
+          sandbox={sandbox?.length !== 0 ? sandbox : undefined}
```
michael-scheurer, line 45: U can write this shorter, because 0 is a falsy value.

`sandbox={!!sandbox?.length ? sandbox : undefined}`

michael-scheurer (dismissed):

examples/website/src/shared/blocks/embedBlock.tsx
penk85: Actually you are right it doesn't make sense to override the data. There was an underlying issue here which was causing issues with the width -it was saved as number format and therefore lost the '%' sign when rendering. Iframe with 100% width got saved as 100px width. I restored the original `data.width` here and it should be working.

camilleory (dismissed):

tomaszdurka (approved):

packages/editor/src/client/api/index.ts
```diff
@@ -342,17 +342,17 @@ export type EmbedBlock = {
...
   styleCustom?: Maybe<Scalars['String']>;
   sandbox?: Maybe<Scalars['String']>;
 };
 
 export type EmbedBlockInput = {
   url?: Maybe<Scalars['String']>;
   title?: Maybe<Scalars['String']>;
-  width?: Maybe<Scalars['Int']>;
-  height?: Maybe<Scalars['Int']>;
+  width?: Maybe<Scalars['String']>;
```
elias-summermatter, line 354: @tomaszdurka shouldn't we add here a mongodb migration that changes int to string? Or isn't this a problem if you have mixed types? I don't know mongodb enough.

packages/editor/src/client/api/index.ts
```diff
@@ -646,7 +646,7 @@ export type MemberPlan = {
   slug: Scalars['String'];
   image?: Maybe<Image>;
   description?: Maybe<Scalars['RichText']>;
-  tags: Array<Scalars['String']>
+  tags?: Maybe<Array<Scalars['String']>>;
```
elias-summermatter, line 649: Is this change really necessary and when yes why?

packages/editor/src/client/blocks/embeds/iframe.tsx
```diff
@@ -4,15 +4,15 @@ import {transformCssStringToObject} from '../../utility'
...
+  width?: string
+  height?: string
   styleCustom?: string
   sandbox?: string
 }
 
 export function IframeEmbed({url, title, width, height, styleCustom, sandbox}: IframeEmbedProps) {
-  const ratio = width !== undefined && height !== undefined ? width / height : 0
-  const noRatio = !!styleCustom && ratio === 0
+  const ratio = width !== undefined && height !== undefined ? parseInt(width) / parseInt(height) : 0
```
elias-summermatter, line 14: Since with and height are strings could you not write !width and !height or is there an other action if the values are null?

packages/editor/src/client/panel/memberPlanEditPanel.tsx
```diff
@@ -108,7 +108,7 @@ export function MemberPlanEditPanel({id, onClose, onSave}: MemberPlanEditPanelPr
     if (data?.memberPlan) {
       setName(data.memberPlan.name)
       setSlug(data.memberPlan.slug)
-      setTags(data.memberPlan.tags)
+      setTags(data.memberPlan.tags ?? [])
```
elias-summermatter, line 111: Why this change?

packages/editor/src/client/utility.ts
```diff
@@ -125,7 +125,12 @@ export function transformCssStringToObject(styleCustom: string): Record<string,
   return styleRules.reduce((previousValue: Record<string, unknown>, currentValue: string) => {
     const [key, value] = currentValue.split(':')
     if (key && value) {
-      return Object.assign(previousValue, {[key.trim()]: value.trim()})
+      return Object.assign(previousValue, {
+        [key
+          .toLowerCase()
+          .replace(/-(.)/gm, ($0, $1) => $1.toUpperCase())
+          .trim()]: value.trim()
+      })
```
elias-summermatter, line 133: Please add a comment and describe what this function does. Nobody understands this.

examples/website/src/shared/utility.ts
```diff
@@ -125,7 +125,12 @@ export function transformCssStringToObject(styleCustom: string): object {
...
+        [key
+          .toLowerCase()
+          .replace(/-(.)/gm, ($0, $1) => $1.toUpperCase())
+          .trim()]: value.trim()
+      })
```
elias-summermatter, line 133: Please add a comment and describe what this function does. Nobody understands this.

elias-summermatter (changes requested):

packages/editor/src/client/api/index.ts
penk85: I don't think this needs a migration. @tomaszdurka what do you think?

packages/editor/src/client/api/index.ts
penk85: It was causing errors compiling . I think others have run into the same, but should probably be fixed somewhere else.

packages/editor/src/client/blocks/embeds/iframe.tsx
penk85: I assume you're right. Will change this.

packages/editor/src/client/utility.ts
penk85: Its to extract the html properties of the pasted embed function and format as React. The function was already here, but needed to be adjusted to handle multi-word properties. `border-radius` -> `borderRadius`

penk85: Will add comment.

examples/website/src/shared/utility.ts
penk85: Yes was necessary to handle properties from the iframe

packages/editor/src/client/panel/memberPlanEditPanel.tsx
penk85: See comment above.

tomaszdurka (dismissed):

packages/editor/src/client/api/index.ts
tomaszdurka: Ah, well it will come as Int, even though it should be String. It won't fail unless some string specific operations are made here.
I agree migration would help.

@penk85 can you write that quickly? @elias-summermatter can you help therE?

elias-summermatter: @penk85 where you able to wrtie this migration?

packages/api-db-mongodb/src/migration.ts
```diff
@@ -702,6 +709,73 @@ export const Migrations: Migration[] = [
...
+            if (block.type === BlockType.Embed) {
+              block.height = String(block.height)
+              block.width = String(block.width)
+            }
+          })
+        }
+        await articles.findOneAndReplace({_id: article._id}, article)
+      }
+
+      const pages = db.collection(CollectionName.Articles)
```
tomaszdurka, line 748: This should be `CollectionName.Pages` right?

tomaszdurka (changes requested):

packages/api-db-mongodb/src/migration.ts
penk85: Whooops yes :+1:

tomaszdurka (approved):
I see you found some more

tomaszdurka (dismissed):

Itrulia (approved):

camilleory (approved):
👍

packages/api-db-mongodb/src/migration.ts
```diff
@@ -789,6 +797,73 @@ export const Migrations: Migration[] = [
...
+  },
+  {
+    // change embed block properties width and height from number to string
+    version: 20,
+    async migrate(db) {
+      const articles = db.collection(CollectionName.Articles)
+      const migrationArticles = await articles.find().toArray()
+
+      for (const article of migrationArticles) {
+        if (article.draft) {
```
elias-summermatter, line 809: Why not using the mongo internal function  "$toString" insted of iterating over every article?

https://www.mongodb.com/docs/manual/reference/operator/aggregation/toString/

packages/api-db-mongodb/src/migration.ts
```diff
@@ -789,6 +797,73 @@ export const Migrations: Migration[] = [
...
+          article.draft.blocks.forEach((block: ArticleBlock) => {
```
elias-summermatter, line 810: Why not use, you can prevent code duplication like this
```
for (const key in article) {
    const blocks = article[key].block
   ....
}
```
Or are there articles of a state you dont want to change?

packages/api-db-mongodb/src/migration.ts
```diff
@@ -789,6 +797,73 @@ export const Migrations: Migration[] = [
...
+        if (page.draft) {
+          page.draft.blocks.forEach((block: PageBlock) => {
+            if (block.type === BlockType.Embed) {
+              block.height = String(block.height)
+              block.width = String(block.width)
+            }
+          })
+        }
+        if (page.published) {
+          page.published.blocks.forEach((block: PageBlock) => {
```
elias-summermatter, line 849: Same here

elias-summermatter (changes requested):

packages/api-db-mongodb/src/migration.ts
elias-summermatter: @tomaszdurka: @penk85  told me you told her to do it the "unsave" way. Don't you think we should use atomic operations for data manipulation that there is less risk of failing operation?

elias-summermatter: Or maybe use Transactions: https://www.digitalocean.com/community/tutorials/how-to-use-transactions-in-mongodb

penk85: @elias-summermatter transactions for migrations will be introduced in a different pull request. If that's okay, can you approve?

Itrulia: Unfortunately transactions in Mongodb require a replica set, we currently are only running a standalone server.

tomaszdurka (approved):

elias-summermatter (approved):
Since its approved by @tomaszdurka its fine for me.

PR #557, opened 2022-04-05
WPC-671: Caching and background autorefresh of remote content 

elias-summermatter:


penk85 (approved):

elias-summermatter:
@tomaszdurka please review and approve.

packages/api/src/context.ts
```diff
@@ -46,6 +46,26 @@ import {Client, Issuer} from 'openid-client'
...
+ */
+const peerCache = new NodeCache({
+  stdTTL: 1800,
+  checkperiod: 60,
+  deleteOnExpire: false,
+  useClones: false
+})
+peerCache.on('expired', async function (key: string, value: peerCacheValue) {
+  // Refresh cache only if last use of cached entry is less than 24h ago
+  if (value.queryParams.lastQueried > new Date().getTime() - 24 * 60 * 60 * 1000) {
```
michael-scheurer, line 62: putting code such as 24 * 60 ... in a meaning full variable:

const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000

michael-scheurer: I would probably wait for 2 days.

packages/api/src/context.ts
```diff
@@ -473,62 +508,107 @@ export function tokenFromRequest(req: IncomingMessage | null): string | null {
   return null
 }
 
+/**
+ * Function that generate the key for the cache
+ * @param params
+ */
+
```
michael-scheurer, line 515: delete new line

packages/api/src/context.ts
```diff
@@ -473,62 +508,107 @@ export function tokenFromRequest(req: IncomingMessage | null): string | null {
...
+      )
+      .digest('hex')
+  )
+}
+
+/**
+ * Function to that refreshes and initializes entries in the cache
+ * @param params
+ */
+
```
michael-scheurer, line 534: delete new line

packages/api/src/context.ts
```diff
@@ -473,62 +508,107 @@ export function tokenFromRequest(req: IncomingMessage | null): string | null {
...
+        `${JSON.stringify(params.hostURL)}${JSON.stringify(
+          params.variables?._v0_id
+        )}${JSON.stringify(params.query)}`
+      )
+      .digest('hex')
+  )
+}
+
+/**
+ * Function to that refreshes and initializes entries in the cache
```
michael-scheurer, line 531: typo: unnecessary "to" in the comment

packages/api/src/context.ts
```diff
@@ -473,62 +508,107 @@ export function tokenFromRequest(req: IncomingMessage | null): string | null {
...
+      headers: {'Content-Type': 'application/json', Authorization: `Bearer ${params.token}`},
+      body: JSON.stringify({
+        query: params.query,
+        variables: params.variables,
+        operationName: params.operationName
+      }),
+      signal: abortController.signal
+    })
+    const res = await fetchResult.json()
+    if (fetchResult?.status != 200) {
```
michael-scheurer, line 553: shouldn't it be !== instead of !=

packages/api/src/context.ts
```diff
@@ -473,62 +508,107 @@ export function tokenFromRequest(req: IncomingMessage | null): string | null {
...
+      queryParams: params
+    }
+    peerCache.set(params.cacheKey, cacheValue)
+    return res
+  } catch (err) {
+    if (err.type === 'aborted') {
+      err = new Error(`Connection to peer (${params.hostURL}) timed out.`)
+    }
+
+    console.error(err)
```
michael-scheurer, line 570: would use the logger() instead console log to be consistent.

michael-scheurer (commented):
Isn't there an old cache library, which could be deleted? And see all other comments. All over all: good work imo. Looking forward to have this functionality implemented.

packages/api/src/context.ts
elias-summermatter: @michael-scheurer why would you wait for 2 days? I don't think that's necessary. Since an article really would need to not be accessed for 24h. And that only in the case an other node is offline for more than 24h. This should not be very likely ;-)

packages/api/src/context.ts
elias-summermatter: I prefer with new line.

packages/api/src/context.ts
elias-summermatter: I prefer with new line.

packages/api/src/context.ts
elias-summermatter: Fixed

packages/api/src/context.ts
elias-summermatter: Fixed

packages/api/src/context.ts
elias-summermatter: Fixed

packages/api/src/context.ts
elias-summermatter: Basicaly just copy paste from prior code but i changed it

michael-scheurer (dismissed):

penk85 (dismissed):

packages/api/src/context.ts
```diff
@@ -130,6 +151,21 @@ export interface Oauth2Provider {
...
+  cacheKey: string
+  lastQueried: number
+  readonly hostURL: string
+  readonly variables: {[p: string]: any} | undefined
+  readonly query: string
+  readonly operationName: string | undefined
+  readonly token: string
+}
+
+interface peerCacheValue {
```
tomaszdurka, line 164: Rename to `PeerCacheValue`

packages/api/src/context.ts
```diff
@@ -130,6 +151,21 @@ export interface Oauth2Provider {
   readonly redirectUri: string[]
 }
 
+interface peerQueryParams {
```
tomaszdurka, line 154: We do interfaces starting with capital letter => `PeerQueryParams`

packages/api/src/context.ts
```diff
@@ -46,6 +46,27 @@ import {Client, Issuer} from 'openid-client'
...
 import {User} from './db/user'
 import {ChallengeProvider} from './challenges/challengeProvider'
+import NodeCache from 'node-cache'
+import {logger} from './server'
+
+/**
+ * Peered article cache configuration and setup
+ */
+const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000
+const peerCache = new NodeCache({
```
tomaszdurka, line 56: `peerCache` init (including setting up event listener) should not be done in definitions file directly. Please move this logic into `createFetcher` function. In general I think whole peers/fetcher/caching concept should be moved to separate file defining those.

tomaszdurka (changes requested):
Hey @elias-summermatter 

Thanks for that pull-request, this looks pretty great and done smart way.
I only have small request about naming conventions and also moving some pieces into methods (or possibly other files).

Please let me know if you have some questions.

packages/api/src/context.ts
elias-summermatter: @tomaszdurka: You cant if you do that you will loose the cache because it it reinitialized after for every requests.... Or did i get something wrong?

tomaszdurka: Took another look. 

Indeed `createFetcher` is used in `createContextFromRequest`, which is indeed done each time. Still not happy with the creation of the node cache in definition file. It should be passed as `ContextOptions` into `WepublishServer` just like the `dbAdapter` and then used in `createContextFromRequest` from function argument.

You can add a function which creates this cache and then just pass it to the server constructor in your index.

tomaszdurka (changes requested):

elias-summermatter:
@tomaszdurka but then this will be another breaking change right?

elias-summermatter:
@tomaszdurka i have changed the name as discuss. Do you can give it free for merging?

tomaszdurka:
@elias-summermatter just last check with Karl. Thanks for patience ! :)

tomaszdurka (approved):

PR #568, opened 2022-04-22
WPC-686-Wrong auto-generated Maybe-Tags

elias-summermatter:


packages/api/src/db/memberPlan.ts
```diff
@@ -39,7 +39,7 @@ export interface MemberPlan {
 export interface MemberPlanInput {
   readonly name: string
   readonly slug: string
-  readonly tags: string[]
+  readonly tags?: string[]
```
penk85, line 42: Do we need this change? Shouldn't it anyhow allow an empty array?

tomaszdurka (approved):

PR #570, opened 2022-04-28
Revert "WPC-657: As a publisher, when uploading an image, i want the …

elias-summermatter:


tomaszdurka:
@elias-summermatter I've got confused. Is this one still neeed?

PR #578, opened 2022-05-02
feat: add .nvmrc with latest 12.x.x node version

Itrulia:
`.nvmrc` allows developers to switch to the expected node version of the project automatically when entering the code base or manually by running `nvm use`.

Example of an automatic node version switch using ZSH

```zsh
autoload -U add-zsh-hook

load-nvmrc() {
    local node_version="$(nvm version)" # Current node version
    local nvmrc_path="$(nvm_find_nvmrc)" # Path to the .nvmrc file

    # Check if there exists a .nvmrc file
    if [ -n "$nvmrc_path" ]; then
    local nvmrc_node_version=$(nvm version "$(cat "${nvmrc_path}")")

    # Check if the node version in .nvmrc is installed on the computer
    if [ "$nvmrc_node_version" = "N/A" ]; then
        # Install the node version in .nvmrc on the computer and switch to that node version
        nvm install
    # Check if the current node version matches the version in .nvmrc
    elif [ "$nvmrc_node_version" != "$node_version" ]; then
        # Switch node versions
        nvm use
    fi
    # If there isn't an .nvmrc make sure to set the current node version to the default node version
    elif [ "$node_version" != "$(nvm version default)" ]; then
    echo "Reverting to nvm default version"
    nvm use default
    fi
}

add-zsh-hook chpwd load-nvmrc
load-nvmrc
```


tomaszdurka (approved):
lgtm

Compared it with 
https://github.com/wepublish/node-docker-base/blob/main/dockerfile
Maybe adjust there too. I know the produced latest tag image is not neat enough.

PR #580, opened 2022-05-03
refactor(editor): eliminate side effect when importing i18n

Itrulia:
Just a little cleanup of a side effect when importing a script. Moved the same content into an initializer and fixed a small typo in a comment 😄

### DIFF
--- a/packages/editor/src/client/index.tsx
+++ b/packages/editor/src/client/index.tsx
@@ -7,7 +7,7 @@ import {ApolloProvider, ApolloClient, ApolloLink, InMemoryCache} from '@apollo/c
 import {createUploadLink} from 'apollo-upload-client'
 import {onError} from '@apollo/client/link/error'
 
-import './i18n'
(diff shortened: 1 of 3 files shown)


packages/editor/src/client/index.tsx
```diff
@@ -129,6 +129,8 @@ const onDOMContentLoaded = async () => {
   )
 }
 
+initI18N()
```
tomaszdurka, line 132: Glad you see the inits in _definitions_ as problem. 👍

tomaszdurka (dismissed):

penk85 (dismissed):

packages/editor/src/client/index.tsx
Itrulia: Yeah I like to keep side effects to a minimum 😄

camilleory (dismissed):

penk85 (approved):

tomaszdurka (approved):

PR #581, opened 2022-05-03
refactor(api-db-mongodb/api/editor): remove ts-ignores

### LINKED ISSUE #76: Get rid of @ts-ignore in code

Itrulia:
A few fixes for ts-ignores, there should be comments to any of the changes.

For the `yarn.lock` changes:

Updating `@types/react` fixes the mismatch between `<root>/node_modules/@types/react` and `<root/packages/editor/node_modules/@types/react`
Updating `@apollo/client` fixes the missmatch between `<root>/packages/editor/node_modules/@apollo/client` and `<root>/packages/editor/node_modules/apollo-upload-client/node_modules/@apollo/client`

While upgrading @apollo/client so that all libraries/dependencies use the same version the tests started to fail.
The errors were that they tried to write a field but it did not exist in the result. This seems to be related to this: https://github.com/apollographql/apollo-client/issues/8677. I don't know if this is a bug by apollo, but it seems like it.

Another error encountered was that `value` that was given to FormControls switched from undefined to string. This should not happen according to https://reactjs.org/docs/forms.html#controlled-components and they should always be strings.

fixes #76

### DIFF
--- a/packages/api-db-mongodb/src/migration.ts
+++ b/packages/api-db-mongodb/src/migration.ts
@@ -628,8 +628,13 @@ export const Migrations: Migration[] = [
         const paymentProvidersCustomersArray: PaymentProviderCustomer[] = []
         const paymentProviderCustomers = Object.keys(user.paymentProviderCustomers)
         paymentProviderCustomers.forEach(ppc => {
-          // @ts-ignore
-          const userPPC = user.paymentProviderCustomers[ppc]
+          interface OldPaymentProviderCustomer extends PaymentProviderCustomer {
+            customerID: never
+            id: PaymentProviderCustomer['customerID']
+          }
+
+          const userPPC = user.paymentProviderCustomers[+ppc] as OldPaymentProviderCustomer
+
           paymentProvidersCustomersArray.push({
             paymentProviderID: ppc,
             customerID: userPPC.id
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -13,7 +13,6 @@ import {JobType, runJob} from './jobs'
 import pino from 'pino'
 import pinoHttp from 'pino-http'
 import TypedEmitter from 'typed-emitter'
-import {DBAdapters} from './db/adapter'
 
 let serverLogger: pino.Logger
 
@@ -43,7 +42,7 @@ export class WepublishServer {
       if (mtp.key in dbAdapter) {
         const dbAdapterKeyTyped = mtp.key as keyof typeof dbAdapter
         mtp.methods.forEach(method => {
-          const adapter = dbAdapter[dbAdapterKeyTyped] as Record<string, any>
+          const adapter = dbAdapter[dbAdapterKeyTyped] as Record<string, Function | any>
           const methodName = `${method}${capitalizeFirstLetter(mtp.key)}`
 
           if (methodName in adapter) {
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -8,10 +8,12 @@ import {MAIL_WEBHOOK_PATH_PREFIX, setupMailProvider} from './mails/mailProvider'
 import {setupPaymentProvider, PAYMENT_WEBHOOK_PATH_PREFIX} from './payments/paymentProvider'
 import {capitalizeFirstLetter, MAX_PAYLOAD_SIZE} from './utility'
 
-import {methodsToProxy} from './events'
+import {methodsToProxy, PublishableModelEvents} from './events'
 import {JobType, runJob} from './jobs'
 import pino from 'pino'
 import pinoHttp from 'pino-http'
+import TypedEmitter from 'typed-emitter'
+import {DBAdapters} from './db/adapter'
 
 let serverLogger: pino.Logger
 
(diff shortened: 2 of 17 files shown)


packages/api-db-mongodb/src/migration.ts
```diff
@@ -628,8 +628,13 @@ export const Migrations: Migration[] = [
         const paymentProvidersCustomersArray: PaymentProviderCustomer[] = []
         const paymentProviderCustomers = Object.keys(user.paymentProviderCustomers)
         paymentProviderCustomers.forEach(ppc => {
-          // @ts-ignore
-          const userPPC = user.paymentProviderCustomers[ppc]
+          interface OldPaymentProviderCustomer extends PaymentProviderCustomer {
```
Itrulia, line 631: Replicating the old interface, as `id` does not exist anymore on `PaymentProviderCustomer` but `customerID` now does.

packages/api/src/server.ts
```diff
@@ -41,34 +43,34 @@ export class WepublishServer {
       if (mtp.key in dbAdapter) {
         const dbAdapterKeyTyped = mtp.key as keyof typeof dbAdapter
         mtp.methods.forEach(method => {
+          const adapter = dbAdapter[dbAdapterKeyTyped] as Record<string, any>
```
Itrulia, line 46: Extracted `dbAdapter[dbAdapterKeyTyped]` to type it once and not have to repeat it.
As all the DB Adapters do not share the same keys, the object keys are being typed as `string`,  as all of the methods are different but not unknown, they have to be `any` it seems. They could also be typed to `Function | any`, but since `any` is broader and includes `Function`, it would simply resolve into `any` regardless.

packages/editor/package.json
```diff
@@ -51,7 +51,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "@apollo/client": "^3.3.21",
+    "@apollo/client": "^3.6.2",
```
Itrulia, line 54: Fixes inconsitencies between `apollo-upload-client` that was ahead of the editor that caused a type error.

packages/editor/src/client/base.tsx
```diff
@@ -41,12 +43,8 @@ const iconStyles = {
   lineHeight: '56px',
   textAlign: 'center' as const
 }
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-const NavItemLink = routeLink(Nav.Item)
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-const DropdownItemLink = routeLink(Dropdown.Item)
+const NavItemLink = routeLink(Nav.Item as ComponentType<LinkHOCCompatibleProps>)
```
Itrulia, line 46: Nav.Item and Dropdown.Item contains the interface `StandardProps` that declares `[key: string]: any;`, this is lost further down the line due to inheritance as seem to be an oversight of the package developer, as the interfaces seem to indicate that certain props exists and the rest gets passed down. 

As href & onClick is not being set on them `routeLink` threw an error. Explicitly setting the `LinkHOCCompatibleProps` props, fixes this issue.

packages/editor/src/client/routes/peerList.tsx
```diff
@@ -43,10 +43,9 @@ import {
 import {DescriptionList, DescriptionListItem} from '../atoms/descriptionList'
 import {NavigationBar} from '../atoms/navigationBar'
 import {PeerInfoEditPanel} from '../panel/peerProfileEditPanel'
+import {ComponentType} from 'react'
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-const ListItemLink = routeLink(List.Item)
+const ListItemLink = routeLink(List.Item as ComponentType<LinkHOCCompatibleProps>)
```
Itrulia, line 48: Same thing as above for `Nav.Item` and `Dropdown.Item`

packages/api-db-mongodb/src/migration.ts
```diff
@@ -628,8 +628,13 @@ export const Migrations: Migration[] = [
...
+            customerID: never
+            id: PaymentProviderCustomer['customerID']
+          }
+
+          const userPPC = user.paymentProviderCustomers[+ppc] as OldPaymentProviderCustomer
```
Itrulia, line 636: as `paymentProviderCustomers` is an array, `ppc` has to be a number else the typing will be `any`, `+` casts the string to a number (also in JavaScript).

packages/editor/src/client/blocks/richTextBlock/richTextBlock.tsx
```diff
@@ -169,8 +169,6 @@ export const RichTextBlock = memo(function RichTextBlock({
           setLocation(editor.selection)
         }}
         onKeyDown={e => {
```
Itrulia, line 171: Fixed by upgrading @react/types to ^17.0.14, as <root> and <root>/packages/editor had different versions but were both using the same interface.

packages/editor/src/client/atoms/typography.tsx
```diff
@@ -93,9 +93,9 @@ export function marginForTypographySpacing(spacing: TypographySpacing): string {
 }
 
 export interface TypographyVariantCSS {
```
Itrulia, line 95: Looks like updating @types/react has also upgraded "csstypes", updated typings accordingly.
We probably should add it to the package.json so that this doesn't happen anymore, what do you think?

packages/editor/src/client/base.tsx
```diff
@@ -41,12 +50,10 @@ const iconStyles = {
   lineHeight: '56px',
   textAlign: 'center' as const
 }
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-const NavItemLink = routeLink(Nav.Item)
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
-const DropdownItemLink = routeLink(Dropdown.Item)
+const NavItemLink = routeLink(Nav.Item as ComponentType<NavProps & LinkHOCCompatibleProps>)
```
Itrulia, line 53: Nav.Item and Dropdown.Item contains the interface StandardProps that declares [key: string]: any;, this is lost further down the line due to inheritance as seem to be an oversight of the package developer, as the interfaces seem to indicate that certain props exists and the rest gets passed down.

As href & onClick is not being set on them routeLink threw an error. Explicitly setting the LinkHOCCompatibleProps props, fixes this issue.

packages/api/src/server.ts
```diff
@@ -41,34 +42,34 @@ export class WepublishServer {
       if (mtp.key in dbAdapter) {
         const dbAdapterKeyTyped = mtp.key as keyof typeof dbAdapter
         mtp.methods.forEach(method => {
+          const adapter = dbAdapter[dbAdapterKeyTyped] as Record<string, Function | any>
```
Itrulia, line 45: Extracted dbAdapter[dbAdapterKeyTyped] to type it once and not have to repeat it.
As all the DB Adapters do not share the same keys, the object keys are being typed as string, as all of the methods are different but not unknown, they have to be any it seems. They could also be typed to Function | any, but since any is broader and includes Function, it would simply resolve into any regardless.

packages/editor/__tests__/specs/authorEditPanel.tsx
```diff
@@ -35,7 +35,13 @@ describe('Author Edit Panel', () => {
             author: {
               __typename: 'Author',
               id: 'fakeId2',
-              name: 'Douglas Cole'
+              name: 'Douglas Cole',
+              slug: '',
```
Itrulia, line 39: Absence of these fields threw an error after updating Apollo. Probably bug by the library itself: https://github.com/apollographql/apollo-client/issues/8677

packages/editor/src/client/panel/authorEditPanel.tsx
```diff
@@ -48,7 +48,7 @@ export interface AuthorEditPanelProps {
 export function AuthorEditPanel({id, onClose, onSave}: AuthorEditPanelProps) {
   const [name, setName] = useState('')
   const [slug, setSlug] = useState('')
-  const [jobTitle, setJobTitle] = useState<Maybe<string>>()
+  const [jobTitle, setJobTitle] = useState('')
```
Itrulia, line 51: React threw an error because a form value was changed from undefined to string (which caused it from going uncontrolled to controlled), this should not happen according to the error and https://reactjs.org/docs/forms.html#controlled-components. It should start out being a string.

The same thing happens in the `userEditPanel.tsx` below

packages/editor/src/client/atoms/typography.tsx
tomaszdurka: Hmm, indeed if we use it directly we should lock it.

packages/editor/src/client/panel/userEditPanel.tsx
```diff
@@ -67,8 +67,8 @@ function updateAddressObject(
 
 export function UserEditPanel({id, onClose, onSave}: UserEditPanelProps) {
   const [name, setName] = useState('')
-  const [firstName, setFirstName] = useState<string | undefined>()
-  const [preferredName, setPreferredName] = useState<string | undefined>()
+  const [firstName, setFirstName] = useState('')
```
tomaszdurka, line 70: Wasn't this for a reason `undefined` instead of empty `string`?

tomaszdurka (commented):
Looks, I would add csstypes.

packages/editor/src/client/panel/userEditPanel.tsx
Itrulia: `firstName` is only used in the formControl in this component. React disallows undefined to string values in form controls, or rather console.errors about it. As they go from uncontrolled to controlled.

Itrulia: This is for example what is happening on master on the edit user panel:

Screenshot 2022-05-09 at 11 12 42

camilleory (approved):
Thanks for commenting and giving documentation :)

penk85 (approved):
Nice improvement :+1:

tomaszdurka (approved):

PR #582, opened 2022-05-04
feat(editor): umlaute should also add an additional e in slugify

Itrulia:
Also changed Eszett to ss, as Haß is Hass in (Swiss) German.

fixes WPC-242


tomaszdurka (approved):

penk85 (approved):
:+1: very thorough.

camilleory (approved):
👍

PR #585, opened 2022-05-10
feat(api-db-mongodb): migrations now run inside a transaction

Itrulia:


packages/api-db-mongodb/src/adapter.ts
```diff
@@ -183,12 +183,23 @@ export class MongoDBAdapter implements DBAdapter {
...
+        await migration.migrate(db, locale)
+        await db.collection<DBMigration>(CollectionName.Migrations).insertOne({
+          version: migration.version,
+          createdAt: new Date()
+        })
+      }
+
+      await session.commitTransaction()
+    } catch (error) {
+      await session.abortTransaction()
```
tomaszdurka, line 200: In current implementation we would fail leaving app state invalid.
After the change we are silently failing, right?

PR #592, opened 2022-05-16
Improve release process (add docs, improve release command)

tomaszdurka:


config/release.sh
```diff
@@ -1,33 +1,43 @@
 #!/usr/bin/env bash
 set -e
 
+# Check if all required libs are present
+if (! which -s git jq gh lerna-changelog lerna semver); then
```
Itrulia, line 5: Does this mean lerna, lerna-changelog and semver has to be installed globally?

tomaszdurka: Mhmm, outdated, but true it would cause some problems...

Itrulia (approved):

PR #593, opened 2022-05-17
settings page

penk85:
WPC-12: Settings page


packages/api/src/db/setting.ts
```diff
@@ -0,0 +1,47 @@
+export type Setting<T = unknown> = {
+  id: string
+
+  name: string
+  value: T | null
+  settingRestriction?: SettingRestriction
+}
+
+export interface SettingInput<T> {
```
antkiewiczk, line 9: maybe we can reuse SettingInput in create and update methods? similar as in e.g. subscriptions or payments

packages/api/src/graphql/query.private.ts
```diff
@@ -1141,6 +1142,29 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+    // Setting
+    // ======
+
+    setting: {
+      type: GraphQLSetting,
+      args: {id: {type: GraphQLID}, name: {type: GraphQLString}},
+      resolve(root, {id, name}, {authenticate, loaders}) {
+        // TODO authenticate
+
+        if ((id == null && name == null) || (id != null && name != null)) {
```
antkiewiczk, line 1155: We might think about using some kind of validations tool like Joi (https://joi.dev/api/?v=17.6.0) - something to think about for the future

packages/api-db-mongodb/src/db/schema.ts
```diff
@@ -435,3 +438,11 @@ export interface DBMailLog {
   mailProviderID: string
   mailData?: string
 }
+
+export interface DBSetting<T> {
```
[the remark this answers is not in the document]
penk85: I think so, unless we decide `value` can be type `any`. It would be string, boolean, or number depending on the case.

packages/api-db-mongodb/src/db/setting.ts
```diff
@@ -0,0 +1,74 @@
...
+      },
+      {returnOriginal: false}
+    )
+    if (!value) return null
+
+    const {_id: outID, ...setting} = value
+    return {id: outID, ...setting}
+  }
+
+  async deleteSetting({id}: DeleteSettingArgs): Promise<string | null> {
```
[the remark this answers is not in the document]
penk85: That sounds like a good idea. I can do that here, but might be worth discussing if it should be changed in other parts of the code and what it would impact.

packages/api/src/db/setting.ts
penk85: For `create` not ,because I thought its best if the restrictions cannot be updated. Perhaps there's a more concise way to set this up though.

packages/api/src/db/setting.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface UpdateSettingArgs {
+  id: string
+  input: SettingInput<any>
+}
+
+export interface DeleteSettingArgs {
+  id: string
+}
+
+export interface SettingRestriction {
```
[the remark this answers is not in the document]
penk85: I had backend in mind here because I felt frontend validation shouldn't be enough for changing setting values, is this correct? I have a check in the mutation file.

packages/editor/src/client/api/setting.graphql
```diff
@@ -0,0 +1,44 @@
...
+            allowedValues
+        }
+    }
+}
+
+mutation UpdateSetting($id: String!, $input: SettingInput!) {
+  updateSetting(id: $id, input: $input) {
+    value
+  }
+}
```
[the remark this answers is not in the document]
penk85: Its for the frontend access to the graphql so I haven't had use for it. In fact I'm not sure if its necessary at all because settings would need a migration to add or remove. Maybe we can discuss if this is the ideal approach.

tomaszdurka:
@penk85 is this one close to be ready? I think it would be great to have some pairs of eyes on it.

@michael-scheurer @elias-summermatter can you check this one - it's mostly feature for operators so would be great to have your feedback here sooner than later. If you already talked it over ignore this one please. Thanks.

penk85:
@tomaszdurka from the structure, yes, it's basically all in place, and feedback would be great. I'm still going to need to implement each setting though which I expect can take a bit time.

packages/api-db-mongodb/src/db/schema.ts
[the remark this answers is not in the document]
Itrulia: If we set T = unknown we don't have to set it as any below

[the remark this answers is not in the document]
Itrulia: We could also use the SettingName and map it to a respective type, if you want to do that and need help, let me know!

packages/api/src/db/setting.ts
Itrulia: We could do `export type SeetingInput<T = unknown> = Pick<Setting<T>,'value'>`.

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1152,8 +1155,43 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+
+    updateSetting: {
+      type: GraphQLSetting,
+      args: {
+        id: {type: GraphQLNonNull(GraphQLString)},
+        input: {type: GraphQLNonNull(GraphQLSettingInput)}
+      },
+
+      async resolve(root, {id, input}, {authenticate, dbAdapter, loaders}) {
+        const fullSetting = await loaders.settingsByID.load(id)
```
Itrulia, line 1171: Alternatively could throw a `NotFound` error too

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1152,8 +1155,43 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+        const restrictions = fullSetting?.settingRestriction
+
+        const {roles} = authenticate()
+        authorise(CanUpdateSetting, roles)
+
+        if (
```
Itrulia, line 1177: Can we break this `if` statement down into smaller steps/have variables for it? It's hard to follow inline like this.

I agree with @antkiewiczk we should start using a library like `joi` or `yup`

packages/api/src/graphql/mutation.public.ts
```diff
@@ -148,20 +149,22 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
-      async resolve(_, {input}, {optionalAuthenticateUser, dbAdapter, challenge}) {
+      async resolve(_, {input}, {optionalAuthenticateUser, dbAdapter, challenge, loaders}) {
         const user = optionalAuthenticateUser()
         let authorType = CommentAuthorType.VerifiedUser
         const commentLength = countRichtextChars(0, input.text)
 
         if (commentLength > MAX_COMMENT_LENGTH) {
           throw new CommentLengthError()
         }
+        const guestCanCommentSetting = await dbAdapter.setting.getSetting(
```
Itrulia, line 160: I would redo this a tiny bit for readability:

ala:
```typescript
        const canGuestsComment = (await dbAdapter.setting.getSetting(
          SettingName.ALLOW_GUEST_COMMENTING
        ))?.value
```

The if below would then read:

```typescript
if (!canGuestsComment) throw new AnonymousCommentsDisabledError()
```

packages/editor/src/client/api/index.ts
```diff
@@ -20,6 +20,7 @@ export type Scalars = {
   Color: string;
   RichText: Node[];
   Slug: string;
+  value: any;
```
Itrulia, line 23: Something doesn't seem right here I think, this probably should be uppercase, no?
Something went wrong with generation?

packages/editor/src/client/api/index.ts
```diff
@@ -1049,6 +1051,12 @@ export type MutationRequestChangesOnCommentArgs = {
   rejectionReason: CommentRejectionReason;
 };
 
+
+export type MutationUpdateSettingArgs = {
+  id: Scalars['String'];
```
Itrulia, line 1056: This should be generated as ID not string I think?

packages/editor/src/client/api/setting.graphql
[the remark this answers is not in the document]
Itrulia: Yeah imo settings should only be deleted/added via migrations

packages/editor/src/client/locales/en.json
```diff
@@ -560,7 +560,17 @@
       "subscriptions": "Subscriptions",
       "userRoles": "User Roles",
       "users": "Users",
-      "usersAndMembers": "Users & Members"
+      "usersAndMembers": "Users & Members",
+      "settings": "Settings",
+      "settingsPanel": {
```
Itrulia, line 565: Probably should also add them to de/fr

packages/editor/src/client/panel/settingsPanel.tsx
```diff
@@ -0,0 +1,131 @@
...
+import {Button, Drawer, Form, InputNumber, Notification, Panel, toaster, Toggle} from 'rsuite'
+import {Setting, SettingName, useSettingListQuery, useUpdateSettingMutation} from '../api'
+
+export type SettingInputProps = {
+  id?: string
+  value: any
+}
+
+export function SettingsPanel() {
+  /*
```
Itrulia, line 13: Can we delete this? :)

packages/api/src/graphql/mutation.public.ts
penk85: I can't access value in the first statement because it's still a promise.

packages/editor/src/client/api/index.ts
penk85: Good catch

packages/api/src/graphql/mutation.public.ts
Itrulia: By wrapping the await in braces you can

camilleory (dismissed):
🥇 Nice job !

packages/api-db-mongodb/src/db/setting.ts
```diff
@@ -0,0 +1,76 @@
...
+    const settings = await this.settings.find({_id: {$in: ids}}).toArray()
+    const settingMap = Object.fromEntries(
+      settings.map(({_id: id, ...setting}) => [id, {id, ...setting}])
+    )
+    return ids.map(id => settingMap[id] ?? null)
+  }
+
+  async updateSettingList(args: UpdateSettingArgs[]): Promise<OptionalSetting[]> {
+    const settings: OptionalSetting[] = []
+    for (const v of args) {
```
Itrulia, line 57: could also use map:

```typescript
return Promise.all(args.map(async ({id, value: val}) => {
  const {value} = await ....
  
  return {_id: outID, ...setting}
})
```

packages/api/src/context.ts
```diff
@@ -296,8 +305,15 @@ export async function contextFromRequest(
               console.error(peer)
               return null
             }
-
-            const fetcher = createFetcher(url.resolve(peer.hostURL, 'admin'), peer.token)
+            // TODO fix types
```
Itrulia, line 308: Can we get that fixed before merging?

packages/api/src/context.ts
```diff
@@ -296,8 +305,15 @@ export async function contextFromRequest(
...
+            const peerTimeout = Number(
```
Itrulia, line 309: don't we want to do a parseInt/+ here?

Itrulia (changes requested):

packages/api-db-mongodb/src/db/setting.ts
penk85: `Promise.all` :ok_hand:

packages/api/src/context.ts
penk85: Regarding type fix. I'm not sure if there's a way to handle how the expected type behaves here. It would always be a number but I get the warning because its expecting unknown. Is there a cleaner way to handle this?

packages/api/src/context.ts
```diff
@@ -269,8 +276,10 @@ export async function contextFromRequest(
               console.error(peer)
               return null
             }
-
-            const fetcher = createFetcher(peer.hostURL, peer.token)
+            const peerTimeout = Number(
+              (await dbAdapter.setting.getSetting(SettingName.PEERING_TIMEOUT_MS))?.value
```
michael-scheurer, line 280: what happens, if the value is undefined or null? what would Numer() return? And how would the createFetcher react?

packages/api/src/db/setting.ts
```diff
@@ -0,0 +1,45 @@
+export type Setting<T = unknown> = {
+  id: string
+
```
michael-scheurer, line 3: unnecessary white line

packages/api/src/db/setting.ts
```diff
@@ -0,0 +1,45 @@
...
+  name: SettingName
+  value: T | null
+  settingRestriction?: SettingRestriction
+}
+
+export enum SettingName {
+  DEFAULT = '',
```
michael-scheurer, line 10: What is that? Empty string seems dangerous to me. When do you use it?

packages/api/src/graphql/mutation.private.ts
```diff
@@ -323,9 +327,13 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
         const user = await dbAdapter.user.getUser(email)
         if (!user) throw new NotFound('User', email)
+        const jwtExpires = (
+          await dbAdapter.setting.getSetting(SettingName.SEND_LOGIN_JWT_EXPIRES_MIN)
+        )?.value
         const token = generateJWT({
           id: user.id,
-          expiresInMinutes: parseInt(process.env.SEND_LOGIN_JWT_EXPIRES_MIN as string)
+          expiresInMinutes:
+            Number(jwtExpires) ?? parseInt(process.env.SEND_LOGIN_JWT_EXPIRES_MIN as string)
```
michael-scheurer, line 336: why do you make it backwards compatible here but not the others?

packages/api/src/graphql/mutation.private.ts
```diff
@@ -352,11 +360,15 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+        const jwtExpires = (
+          await dbAdapter.setting.getSetting(SettingName.SEND_LOGIN_JWT_EXPIRES_MIN)
+        )?.value
         const user = await dbAdapter.user.getUser(email)
         if (!user) throw new NotFound('User', email)
         const token = generateJWT({
           id: user.id,
-          expiresInMinutes: parseInt(process.env.SEND_LOGIN_JWT_EXPIRES_MIN as string)
+          expiresInMinutes:
+            Number(jwtExpires) ?? parseInt(process.env.SEND_LOGIN_JWT_EXPIRES_MIN as string)
```
michael-scheurer, line 371: idem

packages/editor/src/client/locales/de.json
```diff
@@ -889,6 +890,28 @@
         "date": "Deaktivierung"
       }
     },
+    "errorMessages": {
+      "noFirstNameErrorMessage": "Please enter a firstname",
+      "noNameErrorMessage": "Please enter a name",
```
michael-scheurer, line 895: Bitte Namen eingeben

packages/editor/src/client/locales/de.json
```diff
@@ -889,6 +890,28 @@
...
+      "noEmailErrorMessage": "Please enter an email address",
+      "invalidEmailErrorMessage": "Please enter a valid email address",
```
michael-scheurer, line 897: Bitte eine gültige E-Mail eingeben

packages/editor/src/client/locales/de.json
```diff
@@ -889,6 +890,28 @@
...
+      "noUrlErrorMessage": "Please enter a url",
+      "invalidUrlErrorMessage": "Please enter a valid url",
+      "noImageErrorMessage": "Please enter an image",
+      "noCallToActionTextErrorMessage": "Please enter a call to action text",
+      "noCallToActionImageErrorMessage": "Please enter a call to action image",
+      "noAmountErrorMessage": "Default amount will be set to 5.00 CHF if empty",
```
michael-scheurer, line 903: Der Betrag ist 5 CHF, wenn Du nichts eingibst.

packages/editor/src/client/locales/de.json
```diff
@@ -889,6 +890,28 @@
...
+      "noPasswordErrorMessage": "Please enter a password",
+      "passwordTooShortErrorMessage": "Password must be at least 8 characters long",
+      "noPaymentProviderErrorMessage": "Please enter a payment provider",
+      "noMemberPlanErrorMessage": "Please enter a member plan",
+      "noPaymentPeriodicityErrorMessage": "Please enter a payment periodicity",
+      "noPaymentMethodErrorMessage": "Please enter a paymentMethod",
+      "noUserErrorMessage": "Please select a user",
+      "noTokenErrorMessage": "Please enter a token",
+      "invalidRange": "ungültiger Wert. Bitte Zahl zwischen {{min}} und {{max}} eingeben.",
```
michael-scheurer, line 912: Ungültiger Wert. Bitte Zahl zwischen {{min}} und {{max}} eingeben.

packages/api/src/context.ts
Itrulia: If it's already coming as number you should type it as one (value as number). parseInt, + and Number all have runtime changes.

using Number() has the side effect that strings that look like other bases, will not be in base 10. So parseInt/+ is preferred.

packages/api/src/context.ts
penk85: I can use a fallback to the env variable, but that's not guaranteed to be defined either. Would it be better to throw an error or to set a default value?

packages/api/src/context.ts
penk85: And helpful to know about parseInt vs Number() :+1:

Itrulia (approved):

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1144,8 +1157,51 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+            throw new NotFound('setting', name)
+          }
+
+          const restrictions = fullSetting?.settingRestriction
+
+          const allowedValues = restrictions?.allowedValues
+            ? restrictions?.allowedValues?.includes(String(val))
+            : true
+          const inputLength =
+            restrictions?.inputLength !== undefined
```
Itrulia, line 1188: This wouldn't catch null/''/0

```suggestion
            restrictions?.inputLength
```

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1144,8 +1157,51 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+              ? restrictions.inputLength >= String(val).length
+              : true
+          const maxValue =
+            restrictions?.maxValue !== undefined ? restrictions.maxValue >= Number(val) : true
```
Itrulia, line 1192:
```suggestion
            restrictions?.maxValue ? restrictions.maxValue >= Number(val) : true
```

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1144,8 +1157,51 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+          const minValue =
+            restrictions?.minValue !== undefined ? restrictions.minValue <= Number(val) : true
```
Itrulia, line 1194:
```suggestion
            restrictions?.minValue ? restrictions.minValue <= Number(val) : true
```

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1144,8 +1156,51 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+        value.map(async ({id, value: val}: UpdateSettingArgs) => {
+          const fullSetting = await loaders.settingsByID.load(id)
+
+          if (!fullSetting) {
+            throw new NotFound('setting', id)
+          }
+
+          const restrictions = fullSetting?.settingRestriction
+
+          const allowedValues = restrictions?.allowedValues
```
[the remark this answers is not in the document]
Itrulia: I would move all validations in the API. The issue with validations in the database layer is that you handle possible incorrect data before you pass it to the database.

Types of GraphQL can have validations added to them (isTypeOf for Objects or Scalar for Primitives). This is advocated for by the Apollo Documentation.

camilleory (approved):

packages/editor/src/client/locales/de.json
```diff
@@ -889,6 +890,28 @@
         "date": "Deaktivierung"
       }
     },
+    "errorMessages": {
+      "noFirstNameErrorMessage": "Please enter a firstname",
```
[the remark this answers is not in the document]
penk85: These were just placeholders because the validation branch only had English and I wanted the keys to match.

packages/api/src/graphql/mutation.private.ts
```diff
@@ -1144,8 +1156,51 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
+
+    updateSettingList: {
+      type: GraphQLList(GraphQLSetting),
+      args: {
+        value: {type: GraphQLList(GraphQLUpdateSettingArgs)}
+      },
+
+      async resolve(root, {value}, {authenticate, dbAdapter, loaders}) {
+        const {roles} = authenticate()
+        authorise(CanUpdateSetting, roles)
```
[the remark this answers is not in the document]
penk85: Yes, tested. Will return an error if not authorized.

Itrulia (approved):

michael-scheurer (dismissed):
I pushed a fix which prevents a potential runtime error. imo u can merge.

Itrulia (approved):

michael-scheurer (approved):

PR #598, opened 2022-05-25
WPC-640 Disable peer

antkiewiczk:
- add isDisabled field to peer
- don't pull articles from disabled peers
- align example page
- don't allow to show (and pull) peer profile when peer is disabled
- align frontend
- extra: adjust jest config to actually show coverage from the test files


examples/website/src/shared/blocks/blockRenderer.tsx
```diff
@@ -163,6 +162,10 @@ function renderTeaser(key: string, article: PublishedArticle, isPeerArticle = fa
       break
 
     case TeaserType.PeerArticle:
+      if (article.peer?.isDisabled === true) {
```
Itrulia, line 165: Is there a reason this is not 

```suggestion
      if (article.peer?.isDisabled) {
```

packages/api-db-mongodb/src/db/peer.ts
```diff
@@ -79,19 +80,22 @@ export class MongoDBPeerAdapter implements DBPeerAdapter {
 
   async updatePeer(
     id: string,
-    {hostURL, name, slug, token}: UpdatePeerInput
+    {hostURL, name, slug, token, isDisabled}: UpdatePeerInput
   ): Promise<OptionalPeer> {
+    const toUpdate = {
+      modifiedAt: new Date()
+    } as any
```
Itrulia, line 87: To narrow down the typings a bit more, we could use

```suggestion
    } as Record<string, unknown>
```

packages/api/src/graphql/query.public.ts
```diff
@@ -70,12 +70,17 @@ export const GraphQLPublicQuery = new GraphQLObjectType<undefined, Context>({
...
       args: {id: {type: GraphQLID}, slug: {type: GraphQLSlug}},
       description: 'This query takes either the ID or the slug and returns the peer profile.',
-      resolve(root, {id, slug}, {loaders}) {
+      async resolve(root, {id, slug}, {loaders}) {
         if ((id == null && slug == null) || (id != null && slug != null)) {
           throw new UserInputError('You must provide either `id` or `slug`.')
         }
 
-        return id ? loaders.peer.load(id) : loaders.peerBySlug.load(slug)
+        const toReturn = id ? await loaders.peer.load(id) : await loaders.peerBySlug.load(slug)
```
Itrulia, line 78: I think it makes more sense to give that variable a name of what it is, than what it's used for. What do you think? :)

packages/api/src/graphql/query.private.ts
```diff
@@ -203,17 +203,25 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
       args: {id: {type: GraphQLNonNull(GraphQLID)}},
-      resolve(root, {id}, {authenticate, dbAdapter, loaders}) {
+      async resolve(root, {id}, {authenticate, dbAdapter, loaders}) {
         const {roles} = authenticate()
         authorise(CanGetPeer, roles)
-        return loaders.peer.load(id)
+
+        const peer = await loaders.peer.load(id)
+
+        if (peer?.isDisabled === true) {
```
Itrulia, line 220: What's the reason we want `true` here and not just truthy?

```suggestion
        if (peer?.isDisabled) {
```

packages/api/src/graphql/query.private.ts
```diff
@@ -722,9 +728,9 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
 
         after = after ? JSON.parse(base64Decode(after)) : null
 
-        const peers = (await dbAdapter.peer.getPeers()).filter(peer =>
-          peerFilter ? peer.name === peerFilter : true
-        )
+        const peers = (await dbAdapter.peer.getPeers())
+          .filter(peer => (peerFilter ? peer.name === peerFilter : true))
+          .filter(peer => peer.isDisabled !== true)
```
Itrulia, line 733: Same thing here but with `false` and falsy

```suggestion
          .filter(peer => !peer.isDisabled)
```

packages/api/src/graphql/query.public.ts
```diff
@@ -70,12 +70,17 @@ export const GraphQLPublicQuery = new GraphQLObjectType<undefined, Context>({
...
-      resolve(root, {id, slug}, {loaders}) {
+      async resolve(root, {id, slug}, {loaders}) {
         if ((id == null && slug == null) || (id != null && slug != null)) {
           throw new UserInputError('You must provide either `id` or `slug`.')
         }
 
-        return id ? loaders.peer.load(id) : loaders.peerBySlug.load(slug)
+        const peer = id ? await loaders.peer.load(id) : await loaders.peerBySlug.load(slug)
+
+        if (peer?.isDisabled === true) {
```
Itrulia, line 80: What's the reason we want true here and not just truthy?

```suggestion
        if (peer?.isDisabled) {
```

Itrulia (dismissed):

Itrulia (approved):

packages/api/src/db/peer.ts
```diff
@@ -43,7 +44,7 @@ export interface CreatePeerInput extends BasePeer {
   token: string
```
tomaszdurka, line 44: `isDisabled` is now required for `CreatePeerInput`, is that intended?

packages/api/src/graphql/peer.ts
```diff
@@ -91,6 +93,7 @@ export const GraphQLPeer = new GraphQLObjectType<Peer, Context>({
 
     name: {type: GraphQLNonNull(GraphQLString)},
     slug: {type: GraphQLNonNull(GraphQLString)},
+    isDisabled: {type: GraphQLNonNull(GraphQLBoolean)},
```
tomaszdurka, line 96: As above.

tomaszdurka (approved):
Small comments on create api.

PR #600, opened 2022-05-26
bug fix do not return canceled invoices in renewSubscriptionForUser

roosnic1:
This PR fixes a critical bug in the subscription renewal. If the user updates the subscription the api will check if it needs to generate new periods and/or a new invoice. If there are subscription periods that have an ending date in the future the the api didn't check if the corresponding invoice was canceled. This fix checks for canceled invoices and if true will create a new period and a new invoice.

@michael-scheurer and @elias-summermatter not sure if this applies to Bajour and Hauptstadt but i think it's a critical bug.

### DIFF
--- a/packages/api/src/memberContext.ts
+++ b/packages/api/src/memberContext.ts
@@ -242,7 +242,12 @@ export class MemberContext implements MemberContext {
           (paidUntil !== null && periods[periods.length - 1].endsAt > paidUntil))
       ) {
         const period = periods[periods.length - 1]
-        return await this.loaders.invoicesByID.load(period.id)
+        const invoice = await this.dbAdapter.invoice.getInvoiceByID(period.invoiceID)
+        // only return the invoice if it hasn't been canceled. Otherwise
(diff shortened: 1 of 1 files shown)


packages/api/src/memberContext.ts
```diff
@@ -242,7 +242,12 @@ export class MemberContext implements MemberContext {
           (paidUntil !== null && periods[periods.length - 1].endsAt > paidUntil))
       ) {
         const period = periods[periods.length - 1]
-        return await this.loaders.invoicesByID.load(period.id)
+        const invoice = await this.dbAdapter.invoice.getInvoiceByID(period.invoiceID)
```
Itrulia, line 245: What's the reason we avoid the cache now?

roosnic1: no particular reason. But i'm pretty sure in this context we don't gain any speed if we use the cache.

Itrulia (approved):

PR #602, opened 2022-05-31
WPC-728 tmp user bug

michael-scheurer:


roosnic1 (commented):
@tomaszdurka looks good to me. But why do you change an existing migration? AFAIK this migration has already run on all operators. I think it makes more sense to put those changes in a new migration so it runs for everybody.

elias-summermatter:
@Itrulia No its not this pull request is here to migrate all tempuser securly

yarn.lock
```diff
@@ -5554,50 +5554,6 @@
...
-  version "2.15.0"
-  resolved "https://registry.npmjs.org/@wepublish/api-db-mongodb/-/api-db-mongodb-2.15.0.tgz#441f6bc6be6f27c70ba42555a5ed0cb8b5de92c1"
-  integrity sha512-FzcpRfu6QTi1HkpHF7/km71fxAdVs8FazDo6m3XFXPsCUMRfHM/YxSKtUAZQI7EafHMczPxHWn8J9C5yFj8h5g==
-  dependencies:
-    bcrypt "^5.0.0"
-    mongodb "^3.5.4"
-    nanoid "^2.1.11"
-
-"@wepublish/api-media-karma@*":
-  version "2.15.0"
```
elias-summermatter, line 5567: Why do you remove all this packages?

elias-summermatter (changes requested):

tomaszdurka:
@michael-scheurer any update here for @elias-summermatter comment?

yarn.lock
michael-scheurer: thanks for the review. a mistake it was.

tomaszdurka (approved):

PR #606, opened 2022-05-31
filter consistency

camilleory:


packages/editor/src/client/panel/teaserSelectPanel.tsx
```diff
@@ -21,13 +21,12 @@ export interface TeaserSelectPanelProps {
 export function TeaserSelectPanel({onClose, onSelect}: TeaserSelectPanelProps) {
   const [type, setType] = useState<TeaserType>(TeaserType.Article)
 
-  const [peerFilter, setPeerFilter] = useState<ArticleFilter>({title: ''})
+  // const [peerFilter, setPeerFilter] = useState<ArticleFilter>({title: ''})
```
Itrulia, line 24: I know you haven't opened this up for review yet but I wanted to let you know that the peerFilter works a bit different and the filter in the peerArticles is not actually used (or even defined in the args).

peerFilter checks for the name of the peer:

```
.filter(peer => (peerFilter ? peer.name === peerFilter : true))
```

tomaszdurka:
@camilleory please let me know what is the status of this one, do you need something here?

camilleory:
Yes @tomaszdurka it's for this branch I've been asking to check the filter behaviour

packages/api/__tests__/utility.ts
```diff
@@ -33,7 +34,7 @@ class ExampleURLAdapter implements URLAdapter {
     return `https://demo.wepublish.ch/page/${page.id}/${page.slug}`
   }
 
-  getPeeredArticleURL(peer: Peer, article: PublicArticle): string {
+  getPeeredArticleURL(peer: Peer, article: Article): string {
```
Itrulia, line 37: Shouldn't this be PublicArticle? Since the website only gets PublicArticle and not Article?

Itrulia (commented):
Looks good, I have 2 question 1 was regarding the Article/PeerArticle thing.
The other is I think Camille wanted to add pagination to the peers too, but we can add this in a later PR since it's not in the acceptance criteria.

Itrulia (approved):

michael-scheurer (changes requested):
I think, your changes are looking good. But I'm not sure, if the PR implements what was requested from the author of the original story. So for example I can't see or find any filter in the UI like mentioned in the story:

title: String
draft: Boolean
published: Boolean
pending: Boolean
authors: [ID!]
tags: [String!]

@camilleory Would you mind to have a look on it?

michael-scheurer (approved):

penk85 (approved):

PR #613, opened 2022-06-02
test(editor): replace enzyme with @testing-library/react to fix tests

Itrulia:
As enzyme has been deprecated for a while I replaced it with @testing-library/react (formerly react-testing-library).
This fixes the issue of flaky tests it seems, and I could remove the 100ms timeout between every test.


packages/editor/package.json
```diff
@@ -128,6 +125,7 @@
     "nodemon": "^2.0.1",
     "npm-run-all": "^4.1.5",
     "shx": "^0.3.2",
+    "snapshot-diff": "^0.9.0",
```
Itrulia, line 128: Allows to only snapshot the difference. Reduces the size of the snapshots quite a bit

antkiewiczk (approved):

PR #615, opened 2022-06-03
DRAFT Fix tests and types

antkiewiczk:


packages/api-media-karma/jest.config.js
```diff
@@ -5,15 +5,17 @@ module.exports = {
   },
   globals: {
     'ts-jest': {
-      tsConfig: 'src/tsconfig.json'
+      tsconfig: 'src/tsconfig.json'
```
Itrulia, line 8: Should be tsConfig, no? https://huafu.github.io/ts-jest/user/config/tsConfig

Itrulia (changes requested):

packages/api-media-karma/jest.config.js
antkiewiczk: I remember that after updating packages it shouted that `tsConfig` is deprecated and I should use `tsconfig` instead

tomaszdurka:
@antkiewiczk what is the status of this one? :)

packages/api/package.json
```diff
@@ -53,7 +55,7 @@
     "jsonwebtoken": "^8.5.1",
     "mandrill-api": "^1.0.45",
     "node-cache": "^5.1.2",
-    "node-fetch": "^2.6.0",
+    "node-fetch": "^3.2.4",
```
Itrulia, line 58: node-fetch 3.x fails tests in jest, probably due to node version. Can we maybe keep this to the same version as before?

antkiewiczk: yesterday I merged master and did a little bit of more fixing/cleanup but didn't want to spend much more time on that

PR #622, opened 2022-06-10
Migrate delete database mutations from api-db-mongodb to prisma

Itrulia:


Itrulia:
Tests fail because of lack of replica sets

camilleory (approved):

PR #623, opened 2022-06-14
Migrate create database mutations from api-db-mongodb to prisma

Itrulia:
BREAKING CHANGE: Prisma is now an option on the WepublishServer, mongoUrl has been removed
BREAKING CHANGE: sessionTTL is now an option on the WepublishServer
BREAKING CHANGE: hashCostFactor is now an option on the WepublishServer


Itrulia:
Tests fail because of replica sets, will be fixed with SQL PR

examples/api/src/index.ts
```diff
@@ -107,37 +107,60 @@ async function asyncMain() {
       : undefined
   )
 
+  const prisma = new PrismaClient({
```
Itrulia, line 110: Prisma is now passed down instead of MongoURL to allow for seeding in old initialize function

PR #625, opened 2022-06-20
b/WPC-765-auto-renew

michael-scheurer:


michael-scheurer:
> @michael-scheurer this doesn't really work because the limit of a 100 documents per call is hardcoded in the package `api-db-mongodb`
> 
> See here: https://github.com/wepublish/wepublish/blob/master/packages/api-db-mongodb/src/db/defaults.ts#L3

@roosnic1 We never request more than 100 items per page. See the conditions packages/api/src/memberContext.ts on line 326
So that should work. And also I tested it.

packages/api/src/memberContext.ts
```diff
@@ -312,12 +312,18 @@ export class MemberContext implements MemberContext {
     const lookAheadDate = new Date(startDate.getTime() + daysToLookAhead * ONE_DAY_IN_MILLISECONDS)
 
     const subscriptionsPaidUntil: Subscription[] = []
+    // max batches is a security feature, which prevents in case of an auto-renew bug too many people are going to be charged unintentionally.
+    const maxSubscriptionBatch = parseInt(process.env.MAX_AUTO_RENEW_SUBSCRIPTION_BATCH || 'false')
+    // max batch size is 100 given by https://github.com/wepublish/wepublish/blob/master/packages/api-db-mongodb/src/db/defaults.ts#L3
+    const batchSize =
+      !isNaN(maxSubscriptionBatch) && maxSubscriptionBatch < 100 ? maxSubscriptionBatch : 100
```
Itrulia, line 319: could use `Math.min(maxSubscriptionBatch, 100)` here, maybe cleaner?

examples/api/.env.example
```diff
@@ -42,3 +42,5 @@ DEFAULT_FROM_ADDRESS=dev@wepublish.ch
 DEFAULT_REPLY_TO_ADDRESS=reply-to@wepublish.ch
 
 ENABLE_ANONYMOUS_COMMENTS=false
+
+MAX_AUTO_RENEW_SUBSCRIPTION_BATCH=100
```
elias-summermatter, line 46: @michael-scheurer as example make "false"

packages/api/src/memberContext.ts
elias-summermatter: @michael-scheurer true but probably not as efficient ;-)

elias-summermatter (changes requested):

packages/api/src/memberContext.ts
Itrulia: Performance wise? They are equal :D

Itrulia (approved):

packages/api/src/memberContext.ts
michael-scheurer: what happens if maxSubscriptionBatch is NaN inside Math.min()?

Itrulia: Returns NaN, but it's more for the manual checking afterwards.

Im used to write/read Math.min for this so I am a bit biased as its easier to read for me.

Itrulia (dismissed):

Itrulia (approved):

PR #628, opened 2022-06-23
Normalized and validated mail on create and update

elias-summermatter:


packages/api-db-mongodb/src/db/user.ts
```diff
@@ -41,12 +43,23 @@ export class MongoDBUserAdapter implements DBUserAdapter {
   }
 
   async createUser({input, password}: CreateUserArgs): Promise<OptionalUser> {
+    const email = normalizeEmail(input.email)
+    const validInput = new Validator(input, {
```
Itrulia, line 47: Do we maybe want to use a more powerful and more developer friendly validator library like Joi? 
The library is very basic and is using magic strings and since we probably want to start validating more and more I think we should start out with a better foundation. What are your thoughts on that?

packages/api-db-mongodb/package.json
```diff
@@ -35,7 +35,9 @@
   "dependencies": {
     "bcrypt": "^5.0.0",
     "mongodb": "^3.5.4",
-    "nanoid": "^2.1.11"
+    "nanoid": "^2.1.11",
+    "node-input-validator": "^4.5.0",
+    "normalize-email": "^1.1.1"
```
Itrulia, line 40: This library hasn't been updated in years.  If the user explicitly sets a different email, we probably should let them.

I for example use email+company/type to keep track (using filters, labels etc) of emails.

Itrulia (changes requested):
Questioning the libraries used and if we really want to normalize emails instead of just doing toLowercase.

Shouldnt we also do these validations and mutations on the API side and not database side? Database layer seems to be the wrong choice in my opinion.

packages/api-db-mongodb/src/migration.ts
```diff
@@ -911,6 +912,64 @@ export const Migrations: Migration[] = [
...
+        email: string
+      }
+
+      const listSanitizedUsers: List[] = []
+      for (const user of users) {
+        user.email = normalizeEmail(user.email)
+        const duplicatedMail = listSanitizedUsers.find(element => element.email === user.email)
+
+        // If already a user with normalized mail exist merge them
+        if (duplicatedMail) {
```
michael-scheurer, line 942: nice!

packages/api-db-mongodb/src/db/user.ts
antkiewiczk: I think I suggested Joi on Penina's PR, and I definitely agree on this one ;)

Itrulia:
For my statement about doing it on the API side of things, we probably should do something like:

```typescript
const validate = (email: string) => {
  if (typeof email !== "string") {
    throw new GraphQLError(`Value is not string: ${email}`);
  }

  // validation logic goes here

  return email.toLowerCase();
}

const EmailAdress = new GraphQLScalarType({
  name: 'email',
  name: "EmailAddress",
  description: "A valid email address",
  serialize: validate,
  parseValue: validate,
  parseLiteral: ast => {
    if (ast.kind !== Kind.STRING) {
      throw new GraphQLError(
        `Query error: Can only parse strings as email addresses but got a: ${ast.kind}`
      )
    }

    return validate(ast.value)
  }
})
```

packages/api-db-mongodb/src/db/user.ts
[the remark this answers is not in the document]
elias-summermatter: No, validation like if it si a valid email is done by the api in my opiinion

Itrulia: Yeah that's why I brought it up. My personal favourite is zod because of its typescript integration

packages/api-db-mongodb/package.json
```diff
@@ -35,7 +35,9 @@
   "dependencies": {
     "bcrypt": "^5.0.0",
     "mongodb": "^3.5.4",
-    "nanoid": "^2.1.11"
+    "nanoid": "^2.1.11",
+    "node-input-validator": "^4.5.0",
```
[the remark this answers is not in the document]
Itrulia: Some validation is done by Prisma yeah. But not all of it (and probably some it relies on the DB).

packages/api-db-mongodb/src/db/user.ts
elias-summermatter: > https://github.com/omar-dulaimi/prisma-zod-generator
> 
> Just saw this, maybe this would make sense to use

This one we could do later, when we have migrated to prisma :-)

elias-summermatter:
@tomaszdurka please review and approve

michael-scheurer (approved):

PR #629, opened 2022-06-27
Send pageView analytics data

antkiewiczk:


tomaszdurka:
@antkiewiczk can you please ask specific devs involved not all of them ? Thanks !

examples/website/src/shared/route/router.tsx
```diff
@@ -22,6 +29,10 @@ export function Router() {
     }
   }, [])
 
+  useEffect(() => {
+    analytics.page()
+  }, [])
```
Itrulia, line 34: should probably add current to this array, so it sends the page event on every navigation

michael-scheurer:
> So9Hy4STwR

Think, this has to be documented and the password saved into the organisations keypass.

Itrulia (approved):

Itrulia:
I approved this one, for the plugin itself:

setInterval should be cleared when another page event is thrown: https://github.com/wepublish/analytics/blob/main/index.ts#L30

setTimeout should be cleared when another page event is thrown: https://github.com/wepublish/analytics/blob/main/index.ts#L32


Track is missing in implementation which is something we need I guess

Itrulia (approved):

tomaszdurka (approved):

PR #616, opened 2022-06-08
Migrate read database access from api-db-mongodb to prisma

Itrulia:
Things that still need to be done before this goes to master (currently points to branch prisma):
- [x] ~Filter password field out of all API User routes ([currently on roadmap to support it in the schema](https://www.notion.so/Prisma-Roadmap-50766227b779464ab98899accb98295f?p=ccb00ca22090437eb3b1726a722d0ae6))~ Fixed by using select in commit c947cc16acc00a0555aaaf1b7b83fdfafe822592
- [x] ~~Figure out why `@updatedAt` does not work, [even when noted in the documentation](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat)~~ (only doesn't work for nested types)
- [x] Check if we can send unsupported GraphQL parameters for [`3f6b3ef` (#616)](https://github.com/wepublish/wepublish/pull/616/commits/3f6b3ef1d3b5eb387fcfc341f0b4cfb2f8eb1634#diff-65746f23682e2be56d750a3200b07c85e0bc93aa0e6d00aa16812b82891eb8e7R51). As when sites upgrade, some peers will not be upgraded to keep compatibility

BREAKING CHANGE: Prisma requires MongoDB Replica Sets for mutations

BREAKING CHANGE: WepublishServer requires the MongoDB connection string now

BREAKING CHANGE: ArticleList pagination rewritten, now takes a cursor & take argument instead of first/last & before/after. Cursor is not base64 encoded anymore and skip argument is now the amount of items not pages.

BREAKING CHANGE: AuthorList pagination rewritten. Now takes a cursor & take argument instead of first/last & before/after. Cursor is not base64 encoded anymore and skip argument is now the amount of items not pages.

BREAKING CHANGE: CommentList pagination rewritten. Now takes a cursor & take argument instead of first/last & before/after. Cursor is not base64 encoded anymore and skip argument is now the amount of items not pages. CommentList is also not sorted by state anymore.

BREAKING CHANGE: ImageList pagination rewritten. Now takes a cursor & take argument instead of first/last & before/after. Cursor is not base64 encoded anymore and skip argument is now the amount of items not pages.

BREAKING CHANGE: InvoiceList pagination rewritten. Now takes a cursor & take argument instead of first/last & before/after. Cursor is not base64 encoded anymore and skip argument is now the amount of items not pages.

[...]

PR #630, opened 2022-06-28
Migrate update database mutations from api-db-mongodb to prisma #616

Itrulia:
BREAKING CHANGE: WepublishServerOpts does not take a dbAdapter options anymore
BREAKING CHANGE: Removed articleModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed pageModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed subscriptionModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed userModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed peerModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed imageModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed paymentModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed paymentMethodModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed memberPlanModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed authorModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed mailLogModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed invoiceModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed navigationModelEvents, use Prisma Middleware instead
BREAKING CHANGE: Removed userRoleModelEvents, use Prisma Middleware instead


Itrulia:
Tests failing is expected

packages/api/src/db/user.ts
```diff
@@ -39,92 +19,24 @@ export interface UserFilter {
...
-
-export interface DBUserAdapter {
-  updateUser(args: UpdateUserArgs): Promise<OptionalUser>
-
-  resetUserPassword(args: ResetUserPasswordArgs): Promise<OptionalUser>
-
-  addOAuth2Account(args: UserOAuth2AccountArgs): Promise<OptionalUser>
-
-  updatePaymentProviderCustomers(args: UpdatePaymentProviderCustomerArgs): Promise<OptionalUser>
+export const unselectPassword: Record<
```
Itrulia, line 22: Hides the password from the API consumers since they do not need to know (it would be a security concern)

packages/api/src/error.ts
```diff
@@ -121,6 +121,15 @@ export class DuplicatePageSlugError extends ApolloError {
   }
 }
 
+export class DuplicateArticleSlugError extends ApolloError {
```
Itrulia, line 124: Same thing as Pages

packages/api/src/graphql/member-plan/member-plan.private-mutation.ts
```diff
@@ -29,3 +29,18 @@ export const createMemberPlan = (
...
 }
+
+export const updateMemberPlan = (
+  id: string,
+  input: Omit<Prisma.MemberPlanUncheckedUpdateInput, 'modifiedAt' | 'createdAt'>,
+  authenticate: Context['authenticate'],
+  memberPlan: PrismaClient['memberPlan']
+) => {
+  const {roles} = authenticate()
+  authorise(CanCreateMemberPlan, roles)
```
michael-scheurer, line 40: This should probably be "CanUpdateMemberPlan", if exists.

packages/api/src/graphql/mutation.private.ts
```diff
@@ -183,12 +192,8 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
...
       type: GraphQLNonNull(GraphQLPeerProfile),
       args: {input: {type: GraphQLNonNull(GraphQLPeerProfileInput)}},
-      async resolve(root, {input}, {hostURL, authenticate, dbAdapter}) {
-        const {roles} = authenticate()
-        authorise(CanUpdatePeerProfile, roles)
-
-        return {...(await dbAdapter.peer.updatePeerProfile(input)), hostURL}
-      }
+      resolve: (root, {input}, {hostURL, authenticate, prisma: {peerProfile}}) =>
+        upsertPeerProfile(input, hostURL, authenticate, peerProfile)
```
michael-scheurer, line 196: In my opinion the authentication and authorization should be handled right here. Why?
- Separation of concerns: Authentication and authorization is much closer to GraphQL than inside business logic and database requests, like it is now.
- All authorization and authentication is done in one file and in one place which is mutation.private.ts. This makes security concerns much easier, less complex and less distributed.
- We don't have to pass authentication context like upsertPeerProfile(input, hostURL, **authenticate**, peerProfile)

packages/api/src/graphql/mutation.private.ts
```diff
@@ -986,7 +869,7 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
       args: {
         value: {type: GraphQLList(GraphQLUpdateSettingArgs)}
       },
-      async resolve(root, {value}, {authenticate, dbAdapter, prisma}) {
+      async resolve(root, {value}, {authenticate, prisma}) {
```
michael-scheurer, line 872: the updateSettingList business logic should also be outsourced like the other collections.

packages/api/src/graphql/mutation.public.ts
```diff
@@ -127,55 +121,8 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
-          input: {
-            ...input,
-            userID: user?.user.id,
-            authorType,
-            state: CommentState.PendingApproval
-          }
-        })
-      }
+      resolve: (_, {input}, {optionalAuthenticateUser, prisma: {comment, setting}, challenge}) =>
+        addPublicComment(input, optionalAuthenticateUser, challenge, setting, comment)
```
michael-scheurer, line 125: so much better! I can't believe it.

yarn.lock
```diff
@@ -749,15 +749,7 @@
...
-"@babel/helper-function-name@^7.18.6":
-  version "7.18.6"
-  resolved "https://registry.yarnpkg.com/@babel/helper-function-name/-/helper-function-name-7.18.6.tgz#8334fecb0afba66e6d87a7e8c6bb7fed79926b83"
-  integrity sha512-0mWMxV1aC97dhjCah5U5Ua7668r5ZmSC2DLfH2EZnf9c3/dHZKiFa5pRLMH5tjSl471tY6496ZWk/kjNONBxhw==
-  dependencies:
-    "@babel/template" "^7.18.6"
-    "@babel/types" "^7.18.6"
-
-"@babel/helper-function-name@^7.18.9":
+"@babel/helper-function-name@^7.18.6", "@babel/helper-function-name@^7.18.9":
```
michael-scheurer, line 752: why do we use two versions of the same package?

michael-scheurer (changes requested):
First of all: A really great job you did! I very much enjoyed, that you kicked out the boilerplate dbAdapter.
You'll find a bunch of small annotations.
One of the remarks is crucial though: In my opinion the authentication and authorization should be handled right in the methods of the mutation.private.ts file. You'll find my argumentation why we should improve that in a comment in that mentioned file.

packages/api/src/events.ts
```diff
@@ -1,215 +1,132 @@
...
-  | NavigationModelEventsEmitter
-  | PageModelEventEmitter
-  | PaymentModelEventEmitter
-  | PaymentMethodModelEventEmitter
-  | PeerModelEventsEmitter
-  | SubscriptionModelEventsEmitter
-  | UserModelEventsEmitter
-  | UserRoleModelEventsEmitter
-  | SettingModelEventsEmitter
+// @TODO: move into cron job
```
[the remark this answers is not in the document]
Itrulia: https://wepublish.atlassian.net/browse/WPC-829

packages/api/src/events.ts
```diff
@@ -1,215 +1,132 @@
...
+              lte: new Date()
+            }
+          }
+        }
+      }
+    }
+  })
+
+  await Promise.all(
+    pages.map(({id, pending}) =>
```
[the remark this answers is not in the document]
Itrulia: We can yeah. Just 1:1 translated it to make minimal changes

packages/api/src/events.ts
```diff
@@ -1,215 +1,132 @@
...
+}
 
 /**
  * This event listener is used after invoice has been marked as paid. The following logic is responsible to
  * update the subscription periode, eventually create a permanent user out of the temp user and sending mails
  * to the user.
  */
-invoiceModelEvents.on('update', async (context, model) => {
+export const onInvoiceUpdate = (context: Context): Prisma.Middleware => async (params, next) => {
+  if (params.model !== 'Invoice') {
```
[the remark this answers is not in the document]
Itrulia: Yes, else it will run on all models

packages/api/src/graphql/member-plan/member-plan.private-mutation.ts
Itrulia: Only peer profile and settings currently have separate permissions for updating.

yarn.lock
Itrulia: We used to, now we use 1 version. Just multiple packages we rely on have different versions specified.

packages/api/src/graphql/comment/comment.public-mutation.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // Challenge
+  if (!user) {
+    authorType = CommentAuthorType.guestUser
+
+    const guestCanCommentSetting = await settingsClient.findUnique({
+      where: {name: SettingName.ALLOW_GUEST_COMMENTING}
+    })
+    const guestCanComment =
+      guestCanCommentSetting?.value ?? process.env.ENABLE_ANONYMOUS_COMMENTS === 'true'
```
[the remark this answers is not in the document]
Itrulia: Can you explain what bug? As we only care if it's truthy.
In the DB it's a boolean.

This is also currently like this on master.

packages/api/__tests__/specs/comment.ts
```diff
@@ -1,16 +1,12 @@
-import {MongoDBAdapter} from '@wepublish/api-db-mongodb'
-import {createGraphQLTestClientWithMongoDB} from '../utility'
 // import {CommentInput, AddComment} from '../api/public'
 // import {CommentAuthorType, CommentItemType} from '../../lib'
 
 // let testClientPrivate: ApolloServerTestClient
-let dbAdapter: MongoDBAdapter
 
 beforeAll(async () => {
   try {
```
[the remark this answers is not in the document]
Itrulia: As it was before, this is an improvement I would definitely do.
I guess the comment tests were never finished :(

packages/api/src/events.ts
```diff
@@ -1,215 +1,132 @@
...
+    next(params)
+    return
+  }
+
+  if (params.action !== 'update') {
```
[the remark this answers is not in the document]
Itrulia: Because if it isn't update we don't want this middleware to run :)

packages/api/src/graphql/mutation.private.ts
Itrulia: Hey hey, the idea of these functions is so that the mutation/query files aren't huge as they are right now. The function in itself doesn't really authenticate and imo it's a very misleading name.

```typescript
  authenticate() {
      if (!session) {
        throw new AuthenticationError('Invalid session!')
      }

      if (!isSessionValid) {
        throw new TokenExpiredError()
      }

      return session
    }
```

I agree with you on principle (but not implementation) as I think this should be abstracted away and shouldn't have to be repeated in every single method.

Is this blocking for you?

packages/api/src/graphql/mutation.private.ts
Itrulia: Agree with you, came from merge with master. Will be done in this PR

Itrulia: Done!

packages/api/src/graphql/comment/comment.public-mutation.ts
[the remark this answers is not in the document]
michael-scheurer: If it's boolean, it's fine ;)

packages/api/src/graphql/mutation.private.ts
michael-scheurer: Created a Jira story

Itrulia: Amazing, thank you!

michael-scheurer (approved):

PR #633, opened 2022-07-01
Disappearing duplicate page and article

antkiewiczk:


Itrulia:
The expected behaviour changed a little bit after we had the refinement: https://wepublish.atlassian.net/browse/WPC-737
It should open the article now. This is because when you duplicate the article, depending on your sorting you can't see it. Since most of the time you want to edit the article once you duplicate it, it should open it.

I guess this was missed

Itrulia (approved):

tomaszdurka (approved):

PR #638, opened 2022-07-11
Fixed lower mail

elias-summermatter:


Itrulia (approved):

packages/api/src/graphql/mutation.private.ts
```diff
@@ -412,9 +412,11 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
         input: {type: GraphQLNonNull(GraphQLUserInput)},
         password: {type: GraphQLNonNull(GraphQLString)}
       },
-      resolve(root, {input, password}, {authenticate, dbAdapter}) {
+      async resolve(root, {input, password}, {authenticate, dbAdapter}) {
         const {roles} = authenticate()
         authorise(CanCreateUser, roles)
+        input.email = input.email.toLowerCase()
```
michael-scheurer, line 418: Imo we should implement that in the database package to tackle it at the root. Whenever another person will implement an additional user endpoint which includes somehow the email, he or she will not know, that she or he has to implement email.toLowerCase()
This is maintainability imo.

packages/api/src/graphql/mutation.private.ts
```diff
@@ -425,9 +427,11 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
         id: {type: GraphQLNonNull(GraphQLID)},
         input: {type: GraphQLNonNull(GraphQLUserInput)}
       },
-      resolve(root, {id, input}, {authenticate, dbAdapter}) {
+      async resolve(root, {id, input}, {authenticate, dbAdapter}) {
         const {roles} = authenticate()
         authorise(CanCreateUser, roles)
+        input.email = input.email.toLowerCase()
```
michael-scheurer, line 433: Should we implement a migration to transform all existing mails?

elias-summermatter: That was in the commit before that is already merged this is just a fix of missing parameters

packages/api/src/graphql/mutation.private.ts
elias-summermatter: That was how i implemented it first and Karl told me then he thinks thats wrong so i changed this. So we leaf it for now link this ok? ;-)

michael-scheurer: ok. but want to state, that this is less maintainable.

Itrulia: We should fix it on the API, if we fix it on the database layer then everything before is not using the correct data. If  we fix it on the API, everything is using the correct data.

My suggestion was to create a ScalarType for email to validate it, that's the recommended way by Apollo too.

michael-scheurer: Ok. Good argument. Agree.

PR #639, opened 2022-07-11
F/wpc 776 allow admin to generate login token for other users

elias-summermatter:


packages/api/src/error.ts
```diff
@@ -174,3 +174,15 @@ export class UserSubscriptionAlreadyDeactivated extends ApolloError {
     super(msg, ErrorCode.UserSubscriptionAlreadyDeactivated)
   }
 }
+
+export class GivenTokeExpiryToLongError extends ApolloError {
+  constructor() {
+    super(`Given token expiry is to long!`)
```
michael-scheurer, line 180: simple quote marks are sufficient

packages/api/src/error.ts
```diff
@@ -174,3 +174,15 @@ export class UserSubscriptionAlreadyDeactivated extends ApolloError {
...
+  }
+}
+
+export class UserIdNotFound extends ApolloError {
+  constructor() {
+    super(`Given user ID not valid!`)
```
michael-scheurer, line 186: idem

packages/api/src/graphql/query.private.ts
```diff
@@ -189,6 +197,37 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+        userId: {type: GraphQLNonNull(GraphQLString)},
+        expiresInMinutes: {type: GraphQLNonNull(GraphQLInt)}
+      },
+      async resolve(
+        root,
+        {userId, expiresInMinutes},
+        {authenticate, generateJWT, dbAdapter},
+        info
+      ) {
+        const THIRTY_DAYS_IN_MIN = 30 * 24 * 60
```
michael-scheurer, line 212: Should go into settings. If not available yet, open a jira-story, so that we don't forget and can deliver it, as soon as the task is completed.

michael-scheurer (changes requested):
clean implementation overall.

packages/api/src/graphql/query.private.ts
elias-summermatter: https://wepublish.atlassian.net/browse/BAJ-548

PR #642, opened 2022-07-14
Hide components that user has no access to

penk85:
WPC-723: the editor should only display menu items that the user has permissions to access.

[editor](https://editor.fwpc723hi.wepublish.dev/)


packages/editor/src/client/atoms/permissionControl.tsx
```diff
@@ -0,0 +1,16 @@
+import React, {ReactElement} from 'react'
+import {AuthContext} from '../authContext'
+import {authorise} from '../utility'
+
+interface PermissionControlProps {
+  children: ReactElement
```
Itrulia, line 6: Can be removed

packages/editor/src/client/atoms/permissionControl.tsx
```diff
@@ -0,0 +1,16 @@
...
+  requiredPermission: string
+}
+
+export function PermissionControl({children, requiredPermission}: PermissionControlProps) {
```
Itrulia, line 10:
```suggestion
export function PermissionControl({children, requiredPermission}: PropsWithChildren<PermissionControlProps>) {
```

packages/editor/src/client/utility.ts
```diff
@@ -230,3 +230,10 @@ export function getImgMinSizeToCompress(): number {
   )
   return imgMinSizeToCompress
 }
+
+export function authorise(roles: UserRole[] | undefined, permissionID: string): boolean {
+  if (!roles) return false
+  return roles
+    ?.map(role => role.permissions.some(permission => permission.id === permissionID))
```
Itrulia, line 237:
```suggestion
    .map(role => role.permissions.some(permission => permission.id === permissionID))
```

Itrulia (changes requested):

packages/editor/src/client/utility.ts
```diff
@@ -230,3 +230,10 @@ export function getImgMinSizeToCompress(): number {
   )
   return imgMinSizeToCompress
 }
+
+export function authorise(roles: UserRole[] | undefined, permissionID: string): boolean {
+  if (!roles) return false
+  return roles
```
Itrulia, line 236: `return roles.filter(role => role.permissions.some(permission => permission.id === permissionID))`

packages/editor/src/client/routes/peerArticleList.tsx
```diff
@@ -284,3 +285,8 @@ export function PeerArticleList() {
     </>
   )
 }
+
+const CheckedPermissionPeerArticleList = createCheckedPermissionComponent(
+  'CAN_GET_PEER   _ARTICLES'
```
Itrulia, line 290:
```suggestion
  'CAN_GET_PEER_ARTICLES'
```

packages/editor/src/client/atoms/permissionControl.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function PermissionControl({
+  children,
+  requiredPermission,
+  showMessage
+}: PropsWithChildren<PermissionControlProps>) {
+  const {t} = useTranslation()
+  return (
+    <AuthContext.Consumer>
+      {value => (
+        <>
```
michael-scheurer, line 19: You could probably refactor the following lines with additional functions. For now the logic is very unreadable.

packages/editor/src/client/atoms/permissionControl.tsx
```diff
@@ -0,0 +1,44 @@
...
+          {value.session?.roles?.some(role =>
+            role.permissions.some(permission => permission.id === requiredPermission)
+          ) ? (
+            children
+          ) : showMessage ? (
+            <p>{t('permissions.noAccess')}</p>
```
michael-scheurer, line 25: Doesn't look that nice in the UI, when I test it. The sentence seams a bit lost. You may want to add an icon, align the text and choose probably a bigger font size? Maybe even add some more explanatory text?

image

packages/editor/src/client/locales/en.json
```diff
@@ -1109,6 +1109,9 @@
       "login": "Login",
       "peering": "Peering",
       "payment": "Payment"
+    },
+    "permissions": {
+      "noAccess": "You have insuffient permission to access this content"
```
michael-scheurer, line 1114: missing other languages

packages/editor/src/client/panel/imageEditPanel.tsx
```diff
@@ -394,3 +398,9 @@ export function ImagedEditPanel({id, file, onClose, onSave, imageMetaData}: Imag
     </>
   )
 }
+
+const CheckedPermissionComponent = createCheckedPermissionComponent(
```
michael-scheurer, line 402: Could you outline the purpose of this createCheckedPermissionComponent? Please add it to the docs.

packages/editor/src/client/panel/imageSelectPanel.tsx
```diff
@@ -169,3 +176,9 @@ export function ImageSelectPanel({onClose, onSelect}: ImageSelectPanelProps) {
     </>
   )
 }
+
+const CheckedPermissionComponent = createCheckedPermissionComponent(
```
michael-scheurer, line 180: Seems very repetitive, could we find a more generic solution/approach?

packages/editor/src/client/panel/memberPlanEditPanel.tsx
```diff
@@ -249,6 +255,7 @@ export function MemberPlanEditPanel({id, onClose, onSave}: MemberPlanEditPanelPr
             <Form.Group>
               <Form.ControlLabel>{t('articleEditor.panels.tags')}</Form.ControlLabel>
               <TagPicker
+                disabled={isDisabled}
```
michael-scheurer, line 258: instead of disabling each single form element, couldn't we frame the whole form with a <PermissionControl>? I'm afraid it's not sustainable nor maintainable.

michael-scheurer (changes requested):
Hey Penina.
This was quite a bit of work, I guess. 
- Please find my comments in the code.
- Also, please update the docs so that all further developer know the intention behind your work and how to implement your functionality.
- http://localhost:3000/peering isn't caught properly, when log-in with an editor user.

All in all good work!

packages/editor/src/client/atoms/permissionControl.tsx
```diff
@@ -0,0 +1,44 @@
+import React, {ComponentType, PropsWithChildren} from 'react'
+import {AuthContext} from '../authContext'
+import {useTranslation} from 'react-i18next'
+
+interface PermissionControlProps {
+  requiredPermission: string
```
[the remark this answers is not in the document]
Itrulia: They are in @wepublish/api which editor can not import as we rely on dist files and they are compiled for nodejs.

packages/editor/src/client/atoms/permissionControl.tsx
Itrulia: I agree

packages/editor/src/client/panel/imageSelectPanel.tsx
Itrulia: This is already the generic solution. Instead of having to make the check we use a HoC generator here.

Itrulia (changes requested):
I agree with @michael-scheurer that the disable is not optimal.
Imo it should be on the form and not on the input, as the whole form is disabled.

packages/editor/src/client/atoms/permissionControl.tsx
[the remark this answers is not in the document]
penk85: Meaning for a case where multiple permissions are necessary? You mean for example in a delete, to check for both delete and get? Or did you have something else in mind? 

Regarding the typing, I agree it's really not ideal, but I couldn't import for reasons that Karl mentioned above. He reported this as a separate ticket I believe. He advised to handle it as a string for now and hopefully can be refactored later with imports.

packages/editor/src/client/atoms/colorPicker.tsx
```diff
@@ -5,14 +5,16 @@ interface ColorPickerProps {
...
 }
 
-export function ColorPicker({setColor, currentColor, label}: ColorPickerProps) {
+export function ColorPicker({setColor, currentColor, label, disabled}: ColorPickerProps) {
   const id = nanoid()
   return (
     <>
       {label && <label htmlFor={id}>{label}</label>}
       <input
+        disabled={disabled ?? false}
```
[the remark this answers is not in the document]
Itrulia: Yes should be. null/undefined is falsy.

packages/editor/src/client/atoms/permissionControl.tsx
```diff
@@ -0,0 +1,44 @@
...
+          ) : showMessage ? (
+            <p>{t('permissions.noAccess')}</p>
+          ) : null}
+        </>
+      )}
+    </AuthContext.Consumer>
+  )
+}
+
+export const createCheckedPermissionComponent = (permission: string, showMessage?: boolean) => <
```
Itrulia, line 33: Do we want to have showMessage default to true? Would make sense I think.

packages/editor/src/client/panel/memberPlanEditPanel.tsx
penk85: This was the existing form setup. The disabled property was already on the form elements and I just added to the elements that didn't have it. Can we fix that in a separate ticket?

Itrulia (dismissed):

packages/editor/src/client/atoms/permissionControl.tsx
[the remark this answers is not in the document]
penk85: I had trouble importing permissions so its string for now and eventually can be refactored to be more specific when the type can be imported.

penk85: Definitely a better idea.

Itrulia (dismissed):

michael-scheurer (commented):
Just pushed two minor enhancements. You may have a look at them.

Furthermore, I found two, imo, a bit wired behaviour:

1.) I'm logged-in as dev@wepublish.ch, navigating to peers and then click on "create". I only receive a blank panel. Why is that? At least an error message should appear, shouldn't it?

image

2.) Still logged-in as dev@wepublish.ch, now navigating to settings. It seems, that the whole form is disabled. Shouldn't the menu entry not shown up at all or are these different permissions?

image

Itrulia (dismissed):

camilleory (approved):

Itrulia (dismissed):

Itrulia (approved):

camilleory (approved):

michael-scheurer (approved):

PR #649, opened 2022-07-26
fix(editor): editing peers required the token by accident

Itrulia:
Created a new function called toggleRequiredLabel to add the required symbol to labels.
This allows to have an easy way to toggle if it's required or not (adding/editing) and gives us a central way to change the icon if ever desired

### DIFF
--- a/packages/editor/__tests__/specs/__snapshots__/peerEditPanel.tsx.snap
+++ b/packages/editor/__tests__/specs/__snapshots__/peerEditPanel.tsx.snap
@@ -375,7 +375,7 @@ exports[`Peer Edit Panel should render with ID 1`] = `
               for="token"
               id="token-control-label"
             >
-              peerList.panels.token*
+              peerList.panels.token
             </label>
             <div
               class="rs-form-control rs-form-control-wrapper"
--- a/packages/editor/__tests__/specs/toggleRequiredLabel.test.ts
+++ b/packages/editor/__tests__/specs/toggleRequiredLabel.test.ts
@@ -1,0 +1,19 @@
+import {toggleRequiredLabel} from '../../src/client/toggleRequiredLabel'
+
+test('should add the required symbol by default', () => {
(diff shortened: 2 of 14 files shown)


packages/editor/__tests__/specs/__snapshots__/peerEditPanel.tsx.snap
```diff
@@ -375,7 +375,7 @@ exports[`Peer Edit Panel should render with ID 1`] = `
               for="token"
               id="token-control-label"
             >
-              peerList.panels.token*
```
Itrulia, line 378: There's still a snapshot where it's present to make sure it's there

packages/editor/__tests__/specs/toggleRequiredLabel.test.ts
```diff
@@ -0,0 +1,19 @@
+import {toggleRequiredLabel} from '../../src/client/toggleRequiredLabel'
```
camilleory, line 1: I'm curious about testing: is it necessary to write a test for such a small thing ? Why do you think it's necessary ?

camilleory (approved):
Nice solution 👍 
Should we remove the warning or the minimum length of 8 characters for passwords, as some users might have an old password which is shorter ? 
image

packages/editor/__tests__/specs/toggleRequiredLabel.test.ts
Itrulia: Mistakes happen, this just makes sure they don't in the future. Especially when refactoring things like this makes it easier.

If functions are small and pure usually writing tests are quick. I use snapshots for these kind of tests to use as regression testing, as if we change the symbol then we can just press `u` to update the snapshots easily.

There's things I agree that do not need to have unit tests (getters or setters for example), but anywhere there is logic involved, tests should be present.

penk85 (approved):
I like the re-usability :+1:

PR #651, opened 2022-07-26
ci: notify github/jira on successful deployment

Itrulia:


Itrulia:
You can see example Jira output here: https://wepublish.atlassian.net/browse/WPC-729.
Github is above.

Currently jira fails until the secrets are added. Notifications are set to optional.

tomaszdurka (commented):
@Itrulia wdyt about using github pull-request or jira ids instead of the ref-names? The latter ones could duplicates.

Itrulia:
@tomaszdurka could also do GITHUB_REF + random word at the end. I prefer having a readable URL instead of just a number, because in the URL you get some information about what it is about.

tomaszdurka (approved):

PR #655, opened 2022-08-08
feat(editor): disable publish for article/page without draft

Itrulia:
re #WPC-782


tomaszdurka (dismissed):

penk85 (changes requested):
Looks good, but should be on the preview button rather than the publish button.

Itrulia:
Oh Thank you @penk85! opening a new PR

PR #658, opened 2022-08-08
Npm publish only on tags push

tomaszdurka:


tomaszdurka:
@Itrulia any comments? :)

Itrulia:
Looks good :P

PR #662, opened 2022-08-16
docker compose fix

penk85:
`docker compose up` doesn't work consistently on all machines. Now changed to `docker-compose up`.


Itrulia (approved):

tomaszdurka (approved):
Hmmm, it's more of a fallback then fix :>

PR #669, opened 2022-08-22
F/wpc 773-polls-editor

michael-scheurer:


packages/api/src/graphql/poll/poll.private-mutation.ts
```diff
@@ -178,8 +181,8 @@ export const deletePollAnswer = async (
     throw new NotFound('PollAnswer', answerId)
   }
 
-  if (answerWithPoll.poll.answers.length <= 2) {
-    throw new GraphQLError('A poll requires atleast 2 answers')
+  if (answerWithPoll.poll.answers.length <= 1) {
+    throw new GraphQLError('A poll requires at least one answer.')
```
Itrulia, line 185: Any explanation for this change? :D

packages/editor/src/client/atoms/poll/deletePollModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  setPoll(poll: FullPoll | undefined): void
+  afterDelete(): Promise<ApolloQueryResult<PollsQuery>>
+}
+
+export function DeletePollModal({poll, setPoll, afterDelete}: deletePollProps) {
+  const {t} = useTranslation()
+  const [deletePollMutation] = useDeletePollMutation()
+  const [open, setOpen] = useState<boolean>(false)
+
+  useEffect(() => {
```
Itrulia, line 19: We probably should not have an useEffect here and just have open default to true.
We are not gaining much here but a double rerender.

Let's rather not render anything if poll does not exist.

packages/editor/src/client/atoms/poll/pollAnswers.tsx
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  /**
+   * FUNCTIONS
+   */
+  async function createAnswer() {
+    if (!poll) {
+      return
+    }
+    if (newAnswer === '') {
```
Itrulia, line 53:
```suggestion
    if (!newAnswer) {
```

As we want to check if newAnswer is empty, could lead to a bug and opens up the question when reading the code why only empty string.

packages/editor/src/client/atoms/poll/pollAnswers.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const updatedPoll = {
+      ...poll,
+      answers: updatedAnswers
+    }
+    await onPollUpdated(updatedPoll)
+  }
+
+  return (
+    <>
+      {/* iterate poll answers */}
```
Itrulia, line 108: Don't think we need this comment

packages/editor/src/client/atoms/poll/pollAnswers.tsx
```diff
@@ -0,0 +1,179 @@
...
+      onError: onErrorToast
+    })
+    await onAnswerDeleted()
+  }
+
+  async function updateAnswer(updatedAnswer: PollAnswerWithVoteCount) {
+    if (!poll) {
+      return
+    }
+    const updatedAnswers = [...(poll.answers || [])]
```
Itrulia, line 92:
```suggestion
    const updatedAnswers = poll.answers ? [...poll.answers] : []
```

A bit more readable for me

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const {t} = useTranslation()
+  const [newSource, setNewSource] = useState<string | undefined>(undefined)
+  const [sourceToDelete, setSourceToDelete] = useState<PollExternalVoteSource | undefined>(
+    undefined
+  )
+  const [openModal, setOpenModal] = useState<boolean>(false)
+  const [createSourceMutation, {loading}] = useCreatePollExternalVoteSourceMutation({
+    fetchPolicy: 'no-cache'
+  })
+  const [deletePollMutation] = useDeletePollExternalVoteSourceMutation({
```
Itrulia, line 42: I think this should be renamed as the name indicates the poll is being deleted.

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+  onExternalSourceChange(externalVoteSources: PollExternalVoteSource[]): void
+  savePoll(): Promise<void>
+}
+
+export function PollExternalVotes({
+  poll,
+  onExternalSourceCreated,
+  onExternalSourceDeleted,
+  onExternalSourceChange,
+  savePoll
```
Itrulia, line 31: Is this to fix that when the poll is not saved that we don't override it when we create a new external vote?
I think this is a bit hacky and is an unexpected user flow.

We probably should use a lazy query on the parent so the poll doesn't get updated all the time and thus overriding the data.

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+    source.voteAmounts[voteIndex].amount =
+      typeof newAmount === 'string' ? parseInt(newAmount) : newAmount
+
+    onExternalSourceChange(newVoteSources)
+  }
+
+  /**
+   * UI HELPER FUNCTIONS
+   */
+  function tableView() {
```
Itrulia, line 147: Could we make this pure (or even better a component) and move it out of the component?

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+                />
+              )}
+            </Table.Cell>
+          </Table.Column>
+        </Table>
+      </>
+    )
+  }
+
+  function iterateAnswerColumns() {
```
Itrulia, line 181: Same as above

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+                changeSource(answer, externalVoteSource, newValue)
+              }}
+            />
+          )}
+        </Table.Cell>
+      </Table.Column>
+    ))
+  }
+
+  function addSourceView() {
```
Itrulia, line 203: Same as above

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+            <IconButton icon={<PlusIcon />} appearance="primary" onClick={createPoll}>
+              {t('pollExternalVotes.addSourceBtn')}
+            </IconButton>
+          </Col>
+        </Row>
+      </>
+    )
+  }
+
+  function deleteModalView() {
```
Itrulia, line 225: Same as above

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
```diff
@@ -0,0 +1,265 @@
...
+            </Button>
+          </Modal.Footer>
+        </Modal>
+      </>
+    )
+  }
+
+  return (
+    <>
+      {/* table */}
```
Itrulia, line 257: Don't think we need those comments

packages/editor/src/client/base.tsx
```diff
@@ -158,7 +170,7 @@ export function Base({children}: BaseProps) {
                   {t('navbar.authors')}
                 </Nav.Item>
 
-                <Nav.Menu eventKey={'1'} title={t('navbar.usersAndMembers')} icon={<PeoplesIcon />}>
+                <Nav.Menu eventKey={'2'} title={t('navbar.usersAndMembers')} icon={<PeoplesIcon />}>
```
Itrulia, line 173: Out of interest, why the change? I also have no clue what eventKey does, as nothing changes when I press 1 or 2 :D

packages/editor/src/client/routes/polls/pollEditView.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const params = useParams()
+  const navigate = useNavigate()
+  const [poll, setPoll] = useState<FullPoll | undefined>(undefined)
+  const [close, setClose] = useState<boolean>(false)
+  const closePath = '/polls'
+
+  /**
+   * Handling toasts
+   */
+  const onErrorToast = (error: ApolloError) => {
```
Itrulia, line 30: Could we move these functions out of the component?

packages/editor/src/client/routes/polls/pollEditView.tsx
```diff
@@ -0,0 +1,256 @@
...
+    onError: onErrorToast,
+    onCompleted: onCompletedToast
+  })
+  const {t} = useTranslation()
+  const loading = createLoading || updateLoading
+
+  /**
+   * Update poll object after fetching from api
+   */
+  useEffect(() => {
```
Itrulia, line 65: using `useState` instead of `useMemo` causes a double render here, can we switch to useMemo instead?

packages/editor/src/client/routes/polls/pollEditView.tsx
```diff
@@ -0,0 +1,256 @@
...
+   */
+  async function saveOrUpdate(): Promise<void> {
+    if (!poll) {
+      return
+    }
+    const opensAt = poll.opensAt ? new Date(poll.opensAt).toISOString() : null
+    const closedAt = poll.closedAt ? new Date(poll.closedAt).toISOString() : null
+    const externalSources = poll.externalVoteSources?.map((voteSource: PollExternalVoteSource) => ({
+      ...voteSource,
+      __typename: undefined,
```
Itrulia, line 100: could also do `{__typename, ...voteSource}: PollExternalVoteSource` above

packages/editor/src/client/routes/polls/pollEditView.tsx
```diff
@@ -0,0 +1,256 @@
...
+                    }}
+                  />
+                </Panel>
+              </Col>
+              {/* poll external votes */}
+              <Col xs={24}>
+                <Panel header={t('pollEditView.pollExternalVotesPanelHeader')} bordered>
+                  <PollExternalVotes
+                    poll={poll}
+                    onExternalSourceCreated={async () => {
```
Itrulia, line 234: don't think we need async/await here

packages/editor/src/client/routes/polls/pollList.tsx
```diff
@@ -0,0 +1,200 @@
...
+    refetch({
+      take: limit,
+      skip: (page - 1) * limit
+    })
+  }, [page, limit])
+
+  /**
+   * UI HELPERS
+   */
+  function pollStateView(poll: FullPoll) {
```
Itrulia, line 57: Can we extract these functions as components?

packages/editor/src/client/routes/polls/pollList.tsx
```diff
@@ -0,0 +1,200 @@
...
+    const now = new Date()
+    const closedAt = poll.closedAt ? new Date(poll.closedAt) : undefined
+
+    // poll has been closed
+    if (closedAt && now.getTime() >= closedAt.getTime()) {
+      return (
+        <>
```
Itrulia, line 64: Fragments in this function are not needed

packages/editor/src/client/routes/polls/pollEditView.tsx
```diff
@@ -0,0 +1,256 @@
...
+                  <Form.Group controlId="question">
+                    <Form.Control
+                      name="question"
+                      placeholder={t('pollEditView.toBeOrNotToBe')}
+                      value={poll?.question || ''}
+                      onChange={(value: string) => {
+                        if (!poll) {
+                          return
+                        }
+                        setPoll({...poll, question: value})
```
Itrulia, line 165: setState's inside of event handlers in React 17 are batched (in React 18 all setStates), thus this can lead to a bug as `poll` could be stale.

```suggestion
                        setPoll(p => ({...p, question: value}))
```

Would fix this issue.

Itrulia (changes requested):
It's a lot of comments but don't worry overall this has been a very good PR. I have been commenting a lot on coding style that we currently have in our code base, so they are not really blocking by me but I would love to have them fixed going forward with new code, as over time this would improve the code base.

A few things I would like to have changed:

1. There's quite a few comments that we don't really need as the line below them describes already what is going on.

2. There are inline functions inside components that render other components. Those should be extracted as seperate components. Having them as functions inline makes them not reusable (and thus not *unit* testable) and depending on some compiler shenanigans (they are sometimes inlined or not) a memory issue as they get recreated every time.

3. There's an unexpected save for the user when creating a new external vote source. This is simply fixing a symptom and not the problem that the query requeries all the time and overrides the form state. Either we should have the form state fix and not update on the users behalf, or we should make the query a lazy query. This fixes the root cause of this issue.

packages/api/src/graphql/poll/poll.private-mutation.ts
michael-scheurer: Yes. Required by @LaurentAe.

packages/editor/src/client/base.tsx
michael-scheurer: Actually, I don't know, why that changed.

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
michael-scheurer: These functions are closely connected to it's component. It would not make any sense to export them as a new component. Rather it would generate boilerplate and complicate things up.

I just split this code into three parts, so that the code structure is more clear and clean and one line of code within a function has the same level of complexity.
image

packages/editor/src/client/routes/polls/pollEditView.tsx
michael-scheurer: would lead to "TS18004: No value exists in scope for the shorthand property '__typename'. Either declare one or provide an initializer."

packages/editor/src/client/routes/polls/pollList.tsx
michael-scheurer: yes. that's a good one to do so ;)

packages/editor/src/client/routes/polls/pollList.tsx
michael-scheurer: Yep. And added a tooltip too.

packages/editor/src/client/routes/polls/pollEditView.tsx
Itrulia: Oh is __typename not in PollExternalVoteSource?

packages/editor/src/client/atoms/poll/pollAnswers.tsx
```diff
@@ -66,7 +58,8 @@ export function PollAnswers({
     })
     const savedAnswer = answer?.data?.createPollAnswer
     if (savedAnswer) {
-      await onAnswerAdded(savedAnswer)
+      poll.answers?.push(savedAnswer as PollAnswerWithVoteCount)
```
Itrulia, line 61: This is mutating a parent object and shouldn't be done.

packages/editor/src/client/atoms/poll/pollExternalVotes.tsx
michael-scheurer: changed.

Itrulia (approved):
Looks great to me!

PR #670, opened 2022-08-22
feat(api): comment tagging & edit + create comment from editor

Itrulia:
```graphql
mutation createTag($tag:String,$type:TagType!) {
	createTag(tag:$tag,type:$type){
		id
		tag
        type
	}
}

// vars
{
  "tag":"Super Like2",
  "type":"Comment"
}
```

```graphql
mutation deleteTag($id:ID!) {
	deleteTag(id:$id){
		id
		tag
        type
	}
}

// vars
{
  "id":"abcd"
}
```

```graphql
mutation updateTag($id:ID!,$tag:String) {
	updateTag(id:$id,tag:$tag){
		id
		tag
        type
	}
}

// vars
{
  "id":"abcd",
  "tag":"Super Like",
}
```


packages/api/src/graphql/permissions.ts
```diff
@@ -483,6 +483,30 @@ export const CanUpdateSettings: Permission = {
...
+}
+
+export const CanUpdateTag: Permission = {
+  id: 'CAN_UPDATE_TAG',
+  description: 'Allows to update a tag',
+  deprecated: false
+}
+
+export const CanGetTag: Permission = {
+  id: 'CAN_GET_TAG',
```
michael-scheurer, line 499: Should probably be "CAN_GET_TAGS", so plural. In terms of consistency with other permissions.

packages/api/src/graphql/query.private.ts
```diff
@@ -638,6 +640,23 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+      args: {
+        cursor: {type: GraphQLID},
+        take: {type: GraphQLInt, defaultValue: 10},
+        skip: {type: GraphQLInt, defaultValue: 0},
+        filter: {type: GraphQLTagFilter},
+        sort: {type: GraphQLTagSort, defaultValue: TagSort.CreatedAt},
+        order: {type: GraphQLSortOrder, defaultValue: SortOrder.Descending}
+      },
+      resolve: (root, {filter, sort, order, cursor, take, skip}, {authenticate, prisma}) =>
+        getTags(filter, sort, order, cursor, skip, take, authenticate, prisma.tag)
```
michael-scheurer, line 659: I would use object as parameter ;)

packages/api/src/graphql/tag/tag.private-query.ts
```diff
@@ -0,0 +1,95 @@
...
+    tag.findMany({
+      where: where,
+      skip: skip,
+      take: Math.min(take, MaxResultsPerPage) + 1,
+      orderBy: orderBy,
+      cursor: cursorId ? {id: cursorId} : undefined
+    })
+  ])
+
+  const nodes = tags.slice(0, take)
```
michael-scheurer, line 78: Why is this necessary when you have the take property passed in the findMany() function?

packages/api/src/graphql/tag/tag.private-query.ts
```diff
@@ -0,0 +1,95 @@
...
+  const where = createTagFilter(filter)
+
+  const [totalCount, tags] = await Promise.all([
+    tag.count({
+      where: where,
+      orderBy: orderBy
+    }),
+    tag.findMany({
+      where: where,
+      skip: skip,
```
michael-scheurer, line 71: just skip

packages/api/src/graphql/tag/tag.private-query.ts
```diff
@@ -0,0 +1,95 @@
...
+  const orderBy = createTagOrder(sortedField, getSortOrder(order))
+  const where = createTagFilter(filter)
+
+  const [totalCount, tags] = await Promise.all([
+    tag.count({
+      where: where,
+      orderBy: orderBy
+    }),
+    tag.findMany({
+      where: where,
```
michael-scheurer, line 70: just where

packages/api/src/graphql/tag/tag.private-query.ts
```diff
@@ -0,0 +1,95 @@
...
+      skip: skip,
+      take: Math.min(take, MaxResultsPerPage) + 1,
+      orderBy: orderBy,
```
michael-scheurer, line 73: simplify

michael-scheurer (dismissed):
Only minimal remarks =)

packages/api/src/graphql/tag/tag.private-query.ts
Itrulia: Because with cursor based navigation we do not know if there's another page unfortunately

packages/api/src/graphql/query.private.ts
Itrulia: We currently have this everywhere from the prisma refactor, we should change that all at once for consistency sake in a separate PR

package.json
```diff
@@ -117,6 +117,7 @@
       "prettier/standard"
     ],
     "rules": {
+      "object-shorthand": "warn",
```
Itrulia, line 120: Automatically turns `{error: error}` into `{error}` with --fix

packages/api/src/graphql/tag/tag.private-query.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    case TagSort.CreatedAt:
+    default:
+      return {
+        createdAt: sortOrder
+      }
+  }
+}
+
+const createTypeFilter = (filter?: Partial<TagFilter>): Prisma.TagWhereInput => {
```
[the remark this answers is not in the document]
Itrulia: It's optional on graphql side (like the other filters) so that you don't always have to specify an empty filter on the client.

PR #671, opened 2022-08-22
Analytics readme

antkiewiczk:


Itrulia (approved):

docs/README.md
```diff
@@ -50,6 +50,42 @@ Prerequisite: In the API (packages/api) exists a corresponding GraphQL endpoint.
...
+
+then call the method on page load e.g.
+
+    const { current } = useRoute()
+    
+    useEffect(() => {
+    	analytics.page()
+    }, [current])
+
+The tracker will automatically look in the html structure for an element with a specific id and take data from this element. Example element to send peer tracking data should look like the following:
```
camilleory, line 81: Here I would add where the specific id is retrieved from, I was missing this explanation to understand how it works before asking you.

camilleory (approved):

docs/README.md
```diff
@@ -50,6 +50,42 @@ Prerequisite: In the API (packages/api) exists a corresponding GraphQL endpoint.
 4) Now the file `index.ts` will be generated automatically `in editor/src/client/api`
 5) Now you can import your desired endpoint in your .tsx file. See for example `subscriptionEditPanel.tsx`
 
+
+### Analytics
+In order to use analytics and send data to WePublish's Matomo, we prepared an analytics.js plugin that's available here:
```
tomaszdurka, line 55: I would not really mention about matomo directly here. Only note on the very bottom.

docs/README.md
```diff
@@ -50,6 +50,42 @@ Prerequisite: In the API (packages/api) exists a corresponding GraphQL endpoint.
...
+https://github.com/wepublish/analytics
+
+it integrates with https://github.com/DavidWells/analytics
+
+Usage:
+
+Import and initialise the tracker anywhere in the project
+
+    import  Analytics  from  'analytics'
+    import { matomoPlugin } from '@wepublish/analytics'
```
tomaszdurka, line 66: Also here, I would not mention it's `matomoPlugin`... 
Ah that's what we actually create in analytics. Not sure if that's good idea. If we change the underlying solution `matomoPlugin` doesn't correspond well.

What about using plugin `WepublishPlugin`?

tomaszdurka (changes requested):

docs/README.md
tomaszdurka: Yes, let's call the plugin `wepublish` so when you initiate "analytics" with "wepublish" plugin.

Itrulia (approved):

tomaszdurka (approved):

PR #678, opened 2022-08-27
Release helm chart to helm repository

tomaszdurka:
- Release helm chart
- Use on push for test purposes


Itrulia (approved):

.github/workflows/release-helm-chart.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  push:
+    branches:
+      - 'master'
+    paths:
+      - 'helm/charts/**/*'
+
+jobs:
+  push-chart-to-gitlab:
+    runs-on: ubuntu-latest
```
elias-summermatter, line 12: Is there a reason to use latest?  And not specific version? Its always better to use fixed version to prevent the pipeline from failing when the OS is upgraded.
Other question: Why not use debian minimal?

helm/charts/wepublish/templates/deployment-editor.yaml
```diff
@@ -31,13 +31,14 @@ spec:
         - name: {{ include "helm.fullname" . }}
           image: {{ .Values.deployment.image }}
           imagePullPolicy: {{ .Values.deployment.imagePullPolicy }}
+          command: ['npx', 'wepublish-editor']
           ports:
             - name: http
               containerPort: 8000
```
elias-summermatter, line 37: Is port 8000 correct?

helm/charts/wepublish/templates/deployment-api.yaml
```diff
@@ -31,13 +31,15 @@ spec:
         - name: {{ include "helm.fullname" . }}
           image: {{ .Values.deployment.image }}
           imagePullPolicy: {{ .Values.deployment.imagePullPolicy }}
+          command: ['/bin/sh']
+          args: ['-c', 'yarn migrate && npx wepublish-api']
           ports:
             - name: http
               containerPort: 8000
```
elias-summermatter, line 38: Is port 8000 correct?

.github/workflows/release-helm-chart.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - uses: goodsmileduck/helm-push-action@v2
+        env:
+          SOURCE_DIR: 'helm/charts'
+          CHART_FOLDER: 'wepublish'
+          FORCE: 'True'
+          CHARTMUSEUM_URL: 'https://gitlab.com/api/v4/projects/38941346/packages/helm/stable'
+          CHARTMUSEUM_USER: 'wepublish'
```
elias-summermatter, line 21: I would store this as a secret too

helm/charts/wepublish/values.yaml
```diff
@@ -5,13 +5,26 @@ deployment:
   env: {}
 
   api:
-    env: {}
+    env:
+      GOOGLE_APPLICATION_CREDENTIALS: /var/secrets/google/key.json
+      MEDIA_SERVER_URL: https://media
+      MEDIA_ADDRESS: media_server_address
+      MEDIA_PORT: "443"
+      DATABASE_URL: postgres://
```
elias-summermatter, line 13: Why not use different place holders?

elias-summermatter (changes requested):

helm/charts/wepublish/templates/deployment-editor.yaml
elias-summermatter: Forget that one

helm/charts/wepublish/templates/deployment-api.yaml
elias-summermatter: Forget that one

helm/charts/wepublish/values.yaml
tomaszdurka: You mean why put everything into the ENV?
Just ease of adding things

.github/workflows/release-helm-chart.yml
tomaszdurka: Tbh, just copy paste from other place.

.github/workflows/release-helm-chart.yml
tomaszdurka: Done

.github/workflows/release-helm-chart.yml
tomaszdurka: I see we have ubuntu-latest for all of our actions, so let' s keep it.
We can discuss/change it for all in other ticket.

PR #682, opened 2022-08-30
Create entrypoint and bin for api

tomaszdurka:


packages/api/package.json
```diff
@@ -23,6 +23,9 @@
   },
   "typings": "./lib/index.d.ts",
   "license": "MIT",
+  "bin": {
```
Itrulia, line 26: Could also just be:

`"bin": "./bin/run-server.js"``, then it can be called via `npx @wepublish/api`

tomaszdurka:
@Itrulia please re-review

Itrulia:
Looks good!

Itrulia:
@tomaszdurka the merge broke the build

PR #686, opened 2022-09-05
Recognized Metadata if image is uploaded direct via article or page

elias-summermatter:


packages/editor/src/client/atoms/imageMetaData.tsx
```diff
@@ -0,0 +1,57 @@
+import exifr from 'exifr'
+
+export type ImageMetaData = {
```
Itrulia, line 3: nit: could also do: `Record<string, string>`

Itrulia (approved):

michael-scheurer (approved):

PR #687, opened 2022-09-06
Docker builds, entrypoints improvements

tomaszdurka:
- Add example packages binaries


examples/api/package.json
```diff
@@ -2,26 +2,23 @@
   "name": "@wepublish/api-example",
   "version": "0.0.0",
   "private": true,
-  "bin": {
-    "wepublish-api-example": "./dist/index.js"
-  },
   "main": "./dist/",
   "license": "MIT",
+  "bin": {
```
Itrulia, line 7: do we maybe want to do `"bin": "./bin/run-server.js"`?

examples/oauth2/package.json
```diff
@@ -2,18 +2,18 @@
   "name": "@wepublish/oauth2-example",
   "version": "0.0.0",
   "private": true,
-  "bin": {
-    "wepublish-oauth2-example": "./dist/index.js"
-  },
   "main": "./dist/",
   "license": "MIT",
+  "bin": {
```
Itrulia, line 7: Same as above

examples/website/package.json
```diff
@@ -8,6 +8,9 @@
     "node": "^12.0.0",
     "yarn": "^1.16.0"
   },
+  "bin": {
+    "wepublish-example-website": "./bin/run-server.js"
```
Itrulia, line 12: idem

packages/api/prisma/seed-base.ts
```diff
@@ -0,0 +1,140 @@
+import {PrismaClient} from '@prisma/client'
```
Itrulia, line 1: Documentation has to be updated now. This also makes a breaking change.

Itrulia (changes requested):

packages/api/prisma/seed-base.ts
tomaszdurka: That was something I wanted to discuss. Did that in the last moment as could not make seed run without example.

packages/api/prisma/seed-base.ts
```diff
@@ -0,0 +1,140 @@
...
+      id: 'peer',
+      systemRole: true,
+      name: 'Peer',
+      description: 'Peer Role',
+      permissionIDs: []
+    }
+  })
+]
+
+const seedUsers = async (prisma: PrismaClient) => [
```
Itrulia, line 116: Should be moved out into a different seeding file, see prisma documentation

elias-summermatter (approved):

elias-summermatter (changes requested):
Tests also not working

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/wepublish/node:latest
+FROM ghcr.io/wepublish/node:latest as production
```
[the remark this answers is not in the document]
Itrulia: Might our custom node images be the reason it takes around 3 minutes for the images to start once pulled by gke?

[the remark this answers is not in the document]
tomaszdurka: It is using official just adding stuff on top of it.
https://github.com/wepublish/node-docker-base/blob/main/dockerfile

PR #699, opened 2022-09-14
F/wpc 863 - reply to comments from editor

michael-scheurer:


packages/api/prisma/migrations/20220914164547_make_text_in_comment_revision_nullable/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "comments.revisions" ALTER COLUMN "text" DROP NOT NULL;
```
michael-scheurer, line 2: Needed to save a revision, where user only wants to save a title, but no text

Itrulia: What do you think of copying over the text? Would make things easier maybe down the line

michael-scheurer: How do you mean exactly by "copying over the text"? Copy into an existing migration file?

Itrulia: When you create a new revision you copy over all the new data and only update what you send to update.

michael-scheurer: yes. that is exactly the use case.

packages/api/src/graphql/comment/comment.ts
```diff
@@ -208,7 +208,7 @@ export const GraphQLComment: GraphQLObjectType<Comment, Context> = new GraphQLOb
       )
     },
     revisions: {
-      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLCommentRevision)))
+      type: GraphQLList(GraphQLNonNull(GraphQLCommentRevision))
```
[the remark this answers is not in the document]
michael-scheurer: revisions can simply be nullish and doesn't return any array. at least I solved the graphQL error with that change.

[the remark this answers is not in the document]
Itrulia: The typings have revisions as never null even when empty, trying it out locally too.
When do you get revisions being null?

I think that error happens because 

```
includes: {
  revisions: true
}
```

is missing somewhere. This causes revisions not to be there

packages/editor/src/client/routes/comments/commentList.tsx
```diff
@@ -554,7 +563,7 @@ function CommentList() {
                               displayOnly
                               // TODO: remove this
                               onChange={console.log}
-                              value={text}
+                              value={text || []}
```
Itrulia, line 566: This will not work as an empty array is invalid for Slate. See https://github.com/wepublish/wepublish/blob/master/packages/editor/src/client/routes/comments/commentEditView.tsx#L31

Itrulia (changes requested):

packages/api/src/graphql/comment/comment.ts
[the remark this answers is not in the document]
michael-scheurer: fixed

Itrulia (dismissed):

packages/editor/src/client/routes/comments/commentList.tsx
```diff
@@ -321,57 +200,37 @@ function CommentList() {
...
-                      value={rowData?.revisions[rowData?.revisions?.length - 1].text}
+                      value={rowData.revisions[rowData.revisions?.length - 1]?.text || []}
                     />
                   ) : null}
                 </>
               )}
             </Cell>
           </Column>
-          <Column width={150} align="left" resizable>
+          {/* eslint-disable-next-line i18next/no-literal-string */}
```
michael-scheurer, line 223: align={'left'} did not work either. Even though, it should according to https://github.com/edvardchen/eslint-plugin-i18next/blob/next/docs/rules/no-literal-string.md

penk85 (dismissed):

packages/editor/src/client/panel/selectCommentsPanel.tsx
```diff
@@ -194,7 +194,7 @@ export function SelectCommentPanel({
                       disabled
                       // TODO: remove this
```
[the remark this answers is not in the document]
michael-scheurer: copy & pasted that. we would have to change the RichTextBlock though.

packages/editor/src/client/panel/selectCommentsPanel.tsx
```diff
@@ -194,7 +194,7 @@ export function SelectCommentPanel({
...
                       onChange={console.log}
-                      value={rowData?.revisions[rowData?.revisions?.length - 1].text}
+                      value={rowData?.revisions[rowData?.revisions?.length - 1]?.text || []}
```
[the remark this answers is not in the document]
michael-scheurer: yes. it's in Slate structure, so an array, yes

penk85 (approved):

penk85 (approved):

PR #704, opened 2022-09-18
Replace scripts with executables (HTML block)

antkiewiczk:
RE: https://wepublish.atlassian.net/browse/WPC-516?focusedCommentId=12545

### DIFF
--- a/packages/editor/src/client/blocks/htmlBlock.tsx
+++ b/packages/editor/src/client/blocks/htmlBlock.tsx
@@ -1,16 +1,15 @@
 import PencilIcon from '@rsuite/icons/legacy/Pencil'
-import React, {useEffect, useRef, useState} from 'react'
+import InnerHTML from 'dangerously-set-html-content'
+import React, {useEffect, useState} from 'react'
 import {useTranslation} from 'react-i18next'
 import {Drawer, IconButton, Panel} from 'rsuite'
 
 import {BlockProps} from '../atoms/blockList'
 import {PlaceholderInput} from '../atoms/placeholderInput'
 import {HtmlEditPanel} from '../panel/htmlEditPanel'
-import {executeScriptElements} from '../utility'
 import {HTMLBlockValue} from './types'
 
 export const HTMLBlock = ({value, onChange, autofocus}: BlockProps<HTMLBlockValue>) => {
-  const scriptContainer = useRef<HTMLDivElement>(null)
   const [isHtmlDialogOpen, setHtmlDialogOpen] = useState(false)
   const isEmpty = !value.html
   const {t} = useTranslation()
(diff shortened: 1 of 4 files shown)


packages/editor/src/client/blocks/htmlBlock.tsx
```diff
@@ -19,6 +20,10 @@ export const HTMLBlock = ({value, onChange, autofocus}: BlockProps<HTMLBlockValu
     }
   }, [])
 
+  useEffect(() => {
+    executeScriptElements(document.getElementById('html-block') as HTMLElement)
```
Itrulia, line 24: Problem with ID is that it's unique on the page, we should probably not use id here.

instead of document.getElementById we probably should use use `<ref>.querySelectorAll('.html-block')`  instead of document.

packages/editor/src/client/panel/htmlEditPanel.tsx
```diff
@@ -42,7 +42,6 @@ export function HtmlEditPanel({value, onClose, onConfirm}: HtmlEditPanelProps) {
         <div style={{marginTop: '20px'}}>
           <span>{t('blocks.html.warning')}</span>
         </div>
-        <div style={{marginTop: '20px'}} dangerouslySetInnerHTML={{__html: htmlBlock.html}} />
```
Itrulia, line 45: I guess this is intended because of security?

Itrulia (changes requested):

packages/editor/src/client/blocks/htmlBlock.tsx
```diff
@@ -19,6 +20,12 @@ export const HTMLBlock = ({value, onChange, autofocus}: BlockProps<HTMLBlockValu
     }
   }, [])
 
+  useEffect(() => {
+    executeScriptElements(
+      document.querySelectorAll('div[data-html-block]') as NodeListOf<HTMLElement>
```
Itrulia, line 25: This will execute them multiple times now when you have multiple HTML block.
We should use a ref, attach it to the div and do `<ref>.querySelector`

Itrulia (approved):

packages/editor/src/client/utility.ts
```diff
@@ -247,3 +247,23 @@ export function isValueConstructor<T>(value: T | (() => T)): value is () => T {
 export function isFunctionalUpdate<T>(value: React.SetStateAction<T>): value is (value: T) => T {
   return typeof value === 'function'
 }
+
+// replace scripts with executable ones
+export function executeScriptElements(containerElements: NodeListOf<HTMLElement> | HTMLElement[]) {
+  if (!containerElements.length) return
+
+  Array.from(containerElements).forEach(container => {
```
michael-scheurer, line 255: Why do we need exactly three inner forEach() ? Looks a bit shady. Can we resolve that somehow or program it more generic?

michael-scheurer (changes requested):
Hey Kamil. I have some remarks and also see my comment in the code.

- The website doesn't render the blocks properly but fails somehow. See this screenshot
image

- There is another bug in the editor when changing the metadata. See the following screencast:
[Screencast from 27.09.2022 11:20:46.webm](https://user-images.githubusercontent.com/70939054/192487634-4f45ad22-7200-406a-941c-49148ca4e01e.webm)

michael-scheurer (approved):

penk85 (approved):

PR #707, opened 2022-09-22
chore: add git hooks to rebuild prisma when switching or merging branches

Itrulia:
automatically regenerates the prisma schema in case you switch branch or merge. Different branches or merges can cause a different schema so one has to regenerate it to fix build issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,9 @@
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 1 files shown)


penk85 (approved):

tomaszdurka (approved):
OK

PR #708, opened 2022-09-22
feat(api): user votes/rating api

Itrulia:


packages/api/src/graphql/query.public.ts
```diff
@@ -277,13 +278,25 @@ export const GraphQLPublicQuery = new GraphQLObjectType<undefined, Context>({
     comments: {
       type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLPublicComment))),
       args: {
-        itemId: {type: GraphQLID}
+        itemId: {type: GraphQLNonNull(GraphQLID)}
```
Itrulia, line 281: Breaking change but wasn't released yet

michael-scheurer (approved):

PR #718, opened 2022-10-06
feat(api,editor,example/website): Custom Teaser

antkiewiczk:


packages/api/src/graphql/blocks.ts
```diff
@@ -169,9 +175,31 @@ export const GraphQLPageTeaser = new GraphQLObjectType<PageTeaser, Context>({
...
+      type: GraphQLImage,
+      resolve: createProxyingResolver(({imageID}, {}, {loaders}) =>
+        imageID ? loaders.images.load(imageID) : null
+      )
+    },
+
+    preTitle: {type: GraphQLString},
+    title: {type: GraphQLString},
+    lead: {type: GraphQLString},
+    contentUrl: {type: GraphQLString},
```
tomaszdurka, line 193: This is optional?

packages/api/src/graphql/blocks.ts
```diff
@@ -323,9 +351,38 @@ export const GraphQLPublicPageTeaser = new GraphQLObjectType<PageTeaser, Context
...
```
tomaszdurka, line 369: Again here, title maybe the same?

packages/editor/src/client/blocks/types.ts
```diff
@@ -218,21 +220,32 @@ export interface PageTeaserLink {
...
+  | CustomTeaserLink
 
 export interface BaseTeaser {
   style: TeaserStyle
   image?: ImageRefFragment
   preTitle?: string
   title?: string
   lead?: string
+  contentUrl?: string
+  properties?: TeaserMetadataProperty[]
```
tomaszdurka, line 240: This is actually different in graphql? `NonNull(List)`

tomaszdurka (changes requested):

packages/api/src/graphql/blocks.ts
tomaszdurka: No. Definition is ofc required. I understand to make it non-optional you need specify that as:
`{type: GraphQLNonNull(GraphQLString)}`
instead of
`{type:GraphQLString}`

packages/editor/src/client/blocks/types.ts
```diff
@@ -218,7 +220,17 @@ export interface PageTeaserLink {
   page: PageRefFragment
 }
 
-export type TeaserLink = ArticleTeaserLink | PeerArticleTeaserLink | PageTeaserLink
+export interface CustomTeaserLink extends BaseTeaser {
+  type: TeaserType.Custom
+  contentUrl?: string
+  properties?: TeaserMetadataProperty[]
```
tomaszdurka, line 226: One more time about this.

Is it possible to have undefined properties, or we should always force array -> empty array?

packages/api/src/graphql/blocks.ts
```diff
@@ -169,9 +175,31 @@ export const GraphQLPageTeaser = new GraphQLObjectType<PageTeaser, Context>({
...
+    properties: {type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLMetadataProperty)))}
```
tomaszdurka, line 194: Here we say they always there. Never undefined/null.

tomaszdurka (commented):
lgtm still please check comments about the props and let me know how you see it.

tomaszdurka (approved):

penk85 (approved):

Itrulia (approved):

PR #726, opened 2022-10-13
WPC-861: As a subscriber, i want to renew my subscription before it has expired

elias-summermatter:


packages/api/src/payments/paymentProvider.ts
```diff
@@ -240,6 +240,13 @@ export function setupPaymentProvider(opts: WepublishServerOpts): Router {
 
       const {items, ...invoiceData} = invoice
 
+      items.forEach(item => {
+        // Typing problem: Fix me @TODO
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
elias-summermatter, line 245: @Itrulia do you have an idea how i could fix this?

elias-summermatter: image

Itrulia:
```
const {items, ...invoiceData} = invoice

      await prisma.invoice.update({
        where: {id: invoice.id},
        data: {
          ...invoiceData,
          items: {
            deleteMany: {
              invoiceId: invoiceData.id
            },
            create: items.map(({invoiceId, ...item}) => item)
          },
          paidAt: new Date(),
          canceledAt: null
        }
      })
 ```
 
 should do the trick, also no mutation. Else we could also set invoiceId to null since the prisma model has it as nullable but not able to be undefined

PR #727, opened 2022-10-17
Revert "chore: add git hooks to rebuild prisma when switching or merging branches"

tomaszdurka:
@Itrulia that's rather irritating feature tbh. Especially as I switch from branch to branch a lot. What about rather having watch command or regenerate with dev command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,9 +92,7 @@
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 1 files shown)


Itrulia:
I dont really mind as I do it manually, this came from a suggestion by @antkiewiczk. But shouldn't this run in the background?

Itrulia (approved):

PR #733, opened 2022-10-20
feat(api): allow sorting of comments by rating

Itrulia:


tomaszdurka (approved):

tomaszdurka:
Thought: I wish we use props for those complicated optional params functions.

PR #735, opened 2022-10-24
feat(api,editor): extend Article, Peer Article and Page filters

antkiewiczk:


Itrulia (dismissed):

Itrulia (approved):

packages/editor/src/client/atoms/searchAndFilter/articleListFilter.tsx
```diff
@@ -0,0 +1,137 @@
+import CloseIcon from '@rsuite/icons/legacy/Close'
```
penk85, line 1: Are the article, page, and peer article filters all the same? Seems like it would make sense to have a generic filter and avoid repeating it, no?

penk85 (commented):
Looks great!

penk85 (dismissed):

packages/api/src/graphql/article/article.queries.ts
```diff
@@ -58,6 +58,84 @@ const createTitleFilter = (filter: Partial<ArticleFilter>): Prisma.ArticleWhereI
...
+
+    return {
+      OR: [{draft: containsPreTitle}, {pending: containsPreTitle}, {published: containsPreTitle}]
+    }
+  }
+
+  return {}
+}
+
+const createLeadFilter = (filter: Partial<ArticleFilter>): Prisma.ArticleWhereInput => {
```
michael-scheurer, line 78: I would consider writing a generic filter function at least for string filters. Otherwise this is just unnecessary code duplication.

packages/api/src/graphql/page/page.queries.ts
```diff
@@ -58,6 +58,67 @@ const createTitleFilter = (filter: Partial<PageFilter>): Prisma.PageWhereInput =
...
+  }
+
+  return {}
+}
+
+const mapDateFilterToPrisma = (comparison: DateFilterComparison): keyof Prisma.DateTimeFilter => {
+  return comparison === DateFilterComparison.Equal ? 'equals' : comparison
+}
+
+const createPublicationDateFromFilter = (filter: Partial<PageFilter>): Prisma.PageWhereInput => {
```
michael-scheurer, line 86: Generic date filter function?

packages/api/src/graphql/peer-article/peer-article.private-queries.ts
```diff
@@ -169,38 +170,65 @@ export const getAdminPeerArticles = async (
...
+    filtered = filtered.filter(({article}) =>
+      article.latest.lead.toLowerCase().includes(filter.lead?.toLowerCase())
+    )
+  }
+  if (filter.publicationDateFrom && filter.publicationDateTo) {
+    const from = filter.publicationDateFrom.date as Date
+    const to = filter.publicationDateTo.date as Date
+    filtered = filtered.filter(
+      ({article}) =>
+        new Date(article.published.publishedAt).getTime() > new Date(from).getTime() &&
```
michael-scheurer, line 195: Should be greater OR EQUAL, I think

packages/editor/src/client/atoms/searchAndFilter/articleListFilter.tsx
michael-scheurer: yes. I totally agree. We could think of a component where you can pass an array with all the filter you wanna display.

michael-scheurer (changes requested):
Hi Kamil
First of all am I very happy, that we finally improve the filters. And it looks nice!

General remarks
- We should add tags and author to the filter
- The pending state is missing (only draft and published)
- The reset btn sometimes is not aligned
image

packages/editor/src/client/atoms/searchAndFilter/listViewFilters.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+export function ListViewFilters({
+  fields,
+  filter,
+  onSetFilter,
+  setPeerFilter
+}: ListViewFiltersProps) {
+  const {t} = useTranslation()
+  const [resetFilterKey, setResetFilterkey] = useState<string>(new Date().getTime().toString())
+  const [allPeers, setAllPeers] = useState<PeerWithProfileFragment[]>([])
```
Itrulia, line 50: Should be a useMemo or a simple const (const preferred).

```
const allPeers = useMemo(() => peerListData?.peers, [peerListData?.peers])
// or even better
const allPeers = peerListData?.peers;
```

As this double renders

packages/editor/src/client/atoms/searchAndFilter/listViewFilters.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+  const isPeerFilter = fields.includes('peer') && setPeerFilter
+
+  let peerListData = [] as any
+  if (isPeerFilter) {
```
Itrulia, line 55: This goes against the rule of hooks: https://reactjs.org/docs/hooks-rules.html
Should use a lazy query instead  and have the if inside a useEffect that executes the query.

Itrulia (changes requested):

Itrulia (approved):

PR #751, opened 2022-11-09
fix(api): focalpoint missing in queries and mutations

Itrulia:


michael-scheurer (changes requested):
Found the following. If I upload the image and I set the focal point in the top left corner it looks like this:
image

If I edit the same Image again. The focal point seems shifted
image

I assume, this is only a representation issue in the frontend.

Also, I would not point the PR to master because this should be treated as a fix for a new minor release without shipping any new features, which would be the case, if we merge into master.

@elias-summermatter @tomaszdurka

PR #755, opened 2022-11-09
fix(editor): gif or other unsupported images by exfir upload broken

Itrulia:
ignore metadata of images not supported by exifr

they only support a few file formats: https://github.com/MikeKovarik/exifr/tree/8ba7df23f8f58e7b817ef84e6dbb9c47a53e2315/src/file-parsers


michael-scheurer (changes requested):
Like mentioned [in the comment here](https://github.com/wepublish/wepublish/pull/751#pullrequestreview-1176180268), we should not merge this branch into master but into a fix release (which will then be merged back into master).

PR #767, opened 2022-11-15
Backport patch: Fix subscription csv download

michael-scheurer:


packages/api/src/utility.ts
```diff
@@ -49,7 +49,7 @@ export function mapSubscriptionsAsCsv(subscriptions: Subscription[]) {
       'paymentMethodID',
       'deactivationDate',
       'deactivationReason'
-    ].join(',') + '\n'
+    ].join(';') + '\n'
```
michael-scheurer, line 52: Sometimes, users use commas in their fields such as the address (e.g. example street, 13). This causes wrong column separation within the csv file.

packages/api/src/graphql/query.private.ts
```diff
@@ -435,7 +435,35 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+        }
+
+        // map users on subscriptions
+        for (const subscription of subscriptions) {
+          const userIndex = users.findIndex(tmpUser => tmpUser.id === subscription.userID)
+          if (userIndex < 0) {
+            subscriptionsWithUser.push(subscription)
+            continue
+          }
+          subscriptionsWithUser.push({...subscription, user: users[userIndex]})
```
michael-scheurer, line 463: The properties of subscription are readonly.

packages/api/src/utility.ts
elias-summermatter: Shouldn't we be RFC compliant https://www.ietf.org/rfc/rfc4180.txt? Since CSV stands for: Comma-Separated Values Escape commas?

packages/api/src/graphql/query.private.ts
```diff
@@ -435,7 +435,35 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+            order: SortOrder.Descending
+          })
+          users.push(...userResults.nodes)
+          hasMore = userResults.pageInfo.hasNextPage
+          afterCursor = userResults.pageInfo.endCursor
+        }
+
+        // map users on subscriptions
+        for (const subscription of subscriptions) {
+          const userIndex = users.findIndex(tmpUser => tmpUser.id === subscription.userID)
```
elias-summermatter, line 458: Why tmp User could we not change this to user?

packages/api/src/graphql/query.private.ts
```diff
@@ -415,7 +415,7 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
         const joins: SubscriptionJoins = {
           joinMemberPlan: true,
           joinPaymentMethod: true,
-          joinUser: true
+          joinUser: false
```
elias-summermatter, line 418: @michael-scheurer so you identified that just the join of User schema is the problem right?

packages/api/src/graphql/query.private.ts
```diff
@@ -435,7 +435,35 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+          users.push(...userResults.nodes)
+          hasMore = userResults.pageInfo.hasNextPage
+          afterCursor = userResults.pageInfo.endCursor
+        }
+
+        // map users on subscriptions
+        for (const subscription of subscriptions) {
+          const userIndex = users.findIndex(tmpUser => tmpUser.id === subscription.userID)
+          if (userIndex < 0) {
+            subscriptionsWithUser.push(subscription)
```
elias-summermatter, line 460: SInce you not always got an user (if userIndex = -1) the name subscriptionsWithUser is not very descriptive. Maybe better something like subscriptionUserMap or something better. :-)

elias-summermatter (changes requested):
Naming changes are optional. So if you think they are ok thats finde for me too :-)

packages/api/src/graphql/query.private.ts
michael-scheurer: correct.

packages/api/src/utility.ts
michael-scheurer: Good point. Alternatively, I can sanitize all user input files. What do you think? Would that be a better solution? Probably, right?

packages/api/src/graphql/query.private.ts
michael-scheurer: I'm always a bit concerned, that scoping is not working properly. But yeah. I'll change it.

packages/api/src/utility.ts
michael-scheurer: Ok. Refactored that.

packages/api/src/utility.ts
```diff
@@ -93,6 +93,11 @@ export function mapSubscriptionsAsCsv(subscriptions: Subscription[]) {
   return csvStr
 }
 
+// Removes commas and new lines within csv content. Avoids disrupted csv formatting.
+function sanitizeCsvContent(input: string | undefined): string {
```
elias-summermatter, line 97: Shouldn't we escape how it is described in the RFC:
https://www.ietf.org/rfc/rfc4180.txt
image

image

elias-summermatter (changes requested):

Itrulia (approved):

elias-summermatter (approved):

PR #768, opened 2022-11-15
csv export according rfc standards

michael-scheurer:


packages/api/src/utility.ts
```diff
@@ -55,7 +55,7 @@ export function mapSubscriptionsAsCsv(
       'paymentMethodID',
       'deactivationDate',
       'deactivationReason'
-    ].join(',') + '\n'
+    ].join(';') + '\n'
```
elias-summermatter, line 58: Same objection as described in in https://github.com/wepublish/wepublish/pull/767/files#r1022846852

> Shouldn't we be RFC compliant https://www.ietf.org/rfc/rfc4180.txt? Since CSV stands for: Comma-Separated Values Escape commas?

elias-summermatter (changes requested):

packages/api/src/utility.ts
michael-scheurer: solved

Itrulia (dismissed):

antkiewiczk (dismissed):

packages/api/src/utility.ts
```diff
@@ -99,6 +99,11 @@ export function mapSubscriptionsAsCsv(
   return csvStr
 }
 
+// Removes commas and new lines within csv content. Avoids disrupted csv formatting.
+function sanitizeCsvContent(input: string | undefined | null): string {
+  return (input || '').toString().replace(/[#,]|\r?\n|\r/g, '')
```
elias-summermatter, line 104: Copy rfc sanitize from other pull request :-)

elias-summermatter (changes requested):

packages/api/src/utility.ts
michael-scheurer: done

elias-summermatter (dismissed):

Itrulia (dismissed):

antkiewiczk (dismissed):

packages/api/src/utility.ts
```diff
@@ -65,20 +65,20 @@ export function mapSubscriptionsAsCsv(
     csvStr +=
       [
         user?.id,
-        `${user?.firstName ?? ''}`,
-        `${user?.name ?? ''}`,
-        `${user?.preferredName ?? ''}`,
+        `${sanitizeCsvContent(user?.firstName)}`,
```
tomaszdurka, line 68: @michael-scheurer There is no reason to wrap it in backticks/interpolation here, right? Would remove for readability.

Itrulia (approved):

PR #770, opened 2022-11-17
fix(api): slugs and titles of pages to be nullable

Itrulia:
Not a breaking change as just private api

### DIFF
--- a/packages/api/__tests__/api/private/index.ts
+++ b/packages/api/__tests__/api/private/index.ts
@@ -1359,8 +1359,8 @@ export type PageRevision = {
   publishAt?: Maybe<Scalars['DateTime']>
   updatedAt?: Maybe<Scalars['DateTime']>
   publishedAt?: Maybe<Scalars['DateTime']>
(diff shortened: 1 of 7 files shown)


michael-scheurer (commented):
Not sure, if we really want to solve the problem, simply by making the slug voluntary? Just as an idea: How about the application shows a dialog to the user, before copying a page? In the dialog the user would have to fill in a title and slug (even better, if the slug is auto-generated out of the title, but still editable, cf. Wordpress).

Itrulia:
This is how article already works, currently we just set things to empty strings because they can't be null which is a hack IMO.

Empty strings are also being hit by unique keys while null is not.

michael-scheurer:
@Itrulia what do you think or how do you feel: time to resolve that hack somehow similar as I suggested? I probably would do it. One hack less ;)

penk85 (dismissed):

penk85 (approved):

tomaszdurka (approved):

PR #771, opened 2022-11-18
feat(editor): add properties fields to user edit

antkiewiczk:


michael-scheurer (changes requested):
Looks good. One small bug: If I create a new user with properties, the properties are not created, but only the user.

Itrulia (approved):

Itrulia (approved):

michael-scheurer (approved):

penk85 (approved):

PR #772, opened 2022-11-22
newsroom api

penk85:
Combine peer & peer profile to newsroom table.


Itrulia:
I guess you merged from master? The branch you are pointing to is behind master so the PR is a bit hard to read right now. Would you mind merging the other branch also from master?

packages/api/src/context.ts
```diff
@@ -667,7 +667,7 @@ export async function contextFromRequest(
               )?.value as number) ||
               parseInt(process.env.PEERING_TIMEOUT_IN_MS as string) ||
               3000
-            const fetcher = createFetcher(peer.hostURL, peer.token, peerTimeout)
+            const fetcher = createFetcher(peer?.hostURL ?? '', peer?.token ?? '', peerTimeout)
```
Itrulia, line 670: Don't think this change is needed (the peer?.) as if peer is not defined we return

Should we maybe also check if hostURL or token is set? Else we might make an HTTP call we didn't want to.

packages/api/src/context.ts
```diff
@@ -702,8 +702,8 @@ export async function contextFromRequest(
               parseInt(process.env.PEERING_TIMEOUT_IN_MS as string) ||
               3000
             const fetcher = createFetcher(
-              url.resolve(peer.hostURL, 'admin'),
-              peer.token,
+              url.resolve(peer?.hostURL ?? '', 'admin'),
```
Itrulia, line 705: Same thing here

packages/api/src/db/article.ts
```diff
@@ -49,7 +50,7 @@ export interface Article extends ArticleData {
 }
 
 export interface PeerArticle {
-  peerID: string
+  newsroomID: string
```
Itrulia, line 53: Since we are making breaking changes, could we actually call this newsroomId instead? Would be amazing :D

Itrulia (changes requested):
Haven't tested it, looks amazing! Just have a few small comments about peeringSchema

packages/api/src/context.ts
```diff
@@ -659,6 +659,7 @@ export async function contextFromRequest(
               console.error(peer)
               return null
             }
+            if (!peer.hostURL || !peer.token) return null
```
Itrulia, line 662:
```suggestion

            if (!peer.hostURL || !peer.token) {
              return null
            }
            
```

packages/api/src/context.ts
```diff
@@ -693,6 +694,7 @@ export async function contextFromRequest(
...
```
Itrulia, line 697:
```suggestion

            if (!peer.hostURL || !peer.token) {
              return null
            }
            
```

packages/editor/src/client/blocks/teaserGridBlock.tsx
```diff
@@ -176,15 +176,15 @@ export function TeaserBlock({
...
-              {teaser.type !== TeaserType.PeerArticle || !teaser?.peer?.isDisabled ? (
-                <IconButton
-                  icon={<PencilIcon />}
-                  onClick={onEdit}
-                  style={{
-                    margin: 10
-                  }}
-                />
-              ) : null}
+              {/* {teaser.type !== TeaserType.PeerArticle || !teaser?.peer?.isDisabled ? ( */}
```
Itrulia, line 179: Oh just saw can we remove this or bring it back in?

packages/editor/src/client/blocks/teaserGridBlock.tsx
```diff
@@ -238,7 +238,7 @@ export function contentForTeaser(teaser: Teaser, numColumns?: number) {
           title={teaser.title ?? teaser.article?.latest.title ?? ''}
           lead={teaser.lead ?? teaser.article?.latest.lead ?? undefined}
           states={states}
-          peer={teaser.peer}
+          // peer={teaser.peer}
```
Itrulia, line 241: Same thing here

PR #773, opened 2022-11-22
feat(api): allow user to create events to display on the frontend

Itrulia:
Currently no editor block, comes with WPC-950


packages/api/__tests__/api/public/index.ts
```diff
@@ -298,6 +298,44 @@ export type EmbedBlock = {
...
+
+export type EventConnection = {
+  __typename?: 'EventConnection'
+  nodes: Array<Event>
+  pageInfo: PageInfo
+  totalCount: Scalars['Int']
+}
+
+export type EventFilter = {
+  openOnly?: Maybe<Scalars['Boolean']>
```
Itrulia, line 322:
```suggestion
  upcomingOnly?: Maybe<Scalars['Boolean']>
```

packages/api/src/graphql/event/event.ts
```diff
@@ -0,0 +1,95 @@
...
+    nodes: {type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLEvent)))},
+    pageInfo: {type: GraphQLNonNull(GraphQLPageInfo)},
+    totalCount: {type: GraphQLNonNull(GraphQLInt)}
+  }
+})
+
+export const GraphQLEventFilter = new GraphQLInputObjectType({
+  name: 'EventFilter',
+  fields: {
+    openOnly: {type: GraphQLBoolean}
```
Itrulia, line 83:
```suggestion
    upcomingOnly: {type: GraphQLBoolean}
```

packages/editor/src/client/api/index.ts
```diff
@@ -449,6 +450,44 @@ export type EmbedBlockInput = {
...
+
+export type EventConnection = {
+  __typename?: 'EventConnection';
+  nodes: Array<Event>;
+  pageInfo: PageInfo;
+  totalCount: Scalars['Int'];
+};
+
+export type EventFilter = {
+  openOnly?: Maybe<Scalars['Boolean']>;
```
Itrulia, line 474:
```suggestion
  upcomingOnly?: Maybe<Scalars['Boolean']>;
```

packages/api/src/graphql/event/event.queries.ts
```diff
@@ -0,0 +1,119 @@
...
+        startsAt: sortOrder
+      }
+  }
+}
+
+const createUpcomingOnlyFilter = (filter?: Partial<EventFilter>): Prisma.EventWhereInput => {
+  if (filter?.upcomingOnly) {
+    return {
+      startsAt: {
+        lte: new Date()
```
Itrulia, line 54:
```suggestion
        gte: new Date()
```

packages/api/prisma/schema.prisma
```diff
@@ -938,3 +942,43 @@ model PollExternalVote {
...
 }
+
+enum EventStatus {
+  Cancelled   @map("cancelled")
+  Rescheduled @map("rescheduled")
+  Postponed   @map("postponed")
+  Scheduled   @map("scheduled")
+}
+
+model Event {
```
Itrulia, line 953: sub spec of schema.org events. Location is kept as a string for simplicity

packages/api/prisma/schema.prisma
```diff
@@ -938,3 +942,43 @@ model PollExternalVote {
   @@unique([answerId, sourceId])
   @@map("polls.external-votes")
 }
+
+enum EventStatus {
+  Cancelled   @map("cancelled")
+  Rescheduled @map("rescheduled")
+  Postponed   @map("postponed")
+  Scheduled   @map("scheduled")
```
Itrulia, line 950: Leftout movedOnline from schema spec as not needed for now and easy to add

packages/api/src/graphql/comment/comment.ts
```diff
@@ -207,17 +207,18 @@ export const GraphQLComment: GraphQLObjectType<Comment, Context> = new GraphQLOb
     },
     tags: {
       type: GraphQLList(GraphQLNonNull(GraphQLTag)),
-      resolve: createProxyingResolver(async ({id}, _, {prisma: {taggedComments}}) => {
```
Itrulia, line 210: Made a bit more effecient

packages/api/src/graphql/query.public.ts
```diff
@@ -520,6 +527,32 @@ export const GraphQLPublicQuery = new GraphQLObjectType<undefined, Context>({
       },
       resolve: (root, {pollId}, {authenticateUser, prisma: {pollVote}}) =>
         userPollVote(pollId, authenticateUser, pollVote)
+    },
+
+    // Event
+    // =======
+    events: {
```
Itrulia, line 534: Public list api so it can be fully used by operators to build separate pages

packages/api/src/graphql/event/event.ts
```diff
@@ -0,0 +1,95 @@
...
+    status: {type: GraphQLNonNull(GraphQLEventStatus)},
+    description: {type: GraphQLRichText},
+
+    location: {type: GraphQLString},
+    startsAt: {type: GraphQLNonNull(GraphQLDateTime)},
+    endsAt: {type: GraphQLDateTime},
+
+    tags: {
+      type: GraphQLList(GraphQLNonNull(GraphQLTag)),
+      resolve: createProxyingResolver(async ({id}, _, {prisma: {tag}}) => {
```
michael-scheurer, line 48: Could we avoid implementing new proxying resolvers and implement includes (in Prisma terms, or joins in sql terms) instead?

packages/api/src/graphql/permissions.ts
```diff
@@ -567,6 +567,30 @@ export const CanDeletePoll: Permission = {
   deprecated: false
 }
 
+export const CanGetEvent: Permission = {
+  id: 'CAN_GET_EVENT',
+  description: 'Allows to get a event',
```
michael-scheurer, line 572:
```suggestion
  description: 'Allows to get an event',
```

packages/api/src/graphql/permissions.ts
```diff
@@ -567,6 +567,30 @@ export const CanDeletePoll: Permission = {
...
+  deprecated: false
+}
+
+export const CanCreateEvent: Permission = {
+  id: 'CAN_CREATE_EVENT',
+  description: 'Allows to create a event',
```
michael-scheurer, line 578:
```suggestion
  description: 'Allows to create an event',
```

packages/api/src/graphql/permissions.ts
```diff
@@ -567,6 +567,30 @@ export const CanDeletePoll: Permission = {
...
+  deprecated: false
+}
+
+export const CanUpdateEvent: Permission = {
+  id: 'CAN_UPDATE_EVENT',
+  description: 'Allows to update a event',
```
michael-scheurer, line 584:
```suggestion
  description: 'Allows to update an event',
```

packages/api/src/graphql/permissions.ts
```diff
@@ -567,6 +567,30 @@ export const CanDeletePoll: Permission = {
...
+  deprecated: false
+}
+
+export const CanDeleteEvent: Permission = {
+  id: 'CAN_DELETE_EVENT',
+  description: 'Allows to delete a event',
```
michael-scheurer, line 590:
```suggestion
  description: 'Allows to delete an event',
```

michael-scheurer (changes requested):
Great job. Annotations in the code. Plus these:
- In the editor, if you open the tags list of the events (make sure you have an event tag) and then switch to the tags of the comments, the data is not reloaded in the frontend.
image
- On the events list, I would like to see (1) a well formatted, human readable date and time. I could imagine, to put icons in front of date and time each. (2) The publication state. (3) A filter and search functionality.
- We should be able to place events in pages/articles. If not a requirement, we should at least create a Jira story.
- I don't really see the value of the event status. It doesn't have any effect on api side except the filters. If any media uses states, it can be achieved with tags. Event status without any effect seems to me more confusing than helping.

michael-scheurer (changes requested):
Great job. Annotations in the code. Plus these:
- In the editor, if you open the tags list of the events (make sure you have an event tag) and then switch to the tags of the comments, the data is not reloaded in the frontend.
image
- On the events list, I would like to see (1) a well formatted, human readable date and time. I could imagine, to put icons in front of date and time each. (2) The publication state. (3) A filter and search functionality.
- We should be able to place events in pages/articles. If not a requirement, we should at least create a Jira story.
- I don't really see the value of the event status. It doesn't have any effect on api side except the filters. If any media uses states, it can be achieved with tags. Event status without any effect seems to me more confusing than helping.

Itrulia:
@michael-scheurer I used the same formatting as in polls, should we change both?

The status is part of the schema.org spec and they have a meaning, it should not be done with tags as this would have to be implemented all the time in different formats. This is especially important when we import data from other sources.

packages/api/src/graphql/event/event.ts
Itrulia: That's technically an anti pattern in GraphQL, as if you do not care about tags now you do extra database calls.

penk85 (approved):

michael-scheurer:
> > In the editor, if you open the tags list of the events (make sure you have an event tag) and then switch to the tags of the comments, the data is not reloaded in the frontend.
> 
> what you currently see in the deployment in the editor is from a different PR, I will update that there and should be blocking there and not here.
> 
> > On the events list, I would like to see (1) a well formatted, human readable date and time. I could imagine, to put icons in front of date and time each. (2) The publication state. (3) A filter and search functionality.
> 
> @michael-scheurer I used the same formatting as in polls, should we change both?
> 
> The status is part of the schema.org spec and they have a meaning, it should not be done with tags as this would have to be implemented all the time in different formats. This is especially important when we import data from other sources.

Ok for the status aligned with standard specs. But still doesn't have any functional implication, which could lead to misunderstandings in newsrooms. Non-blocking for me.
Date formatting: I would change both, if you may.

michael-scheurer (approved):

PR #777, opened 2022-11-25
Replace @rsuite/icons with better solution

antkiewiczk:


docs/README.md
```diff
@@ -178,6 +178,32 @@ To give further control over the content of HTML block, whitelisting certain tag
...
 
+### Icon Library
+
+The icon library used throughout the editor is react-icons. https://react-icons.github.io/react-icons
+
+It's a great library created reasonably long time ago, with a good support, that includes few of the most widely used icon packages, all for free.
+For consistency, we decided to only use one of the icon sets in our project, and we chose material-icons from Google.
+- it contains lots of icons
+- it utilizes ES6 imports that allows us to include only the icons that we are using in our project
+- it's free
```
michael-scheurer, line 189: Does free mean open source?

docs/README.md
```diff
@@ -178,6 +178,32 @@ To give further control over the content of HTML block, whitelisting certain tag
 
 We have, however, to be aware that it's almost impossible to be 100% sure that none of the code displayed as custom HTML is dangerous.
 
+### Icon Library
```
michael-scheurer, line 181: Documentation is great!

packages/editor/src/client/api/index.ts
```diff
@@ -117,7 +117,7 @@ export type ArticleRevision = {
   canonicalUrl?: Maybe<Scalars['String']>;
   url: Scalars['String'];
   image?: Maybe<Image>;
-  authors: Array<Maybe<Author>>;
```
michael-scheurer, line 120: We shouldn't have any changes in this file, shouldn't we?

michael-scheurer (changes requested):
Thanks for implementing. Looks super good. Before approving this PR, I would like to have a look in a deployed version.

packages/editor/src/client/api/index.ts
Itrulia: The 8.1.0 release was missing the generation of these files which is the reason why

Itrulia (approved):

michael-scheurer (approved):

PR #779, opened 2022-12-01
feat(editor/api): implemented the event block

Itrulia:
Similar to the comment block

Also added filter for tags

### DIFF
--- a/packages/api/__tests__/specs/event.ts
+++ b/packages/api/__tests__/specs/event.ts
@@ -219,7 +219,7 @@ describe('Events', () => {
       expect(res.data.updateEvent.tags).toHaveLength(3)
     })
 
-    test.only('can not be updated with endsAt before startsAt', async () => {
+    test('can not be updated with endsAt before startsAt', async () => {
       const {mutate} = testClientPrivate
 
       const createRes = await mutate({
--- a/packages/api/__tests__/specs/event.ts
+++ b/packages/api/__tests__/specs/event.ts
@@ -93,6 +93,21 @@ describe('Events', () => {
         }
       })
     })
+
+    test('can not be created with endsAt before startsAt', async () => {
+      const {mutate} = testClientPrivate
+
+      const res = await mutate({
+        mutation: CreateEvent,
+        variables: {
+          name: 'Test Event',
+          startsAt: '2020-11-25T20:00:00.000Z',
+          endsAt: '2020-11-24T20:00:00.000Z'
+        }
+      })
+
+      expect(res).toMatchSnapshot()
+    })
   })
 
   describe('update', () => {
@@ -202,6 +217,38 @@ describe('Events', () => {
       })
 
       expect(res.data.updateEvent.tags).toHaveLength(3)
+    })
+
+    test.only('can not be updated with endsAt before startsAt', async () => {
+      const {mutate} = testClientPrivate
+
+      const createRes = await mutate({
+        mutation: CreateEvent,
+        variables: {
+          name: 'Test Event',
+          startsAt: '2020-11-25T20:00:00.000Z'
+        }
+      })
+
+      const res = await mutate({
+        mutation: UpdateEvent,
+        variables: {
+          id: createRes.data.createEvent.id,
+          endsAt: '2020-11-24T20:00:00.000Z'
+        }
+      })
+
+      const res2 = await mutate({
+        mutation: UpdateEvent,
+        variables: {
+          id: createRes.data.createEvent.id,
+          startsAt: '2020-11-27T20:00:00.000Z',
+          endsAt: '2020-11-26T20:00:00.000Z'
+        }
+      })
+
+      expect(res).toMatchSnapshot()
+      expect(res2).toMatchSnapshot()
     })
   })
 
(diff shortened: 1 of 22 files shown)


packages/editor/src/client/blocks/eventBlock.tsx
```diff
@@ -0,0 +1,83 @@
+import PencilIcon from '@rsuite/icons/legacy/Pencil'
```
Itrulia, line 1: Copied from CommentBlock and adapted for EventBlock

packages/editor/src/client/panel/selectEventsPanel.tsx
```diff
@@ -0,0 +1,199 @@
+import {ApolloError} from '@apollo/client'
```
Itrulia, line 1: Copied from CommentBlock and adapted for EventBlock

packages/editor/src/client/atoms/tag/selectTags.tsx
```diff
@@ -65,7 +65,10 @@ export function SelectTags({name, tagType, selectedTags, setSelectedTags}: Selec
       name={name}
       value={selectedTags}
       data={availableTags}
-      onChange={(value: string[]) => setSelectedTags(value)}
+      onChange={(value: string[]) => {
```
Itrulia, line 68: Fixes a bug that you can type in whatever and press enter and you create a new item. That's a bug in the comments currently too.

creatable={false} does not work for w/e reason.

packages/editor/src/client/panel/selectEventsPanel.tsx
```diff
@@ -0,0 +1,199 @@
...
+                      }
+                    />
+                  </div>
+                )}
+              </Table.Cell>
+            </Table.Column>
+          )}
+
+          <Table.Column width={250} resizable>
+            <Table.HeaderCell>{t('blocks.event.name')}</Table.HeaderCell>
```
penk85, line 160: This translation isn't displaying

packages/editor/src/client/panel/selectEventsPanel.tsx
```diff
@@ -0,0 +1,199 @@
...
+            <Table.Cell>{(rowData: EventRefFragment) => rowData.name}</Table.Cell>
+          </Table.Column>
+
+          <Table.Column width={150} align="center" fixed="right">
+            <Table.HeaderCell>{t('events.overview.edit')}</Table.HeaderCell>
```
penk85, line 165: Also this one

penk85 (changes requested):
Looks really great! Small fix with translations I left a comment about but otherwise looks really good.

packages/api/__tests__/specs/event.ts
```diff
@@ -284,5 +331,103 @@ describe('Events', () => {
 
       expect(res.data.events.totalCount).toBeGreaterThanOrEqual(2)
     })
+
+    test('can query a list of events with a specific tag', async () => {
```
michael-scheurer, line 335: like all these tests!

packages/editor/src/client/blocks/eventBlock.tsx
```diff
@@ -0,0 +1,83 @@
...
+                  right: 0
+                }}>
+                <IconButton size={'lg'} icon={<PencilIcon />} onClick={() => setIsDialogOpen(true)}>
+                  {t('blocks.event.edit')}
+                </IconButton>
+              </div>
+
+              <p>
+                {t('blocks.event.events', {
+                  count: events.length ? events.length : filter.events?.length ?? 0
```
michael-scheurer, line 63: Whenever events have been selected, it only shows "n events" to the user. A list (eventually limited to 10) with the events would increase the UX a lot imo. The list could contain event image, title, location and start/end date e.g.

packages/editor/src/client/blocks/eventBlock.tsx
michael-scheurer: Could we re-use the code with a more generic and re-usable component, instead of copying code? I always panic, when I read "copy code" =)

packages/editor/src/client/panel/selectEventsPanel.tsx
michael-scheurer: idem. copy code is evil ;)

michael-scheurer (commented):
Great job! See my comments.

packages/api/__tests__/specs/event.ts
```diff
@@ -203,6 +218,38 @@ describe('Events', () => {
 
       expect(res.data.updateEvent.tags).toHaveLength(3)
     })
+
+    test.only('can not be updated with endsAt before startsAt', async () => {
```
Itrulia, line 222:
```suggestion
    test('can not be updated with endsAt before startsAt', async () => {
```

packages/editor/src/client/panel/selectEventsPanel.tsx
Itrulia: Tbf this one probably should be copied as you said we should show a proper preview in the future (same for comments). The other one I see what I can do D:

tomaszdurka (approved):

Itrulia:
Will open a new PR For the improvements of the editor preview @michael-scheurer  :)

PR #800, opened 2022-12-29
feat(editor): improve event preview

Itrulia:
image


packages/editor/src/client/blocks/types.ts
```diff
@@ -352,7 +352,10 @@ export function unionMapForBlock(block: BlockValue): BlockInput {
     case BlockType.EventBlock:
       return {
         event: {
-          filter: block.value?.filter ?? {}
+          filter: {
```
Itrulia, line 355: fixes bug with __typename when setting things again

PR #801, opened 2022-12-30
feat(api): new phrase API to do full text search on articles and pages

Itrulia:


libs/api/prisma/migrations/20221229154844_unwrap_array_of_jsonb/migration.sql
```diff
@@ -0,0 +1,84 @@
+-- Migrate article revision blocks from jsob[] to jsonb
```
Itrulia, line 1: Reason for this was that in the database it was an array of json objects instead of a JSON array.
It's unnecessary for it to be this way and it makes it harder to work with in SQL.

libs/api/prisma/migrations/20221229154844_unwrap_array_of_jsonb/migration.sql
```diff
@@ -0,0 +1,84 @@
...
+SET blocks2 = jsonBlocks
+FROM (
+    SELECT jsonb_agg(j::jsonb)
+    FROM (
+        SELECT unnest(blocks)
+        FROM "articles.revisions"
+    ) as x(j)
+) as x(jsonBlocks);
+
+ALTER TABLE "articles.revisions" ALTER COLUMN "blocks2" SET NOT NULL;
```
Itrulia, line 14: This is done on purpose before the deletion of the column in case this fails

penk85 (approved):

libs/api/prisma/migrations/20221229154844_unwrap_array_of_jsonb/migration.sql
michael-scheurer, line 1:
```suggestion
-- Migrate article revision blocks from json[] to jsonb
```

libs/api/src/lib/graphql/phrase/phrase.public-queries.ts
```diff
@@ -0,0 +1,103 @@
...
+import {Context} from '../../context'
+
+export const queryPhrase = async (
+  query: string,
+  prisma: PrismaClient,
+  publicArticlesLoader: Context['loaders']['publicArticles'],
+  publicPagesLoader: Context['loaders']['publicPagesByID']
+) => {
+  const [articleRevisions, pageRevisions] = await Promise.all([
+    prisma.$queryRaw<{id: string}[]>`
```
michael-scheurer, line 13: This is very expensive, if a lot of revisions exist. Couldn't we just search in the latest revision?

michael-scheurer (approved):

michael-scheurer (approved):

michael-scheurer (approved):

libs/api/src/lib/graphql/phrase/phrase.public-queries.ts
Itrulia: I tested it with 100k revisions and was fine due to the index. The problem with only searching for published revisions is that I need to join another table which was slower in my testing :(

michael-scheurer: Very interesting!

PR #802, opened 2023-01-04
chore: replace yarn with npm

Itrulia:


.env.local
```diff
@@ -1,56 +0,0 @@
-# Editor
```
Itrulia, line 1: Shouldn't have been commited

docs/README.md
```diff
@@ -188,4 +188,7 @@ Within this event another invoice event will be triggered which is handled in `e
 To perform migrations on the database two steps are necessary.
 
 1. add an object in migrations.ts.
-2. restart the API (yarn watch)
+   <<<<<<< HEAD
```
Itrulia, line 191: Merge conflict

docs/README.md
```diff
@@ -188,4 +188,7 @@ Within this event another invoice event will be triggered which is handled in `e
...
+2. # restart the API (npm run watch)
+3. restart the API (yarn watch)
+   > > > > > > > master
```
tomaszdurka, line 194: should this be here?

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "8.1.1",
+  "version": "10.0.0-alpha.1",
```
tomaszdurka, line 2: oh what's up with this?

Itrulia: That's from my testing :D

libs/api/__tests__/specs/peer.test.ts
```diff
@@ -150,7 +150,6 @@ describe('Peers', () => {
     test('can update peer profile', async () => {
       const input: PeerProfileInput = {
         name: 'test peer profile',
-        logoID: 'logoID123',
```
Itrulia, line 153: This was causing an error and the snapshot included it. We don't have any image uploads in tests so far so I deleted it for now.

.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
Itrulia, line 1: Without it we get an error because of @karma.run requiring React 16 in peerDeps

tomaszdurka (changes requested):
I just found this one yarn in the project still.

```
https://github.com/wepublish/wepublish/blob/f/WPC-963-remove-yarn/config/create-gke-manifest.js
```

Apart from that looks great !

Itrulia:
Good catch @tomaszdurka, updated

tomaszdurka (dismissed):

helm/charts/wepublish/templates/deployment-api.yaml
```diff
@@ -32,7 +32,7 @@ spec:
           image: {{ .Values.deployment.image }}
           imagePullPolicy: {{ .Values.deployment.imagePullPolicy }}
           command: ['/bin/sh']
-          args: ['-c', 'yarn prisma migrate deploy && npx ts-node libs/api/prisma/run-seed.ts && node dist/libs/api/src/runServer.js']
+          args: ['-c', 'npx prisma migrate deploy && npx ts-node libs/api/prisma/run-seed.ts && node dist/libs/api/src/runServer.js']
```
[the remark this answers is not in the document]
Itrulia: No, npx is for binaries such as prisma. npm run is for scripts inside the package.json

PR #804, opened 2023-01-04
B/WPC-981-CancelationBugFixPayrexxSubscription

elias-summermatter:


Itrulia (approved):
I guess we should patch this based on 8.1.1 and not the Nx changes so we can have a 8.1.2 release?

michael-scheurer (approved):

PR #818, opened 2023-01-16
fix/wpc-985 fix bug on subscription period creation

michael-scheurer:


Itrulia (approved):
Shouldn't this cause the same issue of creating the period multiple times?

michael-scheurer:
> Shouldn't this cause the same issue of creating the period multiple times?

nope. and tested today on production. works correct so far. the problem was, when updating via subscription model, the update hook of subscription threw an error, which canceled the creation of the period and therefore more and more invoices were created.

elias-summermatter (approved):

PR #819, opened 2023-01-16
dashboard

penk85:
WPC-974: editor dashboard. Currently has newsfeed and chart for count of subscribers over past year. This is initial set up and future plans include more detailed event logging for feed as well as the display of other stats.


apps/editor/src/app/atoms/dashboard/subscriberChart.tsx
```diff
@@ -0,0 +1,50 @@
+import React, {useMemo} from 'react'
+import {useTranslation} from 'react-i18next'
+import {BarChart, Bar, XAxis, YAxis, ResponsiveContainer} from 'recharts'
```
Itrulia, line 3: Love recharts <3

penk85: I think you mention you just added it as well?

apps/editor/src/app/locales/en.json
```diff
@@ -1316,6 +1317,27 @@
...
+      "newPage": "New page created",
+      "newComment": "New comment added",
+      "newPoll": "New poll opened",
+      "newSubscription": "New subscription",
+      "newAuthor": "New author added",
+      "newUser": "New user added",
+      "newSubscribers": "New subscribers over past year",
+      "createdBy": " by {{creator}}",
+      "subscriberChart": {
+        "fillColor": "#3498ff"
```
penk85, line 1339: Color should not be in translation file, but adding it any other way as a string was creating errors.

Itrulia: You can add it to the .eslintrc file as an exception to the rule. Shouldn't be in translations I agree.
Else you can also exclude the rule for that line.

apps/editor/src/app/atoms/dashboard/subscriberChart.tsx
Itrulia: Yeah :D

apps/editor/src/app/locales/en.json
penk85: Thanks I'll do that.

libs/api/__tests__/api/private/index.ts
```diff
@@ -23,6 +23,25 @@ export type Scalars = {
...
+  __typename?: 'Action';
+  actionType: ActionType;
+  creator?: Maybe<Scalars['String']>;
+  date: Scalars['DateTime'];
+  id: Scalars['ID'];
+  summary?: Maybe<Scalars['String']>;
+};
+
+export enum ActionType {
+  Article = 'ARTICLE',
```
michael-scheurer, line 36: At leas events are missing. Others too?

libs/api/src/lib/graphql/subscription/subscription.private-queries.ts
```diff
@@ -70,3 +71,163 @@ export const getSubscriptionsAsCSV = async (
...
 }
+
+export const getNewSubscribersYear = async (
+  authenticate: Context['authenticate'],
+  subscription: PrismaClient['subscription']
+) => {
+  const {roles} = authenticate()
+  authorise(CanGetSubscriptions, roles)
+  return (
+    await Promise.all([
```
michael-scheurer, line 82: Can you find a cleaner way to cover the same functionality? Because it's too much code duplication imo.

michael-scheurer (changes requested):
Hey Penina
See my comments in the code. Besides that, I have two annotations regarding the UI:
- May you center the horizontal alignement?
image

- Justify the titles on the left side, because.
image

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,123 @@
...
+    <Table wordWrap autoHeight rowHeight={60} data={actions} loading={isLoading}>
+      <Column flexGrow={2}>
+        <HeaderCell dataKey="event">{t('dashboard.event')}</HeaderCell>
+        <Cell dataKey="event">
+          {(rowData: RowDataType<Action>) => {
+            let icon = <></>
+            let path = '/'
+            let type = ''
+
+            switch (rowData.actionType) {
```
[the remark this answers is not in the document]
penk85: Events got merged right before I finished up this PR so I didn't add it, but can still do. This doesn't include all activities and the plan is to refactor to be action based with a table of actions where we can have more details such as who created article and separate actions for updating and publishing.

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,123 @@
...
+                break
+              default:
+                icon = <></>
+            }
+            return (
+              <>
+                <ActivityFeedIcon circle>{icon}</ActivityFeedIcon>
+                <Link to={path}> {type} </Link>
+                {formatDistanceToNow(new Date(rowData.date), {
+                  locale: i18n.language === 'de' ? de : i18n.language === 'fr' ? fr : enUS,
```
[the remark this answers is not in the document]
penk85: Good point.

libs/api/src/lib/graphql/query.private.ts
```diff
@@ -748,6 +752,27 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
         id: {type: GraphQLID}
       },
       resolve: (root, {id}, {prisma: {event}}) => getEvent(id, event)
+    },
+
+    // Stats
+    newSubscribersPastYear: {
```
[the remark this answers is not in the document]
penk85: I like this idea. :+1:

apps/editor/src/app/api/v2/index.ts
```diff
@@ -36,10 +36,8 @@ export type VersionInformationQuery = {
 
 export const VersionInformationDocument = gql`
```
Itrulia, line 37: Hm... No clue what is happening here and why it was generated like this

libs/api/src/lib/graphql/subscription/subscription.private-queries.ts
Itrulia: This could be done with 1 query and then counting on the server, would be less code duplication and faster as multiple DB calls are slower than manual computation generally.

libs/api/src/lib/graphql/query.private.ts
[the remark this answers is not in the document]
Itrulia: I would not suggest to do this because this would be how REST APIs would be designed but not GraphQL.
GraphQL's intention was so that the client decides what it wants and it does not have to be integrated into the server.

If we have multiple endpoints one can do:

```graphql
query Stats($take: Int, $skip: Int) {
  newSubscribersPastYear() {
     ....
  }

  newArticlesPastYear() {
    ....
  }

  newPagesPastYear() {
    ....
  }
}
```

Which would give you everything as 1 call.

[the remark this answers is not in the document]
michael-scheurer: actually, I can agree with that.

apps/editor/src/app/routes/dashboard.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const {data: me} = useMeQuery()
+  const name = useMemo(
+    () => me?.me?.preferredName ?? me?.me?.firstName ?? me?.me?.name ?? 'User',
+    [me]
+  )
+
+  return (
+    <>
+      <h2>{t('dashboard.dashboard')}</h2>
```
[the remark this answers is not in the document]
penk85: :facepalm:

michael-scheurer (changes requested):
Hey Penina
Only a few alignments and visual details, but important, because the dashboard is the first impression a user get's from We.Publish.

- You may align the activity feed panel with the headline
- The chart panel is shaded, the activity feed isn't. I think, neither of them should be bordered, because no other panel is shaded in the entire application.
- The scrollbar in the middle of the dashboard doesn't look that nice. Either style it a bit or extend the height and remove the scrollbar.

image

libs/api/src/lib/graphql/user/user.private-queries.ts
```diff
@@ -14,14 +14,7 @@ export const getMe = (authenticate: Context['authenticate']) => {
...
 
-export const getUserById = (
-  id: string,
-  authenticate: Context['authenticate'],
-  user: PrismaClient['user']
-) => {
-  const {roles} = authenticate()
-  authorise(CanGetUser, roles)
-
+export const getUserById = (id: string, user: PrismaClient['user']) => {
```
Itrulia, line 17: should probably not be in this file anymore as it's not private anymore

libs/api/src/lib/graphql/query.private.ts
```diff
@@ -259,7 +266,12 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
     user: {
       type: GraphQLUser,
       args: {id: {type: GraphQLID}},
-      resolve: (root, {id}, {authenticate, prisma: {user}}) => getUserById(id, authenticate, user)
+      resolve: (root, {id}, {authenticate, prisma: {user}}) => {
+        const {roles} = authenticate()
```
Itrulia, line 270: Can we move this out to a function again please!

libs/api/src/lib/graphql/query.private.ts
```diff
@@ -281,8 +293,11 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
     subscription: {
       type: GraphQLSubscription,
       args: {id: {type: GraphQLNonNull(GraphQLID)}},
-      resolve: (root, {id}, {authenticate, prisma: {subscription}}) =>
-        getSubscriptionById(id, authenticate, subscription)
+      resolve: (root, {id}, {authenticate, prisma: {subscription}}) => {
+        const {roles} = authenticate()
```
Itrulia, line 297: Same here

libs/api/src/lib/graphql/query.private.ts
```diff
@@ -762,6 +780,38 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
       },
       resolve: (root, {id}, {prisma: {event}}) => getEvent(id, event)
+    },
+
+    // Stats
+    newSubscribersPerMonth: {
+      type: GraphQLList(GraphQLSubscribersPerMonth),
+      args: {monthsBack: {type: GraphQLInt}},
+      resolve: async (root, {monthsBack}, {authenticate, prisma: {subscription}}) => {
+        return await getNewSubscribersPerMonth(authenticate, subscription, monthsBack)
```
Itrulia, line 790: No return or async/await is needed here

libs/api/src/lib/graphql/query.private.ts
```diff
@@ -762,6 +780,38 @@ export const GraphQLQuery = new GraphQLObjectType<undefined, Context>({
...
+
+    // Actions
+    // =======
+    actions: {
+      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLAction))),
+      resolve: async (
+        root,
+        _,
+        {authenticate, prisma: {article, page, comment, subscription, author, poll, user, event}}
+      ) => {
```
Itrulia, line 802: No return or async is needed here

libs/api/src/lib/graphql/subscription/subscription.private-queries.ts
```diff
@@ -71,3 +65,42 @@ export const getSubscriptionsAsCSV = async (
...
+      return (
+        subsc.startsAt > startOfMonth(subMonths(new Date(), i)) &&
+        subsc.startsAt < lastDayOfMonth(subMonths(new Date(), i))
+      )
+    }).length
+    const month = new Date()
+    month.setMonth(month.getMonth() - i)
+    res.push({month: format(month, 'MMM-yy'), subscriberCount: count})
+  }
+  return res.reverse()
```
Itrulia, line 105: Instead of reversing here, can we start the loop in reverse?

Itrulia (changes requested):
See comments

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,303 @@
...
+  padding: 4px;
+`
+
+type Action = NonNullable<RecentActionsQuery['actions']>[number]
+
+export function ActivityFeed() {
+  const {data, error} = useRecentActionsQuery({fetchPolicy: 'no-cache'})
+  const [actions, setActions] = useState<Action[] | undefined>([])
+
+  useEffect(() => {
```
Itrulia, line 65: instead of an useEffect or useState we can do `const actions = data?.actions`

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,303 @@
...
+  key: number
+  action: Action
+}
+
+function TimelineItemContainer(props: TimelineItemContainerProps) {
+  const {action, key} = props
+  const {t} = useTranslation()
+
+  switch (action.__typename) {
+    case 'ArticleCreatedAction':
```
Itrulia, line 109: Are there any enums we can use?

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,303 @@
...
+          summary={
+            <TranslationWithLink
+              translationKey={t('dashboard.newComment')}
+              link={`/comments/edit/${action.comment.id}`}
+            />
+          }
+          date={action.date}
+          details={
+            <>
+              {`${action.comment.user?.name ?? action.comment.guestUsername ?? ''} ${
```
Itrulia, line 157: Can we extract this into a function?

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,303 @@
...
+                action.comment.revisions[action.comment.revisions.length - 1]?.title
```
Itrulia, line 158: Same here

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,303 @@
...
+            <TranslationWithLink
+              translationKey={t('dashboard.newAuthor')}
+              link={`/authors/edit/${action.author.id}`}
+            />
+          }
+          date={action.date}
+          details={
+            <>
+              {action.author.name}
+              {action.author.jobTitle ? ', ' + action.author.jobTitle : ''}
```
Itrulia, line 228:
```suggestion
              {action.author.jobTitle  && `, ${action.author.jobTitle}`}
```

apps/editor/src/app/atoms/dashboard/activityFeed.tsx
```diff
@@ -0,0 +1,303 @@
...
+          details={
+            <>
+              {action.event.name}
+              {action.event.location}
+            </>
+          }
+        />
+      )
+    default:
+      return <></>
```
Itrulia, line 269:
```suggestion
      return null
```

apps/editor/src/app/atoms/dashboard/subscriberChart.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <ResponsiveContainer width="100%" height={300}>
+      {hasSubscriptions && !loading ? (
+        <BarChart height={300} data={pastYearSubscriptions} margin={{left: -40}}>
+          <XAxis minTickGap={0} fontSize={10} width={10} dataKey="month" />
+          <YAxis allowDecimals={false} />
+          <Bar radius={[5, 5, 0, 0]} dataKey="subscriberCount" fill="#3498ff" />
+        </BarChart>
+      ) : (
+        <BarChartErrorText>{t('dashboard.noData')}</BarChartErrorText>
```
Itrulia, line 37: This will be displayed while the data is loading, is this intended?

apps/editor/src/app/atoms/dashboard/subscriberChart.tsx
```diff
@@ -0,0 +1,41 @@
...
+  font-style: italic;
+  text-align: center;
+  margin-top: 2rem;
+`
+
+export function SubscriberChart() {
+  const {t} = useTranslation()
+  const {data, loading} = useNewSubscribersPerMonthQuery({variables: {months: 12}})
+
+  const pastYearSubscriptions = useMemo(() => data?.newSubscribersPerMonth ?? [], [data])
```
Itrulia, line 19:
```suggestion
  const pastYearSubscriptions = data?.newSubscribersPerMonth ?? []
```

Itrulia: no useMemo needed as the CPU impact is super minimal but the memory impact is a lot bigger with useMemo

apps/editor/src/app/base.tsx
```diff
@@ -41,6 +42,7 @@ import {
 
 import {PermissionControl} from './atoms/permissionControl'
 import Version from './atoms/version'
+import {fr, de, enUS} from 'date-fns/locale'
```
Itrulia, line 45: Why are these called locales :( Only enUS is a locale here.

apps/editor/src/app/routes/dashboard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  max-height: 100%;
+`
+
+export function Dashboard() {
+  const {t} = useTranslation()
+
+  const {data: me} = useMeQuery()
+  const name = useMemo(
+    () => me?.me?.preferredName ?? me?.me?.firstName ?? me?.me?.name ?? t('dashboard.user'),
+    [me]
```
Itrulia, line 27: t is also needed in here technically.

But we can go without useMemo here as the CPU impact isn't big enough to save it in the memory.

libs/api/src/lib/graphql/action/action.private-queries.ts
```diff
@@ -0,0 +1,132 @@
...
+  comment: PrismaClient['comment'],
+  subscription: PrismaClient['subscription'],
+  author: PrismaClient['author'],
+  poll: PrismaClient['poll'],
+  user: PrismaClient['user'],
+  event: PrismaClient['event']
+): Promise<Array<Action>> => {
+  const {roles} = authenticate()
+
+  const articles = hasPermission(CanGetArticles, roles)
```
Itrulia, line 39: Currently all of these are loaded after each other while we can load them all at the same time.

apps/editor/src/app/base.tsx
penk85: Currently using enUS for the translations so it makes sense to match that, no?

apps/editor/src/app/routes/dashboard.tsx
penk85: Not sure what you mean with t?

libs/api/src/lib/graphql/action/action.private-queries.ts
penk85: I hope this works better now? Although not sure how much is necessary to optimize since anyhow this will be changed after actions table is added.

apps/editor/src/app/base.tsx
Itrulia: Oh we do? Then nvm, thats another topic to switch to enGB

Itrulia (dismissed):

Itrulia (approved):

PR #810, opened 2023-01-09
Implement nestjs mvp

tomaszdurka:
- Add nestjs support
- Introduce nestjs application into api-example

PR #820, opened 2023-01-17
refactor(editor): split off api definitions into seperate libraries

Itrulia:
Allows to re-use it in different libraries in the future such as the membership/editor library


libs/editor/api-v2/README.md
```diff
@@ -0,0 +1,7 @@
+# editor-api-v2
```
michael-scheurer, line 1: what is the intention of api v2?

libs/editor/api/src/lib/graphql.ts
```diff
@@ -1,8 +1,7 @@
 // THIS FILE IS AUTOGENERATED, EDIT WITH CAUTION
+import {Node} from 'slate'
```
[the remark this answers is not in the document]
Itrulia: Yeah it's because it wasn't in the prettier ignore

libs/editor/api-v2/README.md
Itrulia: Comes from https://github.com/wepublish/wepublish/pull/810, it's the version we are working on right now to use nestjs.

michael-scheurer (dismissed):

apps/editor/src/app/routes/articleList.tsx
```diff
@@ -17,7 +10,40 @@ import {
   useDeleteArticleMutation,
   useDuplicateArticleMutation,
   useUnpublishArticleMutation
-} from '../api'
+} from '@wepublish/editor/api'
+import {useEffect, useEffect, useMemo, useMemo, useState, useState} from 'react'
```
Itrulia, line 14:
```suggestion
import {useEffect, useMemo, useMemo, useState, useState} from 'react'
```

PR #823, opened 2023-01-18
revert(editor): bring back comment context

Itrulia:


michael-scheurer (changes requested):
Found some bugs.

- I added some tags. Now they are looking like this
image

and in a comment view like this
image

- When I link a user to a comment, it is not saving correctly or not showing up the saved user in the dropdown.
image

Itrulia:
@michael-scheurer the tags are correct as they were empty. When I add a name they correctly show up.
The user drop-down is bugged, the value is correctly saved. Will merge this.

Itrulia:
fix for user dropdown will come in a seperate PR

PR #829, opened 2023-01-20
feat(api): membership dashboard api for nestjs

Itrulia:


apps/api-example/src/app/index.ts
```diff
@@ -292,85 +289,4 @@ export async function runServer(app?: Application | undefined) {
     },
     app
   )
-
```
Itrulia, line 295: Deleted it so i could turn on strict mode. Was not used anywhere

apps/editor/src/app/panel/memberPlanEditPanel.tsx
```diff
@@ -240,9 +240,9 @@ function MemberPlanEditPanel({id, onClose, onSave}: MemberPlanEditPanelProps) {
       </Drawer.Header>
       <Drawer.Body>
         <Panel>
-          <Form.Group controlId="memberPlanName">
```
Itrulia, line 243: Fixes membership creation. Already fixed on master but needed it for testing things out

apps/editor/src/app/panel/subscriptionEditPanel.tsx
```diff
@@ -240,7 +241,7 @@ function SubscriptionEditPanel({id, onClose, onSave}: SubscriptionEditPanelProps
     startsAt: startsAt.toISOString(),
     paidUntil: paidUntil ? paidUntil.toISOString() : null,
     properties,
-    deactivation
+    deactivation: deactivation ? stripTypename(deactivation) : deactivation
```
Itrulia, line 244: Fixes saving a subscription that is deactivated. Currently on master it throws an error

apps/editor/tsconfig.json
```diff
@@ -4,7 +4,6 @@
     "allowJs": false,
     "esModuleInterop": false,
     "allowSyntheticDefaultImports": true,
-    "strict": true,
```
Itrulia, line 7: now done in the tsconfig.base.json

Itrulia:
Added you all here so you can comment on the new way to write graphql queries with nestjs.

1 thing still missing from nestjs implementation is authentication @tomaszdurka, something you wanna work on or should I?

libs/membership/api/src/lib/dashboard/dashboard.module.ts
```diff
@@ -0,0 +1,17 @@
+import {Module} from '@nestjs/common'
+import {PrismaModule} from '@wepublish/nest-modules'
+import {DashboardInvoiceResolver} from './invoices/dashboard-invoice.resolver'
+import {DashboardInvoiceService} from './invoices/dashboard-invoice.service'
+import {DashboardSubscriptionResolver} from './subscriptions/dashboard-subscription.resolver'
+import {DashboardSubscriptionService} from './subscriptions/dashboard-subscription.service'
+
```
michael-scheurer, line 7: The path of this file is endless. Could we simplify? We have now lib in lib folder.

libs/membership/api/src/lib/dashboard/invoices/dashboard-invoice.model.ts
```diff
@@ -0,0 +1,16 @@
+import {Field, Int, ObjectType} from '@nestjs/graphql'
+
```
michael-scheurer, line 2: Even longer path. How about libs/membership/models/dashboard-invoice.model.ts?

libs/membership/api/src/lib/dashboard/invoices/dashboard-invoice.service.ts
```diff
@@ -0,0 +1,102 @@
+import {Injectable} from '@nestjs/common'
+import {PrismaClient} from '@prisma/client'
+import {DashboardInvoice} from './dashboard-invoice.model'
+import {ok} from 'assert'
+
+@Injectable()
+export class DashboardInvoiceService {
```
michael-scheurer, line 7: Very much like, that we use models and services along with classes. But same problem here: Can we simplify the path?

apps/editor/src/app/panel/subscriptionEditPanel.tsx
Itrulia: Just didn't get around it yet :D

libs/membership/api/src/lib/dashboard/dashboard.module.ts
Itrulia: Paths shouldnt matter for us. A clear path is better and recommended by nest itself.

libs/membership/api/src/lib/dashboard/invoices/dashboard-invoice.model.ts
Itrulia: That reduces cohesivenees. To change the query you would then need to change multiple directories.

penk85 (approved):

libs/membership/api/setup-database.js
```diff
@@ -0,0 +1,9 @@
+import execa from 'execa'
+import {name} from './jest.config'
+
+export default async () => {
+  const databaseUrl = `postgresql://postgres@localhost:5432/${name}?schema=public`
+  process.env.DATABASE_URL = databaseUrl
```
tomaszdurka, line 6: Shouldn't this be take from the process.env ? :o

Itrulia: This is a test database :D

PR #832, opened 2023-01-24
WPC-788: Subscription edit as separate view

antkiewiczk:


tomaszdurka:
@antkiewiczk is this part of specific task ?

Itrulia (dismissed):

apps/editor/src/app/routes/userEditView.tsx
```diff
@@ -697,7 +697,7 @@ function UserEditView() {
             <Col xs={12}>
               <Grid fluid>
                 <RPanel bordered header={t('userCreateOrEditView.subscriptionsHeader')}>
-                  <UserSubscriptionsList subscriptions={user?.subscriptions} />
+                  <UserSubscriptionsList subscriptions={user?.subscriptions} userId={user?.id} />
```
tomaszdurka, line 700: Render whole Subscription Column only if the user exists. Get rid of `user?.*`

apps/editor/src/app/routes/subscriptionList.tsx
```diff
@@ -301,7 +323,7 @@ function SubscriptionList() {
           setEditModalOpen(false)
           navigate('/subscriptions')
         }}>
-        <SubscriptionEditPanel
+        <SubscriptionEditView
```
michael-scheurer, line 326: This drawer can be removed. Instead you need to route the user.

michael-scheurer (changes requested):
Good work! See one comment in the code and the screenshot below. Also you might align the deactivate button. I would move it to the top right part. See for example how we did it on the comment single view.

image

Itrulia (approved):

penk85 (approved):

michael-scheurer (approved):

PR #836, opened 2023-01-26
Mail template APIs

LukasSkywalker:
Replaced the deprecated mandrill-api library with the new official Mailchimp library, added tests for all MailProvider methods.


michael-scheurer (changes requested):
Great work so far. See my comments in the code. Imo it is missing the following:
- fix linting (CI)
- fix tests (CI)
- the sync logic
- make the mailtemplates storable and connect it with the database schema of https://github.com/wepublish/wepublish/pull/822
- add private graphql interface, so we can get the data in the editor UI

libs/api/__tests__/specs/mailchimpMailProvider.test.ts
```diff
@@ -1,32 +1,60 @@
...
-
-jest.mock('../../src/lib/server', () => {
-  const originalModule = jest.requireActual('../../src/lib/server')
-  return {
-    __esModule: true,
-    ...originalModule,
-    logger: jest.fn(() => ({error: jest.fn()}))
-  }
-})
+import nock from 'nock'
```
Itrulia, line 1: We already have supertest, could we use that instead or does supertest not work?

PR #837, opened 2023-01-26
add onDelete relation to session

antkiewiczk:


libs/api/prisma/migrations/20230126155832_session_on_delete_cascade/migration.sql
```diff
@@ -0,0 +1,56 @@
+-- DropForeignKey
+ALTER TABLE "articles.revisions.author" DROP CONSTRAINT "articles.revisions.author_authorId_fkey";
```
Itrulia, line 2: Theres a lot going on here that shouldnt be here. 

Have you had by chance the GUID branch checked out once and ran migrations there?

Itrulia (changes requested):

Itrulia (approved):

tomaszdurka (approved):

PR #840, opened 2023-01-27
Fix comments query

antkiewiczk:


apps/editor/src/app/atoms/modelTitle.tsx
```diff
@@ -86,11 +100,18 @@ export function ModelTitle({
...
       <FlexboxGrid.Item colspan={12}>
         <FlexboxGrid align="middle">
-          <Link to={closePath}>
-            <ChevronLeft />
-          </Link>
-
-          <h1>{titleView()}</h1>
+          <FlexRow>
+            <PaddedCol xs={3}>
+              <h1>
```
Itrulia, line 105: We shouldn't use h1 for this as it's not a title, we can have both {titleView} and the chevron in it though

Itrulia: H1 would be semantically correct though, so we should change the size then.

Itrulia (approved):

PR #845, opened 2023-01-31
feat(890): settings - warnings, tooltips, hide from non-admin

antkiewiczk:


apps/editor/src/app/base.tsx
```diff
@@ -478,8 +478,7 @@ export function Base({children}: BaseProps) {
                     </PermissionControl>
                   </Nav.Menu>
                 </PermissionControl>
-                <PermissionControl
-                  qualifyingPermissions={['CAN_GET_SETTINGS', 'CAN_UPDATE_SETTINGS']}>
+                <PermissionControl qualifyingPermissions={['CAN_UPDATE_SETTINGS']}>
```
Itrulia, line 481: Is that on purpose that the CAN_GET_SETTINGS isn't needed anymore?

Itrulia: Then we should remove CanGetSettings from the editor's permission instead. There's other places (like the actual component) where that permission is also used.

Itrulia (changes requested):
CanGetSettings should be removed from the editor's permission instead

michael-scheurer (changes requested):
Great work. Just one detail. Could we improve the look of the info icon a little bit? I would consider to display the info icon not surrounded by a button.
Thus, the icon as it is, but in blue and remove the outer button boundary. What do you think?

libs/api/src/lib/graphql/permissions.ts
```diff
@@ -708,7 +708,6 @@ export const EditorPermissions: Permission[] = [
   CanGetPeers,
   CanGetPeerProfile,
   CanLoginEditor,
-  CanGetSettings,
```
penk85, line 711: I think we should still keep this permission in the api. Just remove the qualifying permission of "CAN_GET_SETTINGS" from the settings page .

penk85 (changes requested):
Looks great. Just I'd keep the permission in the api and update in the editor. Both the settingList and the settings tab in the base sidebar.

libs/api/src/lib/graphql/permissions.ts
penk85: Ah you're right. This makes sense.

penk85 (approved):

michael-scheurer (approved):

PR #846, opened 2023-02-01
fix(editor/api): taglist public, event filter, editor improvements

Itrulia:
1. Changed icons from rsuite to gmaterial
2. Made taglist publicly available
3. Added from & to filters to events


.eslintrc.json
```diff
@@ -1,11 +1,12 @@
 {
   "root": true,
   "ignorePatterns": ["**/*"],
-  "plugins": ["@nrwl/nx"],
+  "plugins": ["@nrwl/nx", "unused-imports"],
```
Itrulia, line 4: Already on master, came from resetting, decided to leave it in as it doesn't hurt

libs/api/__tests__/utility.ts
```diff
@@ -150,7 +151,8 @@ export async function createGraphQLTestClientWithPrisma(): Promise<TestClient> {
 
   return {
     testServerPublic,
-    testServerPrivate
+    testServerPrivate,
+    prisma
```
Itrulia, line 155: So we can clear the database if needed for a test

penk85 (approved):

PR #856, opened 2023-02-07
feat(api): session authentication for v2

Itrulia:
- [x] Write tests
- [x] Extract session code


package.json
```diff
@@ -8,34 +8,21 @@
     "migrate": "prisma migrate deploy && prisma db seed",
     "dev": "run-s start:docker watch",
     "clean": "run-p --print-label clean:**",
-    "test": "run-p --print-label test:**",
-    "build": "run-p --print-label build:api build:editor build:api-example build:website-example",
+    "clean:media": "docker-compose exec media rm -rf ./.media",
+    "lint": " nx run-many --target=lint --exclude website-example",
```
Itrulia, line 12: now uses nx to parallelize things

website-example excluded from listing as it's the only one that fails since we never had linting on it before nx

libs/api/src/lib/context.ts
```diff
@@ -223,70 +228,6 @@ const createOptionalsArray = <Data, Attribute extends keyof Data, Key extends Da
   return keys.map(id => dataMap[id] ?? null)
 }
 
-const getSessionByToken = async (
```
Itrulia, line 226: Moved to @wepublish/authentication/api

libs/api/src/lib/db/user.ts
```diff
@@ -26,39 +26,6 @@ export interface UserFilter {
   readonly text?: string
 }
 
-export const unselectPassword: Record<
```
Itrulia, line 29: Moved to @wepublish/user/api so it can be used in @wepublish/authentication/api

libs/api/src/lib/graphql/article/article.private-queries.ts
```diff
@@ -20,8 +20,8 @@ export const getArticleById = async (
 ): Promise<ArticleWithRevisions | null> => {
   const {roles} = authenticate()
 
-  const canGetArticle = isAuthorised(CanGetArticle, roles)
-  const canGetSharedArticle = isAuthorised(CanGetSharedArticle, roles)
+  const canGetArticle = hasPermission(CanGetArticle, roles)
```
Itrulia, line 23: Renamed to better express what it does

libs/api/src/lib/index.ts
```diff
@@ -20,7 +20,6 @@ export * from './graphql/permissions'
 
 export * from './db/user'
 export * from './db/userRole'
-export * from './db/session'
```
Itrulia, line 23: Maybe re-export @wepublish/authentication/api & @wepublish/user/api & @wepublish/permissions/api here?

libs/authentication/api/src/lib/optional-authentication.guard.ts
```diff
@@ -0,0 +1,27 @@
...
+    const ctx = GqlExecutionContext.create(context)
+    return ctx.getContext().req
+  }
+
+  public async canActivate(context: ExecutionContext): Promise<boolean> {
+    try {
+      await super.canActivate(context)
+    } catch (e) {
+      if (e instanceof UnauthorizedException) {
+        return true
```
Itrulia, line 19: return true if user is not logged in as we don't mind. Other errors should still throw

libs/membership/api/src/lib/dashboard/invoices/dashboard-invoice.resolver.ts
```diff
@@ -13,6 +14,7 @@ export class DashboardInvoiceResolver {
       Excludes cancelled or manually set as paid invoices.
     `
   })
+  @Permissions(CanGetInvoices)
```
Itrulia, line 17: How it can be used.

libs/permissions/api/src/lib/has-permission.ts
```diff
@@ -0,0 +1,32 @@
+import {UserRole} from '@prisma/client'
+import {Permission, AllPermissions, EditorPermissions, PeerPermissions} from './permissions'
+
+export function hasPermission(
+  neededPermissions: Permission | Permission[],
+  userRoles: UserRole[]
+): boolean {
+  const perms = Array.isArray(neededPermissions) ? neededPermissions : [neededPermissions]
```
Itrulia, line 8: Now also accepts multiple permissions

apps/api-example/src/nestapp/app.module.ts
```diff
@@ -10,7 +10,11 @@ import {DashboardModule} from '@wepublish/membership/api'
       driver: ApolloDriver,
       autoSchemaFile: './apps/api-example/schema-v2.graphql',
       sortSchema: true,
-      path: 'v2'
+      path: 'v2',
+      cors: {
```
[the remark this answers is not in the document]
tomaszdurka: lgtm

libs/authentication/api/src/lib/authentication.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    return null
+  }
+
+  public isSessionValid(session: AuthSession | null) {
+    if (!session) {
+      return false
+    }
+
+    return session.type === 'user' ? session.expiresAt > new Date() : true
```
tomaszdurka, line 77: Use `AuthSessionType.User`

I am not huge fan of this tennary.

Rather use
```
return session.type === AuthSessionType.User && session.expiredAt <= new Date();
```
?

libs/authentication/api/src/lib/optional-authentication.guard.ts
tomaszdurka: Why there is `UnauthorizedException` when we just try to authenticate.
Plus authentication is actually optional.

tomaszdurka: Where is this exception thrown?

libs/authentication/api/src/lib/authentication.guard.ts
```diff
@@ -0,0 +1,10 @@
+import {AuthGuard} from '@nestjs/passport'
+import {ExecutionContext} from '@nestjs/common'
+import {GqlExecutionContext} from '@nestjs/graphql'
+
+export class AuthenticationGuard extends AuthGuard('session') {
+  public getRequest(context: ExecutionContext) {
+    const ctx = GqlExecutionContext.create(context)
+    return ctx.getContext().req
+  }
+}
```
tomaszdurka, line 10: I understand this guard now checks if user has session and if it's valid, right?
If it is it's attaching user to context, right?

libs/authentication/api/src/lib/auth-session.ts
```diff
@@ -1,20 +1,20 @@
 import {User, UserRole} from '@prisma/client'
 
-export enum SessionType {
+export enum AuthSessionType {
```
tomaszdurka, line 3: I was actually ok with just `Session`.  Session could hold different data for different reasons. Just how it's used later determines if it's Auth piece.

tomaszdurka: Can still be `Auth`, was just curious about your reasoning.

libs/api/src/lib/index.ts
tomaszdurka: 👍

libs/api/src/lib/context.ts
```diff
@@ -939,7 +876,7 @@ export async function contextFromRequest(
     },
 
     authenticateUser() {
-      if (!session || session.type !== SessionType.User) {
+      if (!session || session.type !== AuthSessionType.User) {
```
tomaszdurka, line 879: Can you use `isSessionValid` ? Convert `isSessionValid` to static?

libs/api/src/lib/graphql/query.public.ts
```diff
@@ -335,7 +335,7 @@ export const GraphQLPublicQuery = new GraphQLObjectType<undefined, Context>({
       type: GraphQLPublicUser,
       description: 'This query returns the user.',
       resolve(root, args, {session}) {
-        return session?.type === SessionType.User ? session.user : null
+        return session?.type === AuthSessionType.User ? session.user : null
```
tomaszdurka, line 338: Use `isSessionValid` ?

libs/authentication/api/src/lib/session.strategy.ts
```diff
@@ -0,0 +1,24 @@
...
+      session: false // would use a cookie if set to true
+    })
+  }
+
+  public async validate(token: string) {
+    const session = await this.authService.getSessionByToken(token)
+    const isSessionValid = this.authService.isSessionValid(session)
+
+    if (!isSessionValid) {
+      return false
```
tomaszdurka, line 19: Is this throwing the `UnauthorizedException`?

libs/authentication/api/src/lib/optional-authentication.guard.ts
tomaszdurka: Just checked how I did that in one project. Actually it's not so great. I check "public override"

```
canActivate(context: ExecutionContext) {
    const isPublic = this.reflector.getAllAndOverride<boolean>(IS_PUBLIC_KEY, [
      context.getHandler(),
      context.getClass(),
    ]);
    if (isPublic) {
      return true;
    }
    return super.canActivate(context);
  }
```

libs/permissions/api/src/lib/permission.guard.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    const request = ctx.getContext().req
+    const session = request.user
+
+    if (!session) {
+      return false
+    }
+
+    return hasPermission(permissions, session.roles)
```
tomaszdurka, line 31: Looks good !

libs/permissions/api/src/lib/permission.module.ts
```diff
@@ -0,0 +1,13 @@
+import {Module} from '@nestjs/common'
+import {APP_GUARD} from '@nestjs/core'
+import {PermissionsGuard} from './permission.guard'
+
+@Module({
+  providers: [
+    {
+      provide: APP_GUARD,
+      useClass: PermissionsGuard
```
tomaszdurka, line 9: So it's on everywhere ! 👍

tomaszdurka (commented):
@Itrulia great job! Looks good. Left some comments, some optional, some good to change !

Just my thought it would be great to divide it into 2 PRs. One would be only about the Nestjs auth guards, and other all the refactoring. On the other hand you can filter out some commits maybe.

libs/authentication/api/src/lib/optional-authentication.guard.ts
Itrulia: That exception is thrown in the nestjs internals because the passport strategy http://www.passportjs.org/packages/passport-http-bearer/ returned false on the validate function.

This guard is only here to attach the user object to the request and nothing more.

libs/authentication/api/src/lib/authentication.service.ts
Itrulia: Oh yeah missed that, switched it to a string when I didn't fix the circular dependency before

libs/authentication/api/src/lib/authentication.guard.ts
Itrulia: Attaching the user is already done by the optional authentication guard, but this will also do it. 
This is for public routes where we want the user to be authenticated but doesn't need any specific roles

libs/authentication/api/src/lib/session.strategy.ts
Itrulia: Also, yes

libs/authentication/api/src/lib/auth-session.ts
Itrulia: I renamed to be with Auth as we have also `Session` in the database and it was confusing which one is which

Itrulia:
Was impossible to not do most of the refactoring without making the guards @tomaszdurka / @antkiewiczk because I can't import @wepublish/api, because @wepublish/api relies on that

libs/api/src/lib/graphql/query.public.ts
Itrulia: session is undefined if not valid in here

libs/authentication/api/src/lib/optional-authentication.guard.ts
tomaszdurka: For me now it seems we might do it wrong.

Guard is there to protect something from access (kinda). While we are actually extracting user from session if it exists.

Theoretically we should have:
1. Pipe (or such) which transform request enriching it with user if it exists in session.
2. Explicitly specified Guard to prevent request without specific permission (based on optional user within request).

libs/api/src/lib/graphql/query.public.ts
tomaszdurka: Wdym?

I mean this.

```
if (!session) {
  return false
}

return session.type === 'user' ? session.expiresAt > new Date() : true
```

Also the expiration is not considered in the query.public.ts

libs/authentication/api/src/lib/optional-authentication.guard.ts
Itrulia: That's as far as I have read up the recommended way by Nestjs and best practice  in express to attach the user to the request.

libs/api/src/lib/graphql/query.public.ts
Itrulia: It is.

`session: isSessionValid ? session : null,` in context.ts

penk85 (approved):

PR #866, opened 2023-02-24
f/WPC-953-public-endpoint-switch

Itrulia:
* Added `OneOfGuard` which replaced the `PermissionsGuard` as the global App Guard. This will now check if a guard is set and if one of them returns true, the endpoint is accessible.
* Created `AddMetadata` decorator to add to existing metadata.
* Created `SettingsGuard` which will check if a setting is set to true
* Created new settings for the dashboard APIs
* Rewrote parts of the settingList component to be more DRY and with less magic strings.


libs/api/src/index.ts
```diff
@@ -59,3 +58,8 @@ export * from './lib/challenges/algebraicCaptchaChallenge'
 export * from './lib/payments/payrexxSubscriptionPaymentProvider'
 
 export {getUserForCredentials} from './lib/graphql/user/user.queries'
+
+export * from '@wepublish/settings/api'
```
Itrulia, line 62: Re-export so it can be imported from the npm package

libs/api/src/lib/index.ts
```diff
@@ -1,66 +0,0 @@
-export * from './graphql/session'
```
Itrulia, line 1: Duplicate, came from merge conflicts when moving library with the Nx PR

apps/editor/src/app/locales/en.json
```diff
@@ -1274,7 +1281,13 @@
         "passwordToken": "How long (in minutes) the link sent via email to reset password will be valid. Defaults to one day (1440 / 60 / 24).",
         "peerToken": "Timeout in milliseconds after which we will no longer try to pull peer articles. Defaults to 3 seconds.",
         "invoiceReminders": "How many times we will remind user to pay the subscription (once per day).",
-        "askOperators": "Warning: changing one or more of the settings may break the page or have other unpredictable consequences. If you are unsure, please try talking to the operators first. If you are sure, feel free to continue."
+        "askOperators": "Warning: changing one or more of the settings may break the page or have other unpredictable consequences. If you are unsure, please try talking to the operators first. If you are sure, feel free to continue.",
+        "newSubscriptionsApiPublic": "Exposes subscriptions anonymized to the public.",
```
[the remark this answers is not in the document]
Itrulia: They are

libs/authentication/api/src/lib/one-of.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    if (!guards.length) {
+      return true
+    }
+
+    const checks = await Promise.all(
+      guards.map(async guard => {
+        try {
+          const canActivate$ = guard.canActivate(context)
```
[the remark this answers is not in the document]
Itrulia: Finish notation for observables

apps/editor/src/app/locales/en.json
```diff
@@ -1266,6 +1266,13 @@
       "allowGuestCommentRating": "Allow guest ratings on comments",
       "polls": "Polls",
       "guestPollVote": "Allow guest votes on polls",
+      "memberships": "Memberships",
```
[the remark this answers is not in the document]
Itrulia: Was told I can leave those out and Laurent will do it via the translation backend.

apps/editor/src/app/locales/en.json
```diff
@@ -1275,7 +1282,13 @@
...
         "peerToken": "Timeout in milliseconds after which we will no longer try to pull peer articles. Defaults to 3 seconds.",
         "invoiceReminders": "How many times we will remind user to pay the subscription (once per day).",
-        "askOperators": "Warning: changing one or more of the settings may break the page or have other unpredictable consequences. If you are unsure, please try talking to the operators first. If you are sure, feel free to continue."
+        "askOperators": "Warning: changing one or more of the settings may break the page or have other unpredictable consequences. If you are unsure, please try talking to the operators first. If you are sure, feel free to continue.",
+        "newSubscriptionsApiPublic": "Exposes subscriptions anonymized to the public.",
+        "activeSubscriptionsApiPublic": "Exposes subscriptions anonymized to the public.",
+        "renewingSubscriptionsApiPublic": "Exposes subscriptions anonymized to the public.",
+        "newDeactivationsApiPublic": "Exposes subscriptions anonymized to the public.",
```
penk85, line 1289: These all meant to have the same text? Should each one have something more descriptive?

apps/editor/src/app/routes/settingList.tsx
```diff
@@ -210,81 +178,83 @@ function SettingList() {
...
-          passwordExpire: resetPwdJwtExpiresMin.value,
-          peeringTimeout: peeringTimeoutMs.value,
-          invoiceTries: invoiceReminderTries.value,
-          invoiceFrequency: invoiceReminderFreq.value
-        }}
-        onSubmit={async validationPassed => {
-          validationPassed && (await setShowWarning(true))
-        }}>
+        formValue={formValue}
+        onSubmit={validationPassed => validationPassed && handleSettingListUpdate()}>
```
penk85, line 253: The setShowWarning isn't showing anymore. I assume you wanted it in the function? Also it seems like the validations aren't always preventing the form submit now because I can submit invalid data and get an error. I don't think that happened before.

apps/editor/src/app/locales/en.json
Itrulia: No they are intended to be like this.

apps/editor/src/app/routes/settingList.tsx
```diff
@@ -210,81 +178,83 @@ function SettingList() {
...
-          loginToken: sendLoginJwtExpiresMin.value,
-          passwordExpire: resetPwdJwtExpiresMin.value,
-          peeringTimeout: peeringTimeoutMs.value,
-          invoiceTries: invoiceReminderTries.value,
-          invoiceFrequency: invoiceReminderFreq.value
-        }}
-        onSubmit={async validationPassed => {
-          validationPassed && (await setShowWarning(true))
-        }}>
+        formValue={formValue}
```
tomaszdurka, line 252: Tried locally to use formValue and `Form.Control`, but indeed there is a problem with using `Toggle` via `accepter` due to value/checked *mismatch*. 
@penk85 we can try looking into that briefly once this one is merged.

libs/permissions/api/src/lib/permission.decorator.ts
```diff
@@ -1,7 +1,14 @@
...
+import {applyDecorators} from '@nestjs/common'
+import {OneOf} from '@wepublish/authentication/api'
+import {PermissionsGuard} from './permission.guard'
 import {Permission} from './permissions'
+import {AddMetadata} from '@wepublish/nest-modules'
 
 export const PERMISSIONS_METADATA_KEY = 'permissions'
 
+/**
+ * Causes the method/class to require that the current user has the permission to do the action.
```
tomaszdurka, line 10: @Itrulia 
If remember correctly these annotation influencing guard behaviour cannot be put on any class/method but just those related to endpoints like controller or resolvers, right?

tomaszdurka (approved):
Looks very good. As always good job !

I left some minor comments, small discussions. Feel free to reply.

libs/permissions/api/src/lib/permission.decorator.ts
Itrulia: They work now because I changed it to use getAllAndMerge (there's also getAllAndOverride) instead of get. I pass in the class and handler like so:

```
context.getHandler(),
context.getClass()
```

penk85 (approved):

apps/editor/src/app/routes/settingList.tsx
```diff
@@ -86,93 +84,72 @@ function SettingList() {
...
+    },
+    [SettingName.MakeExpectedRevenueApiPublic]: {
+      value: false,
+      name: SettingName.MakeExpectedRevenueApiPublic
+    },
+    [SettingName.MakeRevenueApiPublic]: {
+      value: false,
+      name: SettingName.MakeRevenueApiPublic
+    }
+  } as Record<SettingName, Setting>)
```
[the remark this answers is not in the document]
Itrulia: The reduce on itself works on a dynamic list, those are just the default values.

PR #883, opened 2023-03-30
feat(api-example): payrexx pm as env var and introduce cron jobs

Itrulia:


apps/api-example/src/app/index.ts
```diff
@@ -290,5 +287,22 @@ export async function runServer(app?: Application | undefined) {
     app
   )
 
+  const checkInvoiceCron = new CronJob('0 8 * * *', async function () {
```
Itrulia, line 290: daily at 8 it will be executed

PR #887, opened 2023-04-12
WPC-1099: Improved user first/lastname search

elias-summermatter:


elias-summermatter:
@michael-scheurer @Itrulia please have a look :-)

libs/api/src/lib/graphql/user/user.queries.ts
```diff
@@ -57,25 +57,159 @@ const createNameFilter = (filter: Partial<UserFilter>): Prisma.UserWhereInput =>
...
+const createUserNameFilter = (filter: Partial<UserFilter>): Prisma.UserWhereInput => {
+  const splitedString = filter.text.split(' ')
 
-const createTextFilter = (filter: Partial<UserFilter>): Prisma.UserWhereInput => {
-  if (filter?.text) {
+  if (splitedString.length === 1) {
     return {
       OR: [
         {
-          preferredName: {
```
penk85, line 66: Shouldn't the search still check `preferredName` in addition to first and last?

elias-summermatter:
@Itrulia what do you mean by "every possibility"? Isn't that an bit overkill? Or do you have an idea how to achive "every possibility"?

libs/api/src/lib/graphql/user/user.queries.ts
elias-summermatter: @penk85 ist still there just look at the new file without the changes there you will see that its still in the filter :-)

elias-summermatter:
@penk85 @antkiewiczk can someone of you aprove this too? :-)

penk85 (approved):

PR #899, opened 2023-04-27
WPC-1129: Importable events

antkiewiczk:
NOTE: lots of these files are just atoms/blocks moved from editor to ui library


libs/api/src/lib/graphql/event/event.query.ts
```diff
@@ -182,3 +182,17 @@ export const getEvents = async (
...
+export const getImportedEventsIds = async (event: PrismaClient['event']) => {
+  const externalEventsIds = event
+    .findMany({
+      where: {
+        externalSourceId: {
+          not: null
+        }
+      }
+    })
+    .then(res => res.map(single => single.externalSourceId))
```
Itrulia, line 195: prob should use async await here

libs/editor/api-v2/src/lib/graphql.ts
```diff
@@ -61,10 +68,49 @@ export type DashboardSubscription = {
   startsAt: Scalars['DateTime'];
 };
 
+export type Event = {
+  __typename?: 'Event';
+  createdAt: Scalars['DateTime'];
+  description: Scalars['JSON'];
```
Itrulia, line 74: This should be RichText, no?

Itrulia (changes requested):
We should delete the components in the editor that were moved out. Else we will be running into problems.

antkiewiczk:
> We should delete the components in the editor that were moved out. Else we will be running into problems.

but the components are used by other components in editor. Can this be done as separate PR? Probably together with this: https://wepublish.atlassian.net/browse/WPC-1199

Itrulia:
Whats the problem with importing them from the editor or am I missing something?

antkiewiczk:
> Whats the problem with importing them from the editor or am I missing something?

I don't think that it's possible to import the components from editor in an nx library. Moving components to libs/ui was introduced in previous task - consents, to solve this problem (being able to use atoms/blocks etc in libs/xxx). This is also when the components were duplicated and existed in editor and libs/ui. I can of course change this and import these components in editor and remove them, but this PR would be much bigger than now. I can of course do it in separate PR, but I asked about it some time ago and we decided to keep it together. I am aware of these problems (duplicating components, possibly having to change something twice - in editor and libs/ui etc), but this task is about importing events and I did it to make it work as this is the priority. That's why I proposed to do all of the components - atoms/blocks/panels etc in separate task, and I suggest doing it asap - right after merging this PR - to avoid similar problems in the future.

tomaszdurka:
Hey @antkiewiczk, cc @Itrulia  please move the UI moving from this PR.
We talked about it last Wednesday and we were clear in order to get smaller PR we need to plan our work better.
Additionally, currently the UI pieces are not moved, but copied which actually results in duplicated code and potential code split. 
It's probably like 2-3 hours of work to create the PR and then rebase this PR into the the new one another few.
Cheers !

PR #903, opened 2023-05-03
build(js): fix the same library being moved by nx using a patch

Itrulia:
Temporary fix until https://github.com/nrwl/nx/pull/16127 is merged and released

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,7 +12,7 @@
         "@nrwl/nx/enforce-module-boundaries": [
           "error",
           {
-            "enforceBuildableLibDependency": true,
+            "enforceBuildableLibDependency": false,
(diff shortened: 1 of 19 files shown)


.eslintrc.json
```diff
@@ -12,7 +12,7 @@
         "@nrwl/nx/enforce-module-boundaries": [
           "error",
           {
-            "enforceBuildableLibDependency": true,
```
Itrulia, line 15: with "external": "all", non buildable libraries can be dependent on

patches/@nrwl+js+15.3.0.patch
Itrulia, line 1: Applied by the patch-package library

tomaszdurka (approved):

PR #909, opened 2023-05-10
refactor(api): move oldContext to nest-modules, use modules instead of providers

Itrulia:


libs/nest-modules/src/lib/context/oldContext.service.ts
```diff
@@ -1,9 +1,13 @@
+/* eslint-disable @nrwl/nx/enforce-module-boundaries */
 import {Injectable, OnModuleInit} from '@nestjs/common'
-import {Context} from './context'
+
+// nx-ignore-next-line
```
Itrulia, line 4: import type means no actual circular dependency happens in code, just types. This is a temporary workaround until context has been moved.

nx-ignore-next-line removes it from the dependency graph check, so no circular dependency

libs/api/tsconfig.lib.json
```diff
@@ -7,5 +7,5 @@
     "types": ["node"]
   },
   "exclude": ["jest.config.ts", "src/**/*.spec.ts", "src/**/*.test.ts"],
-  "include": ["src/**/*.ts"]
+  "include": ["src/**/*.ts", "../nest-modules/src/lib/oldContext.service.ts"]
```
Itrulia, line 10:
```suggestion
  "include": ["src/**/*.ts"]
```

PR #910, opened 2023-05-10
Add gallery view to image list

penk85:
WPC-1163: Add gallery display to image list

### DIFF
--- a/apps/editor/src/app/routes/imageList.tsx
+++ b/apps/editor/src/app/routes/imageList.tsx
@@ -19,16 +19,27 @@ import {
 } from '@wepublish/ui/editor'
 import React, {useEffect, useState} from 'react'
 import {useTranslation} from 'react-i18next'
-import {MdDelete, MdEdit, MdOutlineAddPhotoAlternate, MdSearch} from 'react-icons/md'
+import {
+  MdDelete,
+  MdEdit,
+  MdOutlineAddPhotoAlternate,
+  MdSearch,
+  MdViewList,
+  MdViewModule
+} from 'react-icons/md'
 import {Link, useLocation, useNavigate, useParams} from 'react-router-dom'
 import {
   Button,
+  ButtonGroup,
+  Col as RCol,
   Drawer,
+  Grid as RGrid,
   IconButton as RIconButton,
   Input,
   InputGroup,
   Modal,
   Pagination,
+  Row,
   Table as RTable
 } from 'rsuite'
 import {RowDataType} from 'rsuite-table'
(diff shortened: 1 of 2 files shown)


apps/editor/src/app/routes/imageList.tsx
```diff
@@ -65,13 +134,18 @@ function ImageList() {
...
+  const [limit, setLimit] = useState(DEFAULT_TABLE_IMAGE_PAGE_SIZES[0])
 
   const [isUploadModalOpen, setUploadModalOpen] = useState(isUploadRoute)
   const [isEditModalOpen, setEditModalOpen] = useState(isEditRoute)
 
   const [editID, setEditID] = useState<string | undefined>(isEditRoute ? id : undefined)
 
+  const isGridKey = 'isGrid'
+  const [isGrid, setIsGrid] = useState<boolean>(
+    JSON.parse(localStorage.getItem(isGridKey) || 'true')
```
antkiewiczk, line 146: I'd maybe think about moving the value of isGridKey to utility.ts line 10 (enum LocalStorageKey).

antkiewiczk (commented):
Looks good! I'd maybe only just push the view change icons a bit to bottom so there's a little space between them and the search input.

apps/editor/src/app/routes/imageList.tsx
```diff
@@ -112,6 +199,118 @@ function ImageList() {
     }
   }, [location])
 
+  const ImageListView = () => (
```
Itrulia, line 202: Could we move these out of the component? These get recreated every render.

apps/editor/src/app/utility.ts
```diff
@@ -8,7 +8,8 @@ import {ElementID} from '../shared/elementID'
 import {ClientSettings} from '../shared/types'
 
 export enum LocalStorageKey {
-  SessionToken = 'sessionToken'
+  SessionToken = 'sessionToken',
+  isGridView = 'isGrid'
```
Itrulia, line 12: We probably want a more descriptive and specific key. Like `ImageListLayout` which is a enum/string

apps/editor/src/app/utility.ts
```diff
@@ -168,7 +169,7 @@ export function mapTableSortTypeToGraphQLSortOrder(sortType: SortType): SortOrde
 }
 
 export const DEFAULT_TABLE_PAGE_SIZES = [10, 20, 50, 100]
-export const DEFAULT_TABLE_IMAGE_PAGE_SIZES = [5, 10, 15]
+export const DEFAULT_TABLE_IMAGE_PAGE_SIZES = [10, 20, 50, 100]
```
Itrulia, line 172: Delete and use `DEFAULT_TABLE_PAGE_SIZES` instead

apps/editor/src/app/routes/imageList.tsx
```diff
@@ -47,6 +60,81 @@ const Img = styled.img`
...
+  position: relative;
+  & a {
+    color: unset;
+  }
+  &:hover {
+    & img {
+      height: 145px;
+      transition: 0.2s ease;
+    }
+    & div {
```
Itrulia, line 133:
```suggestion
    & ${Overlay} {
```

apps/editor/src/app/routes/imageList.tsx
```diff
@@ -47,6 +60,81 @@ const Img = styled.img`
...
+  flex-wrap: wrap;
+  row-gap: 20px;
+  margin: 20px 0;
+`
+
+const ImageWrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 20%;
```
Itrulia, line 103: nit: If we specify hard widths like these, we should use grid and not flex.
It's fine to keep it like this though, as the benefit now is not worth to spend more time on it.

PR #915, opened 2023-05-10
Move Mails outside API

tomaszdurka:


tomaszdurka:
I patchwork merged mailing "things" into mails library.
Not sure about exact role of mailcontroller in the subscription rewrite, but as MailProvider now requires pieces of mailcontroller I've moved it all together into the mails library. 
That can be adjusted a bit before the final merge of subs branch.

@elias-summermatter @LukasSkywalker @Itrulia please review

PR #916, opened 2023-05-10
Move logger out of API

tomaszdurka:


tomaszdurka:
Moving it to utility is not the best, and general implementation with global holders is pretty awful , but I wanted to move logger out of the API. It is generally used in modules like payments.

tomaszdurka:
@Itrulia @LukasSkywalker @elias-summermatter please review

Itrulia (approved):

PR #922, opened 2023-05-15
Media adapter lib

antkiewiczk:


libs/image/api/src/lib/media-adapter.service.ts
```diff
@@ -0,0 +1,24 @@
+import {Injectable} from '@nestjs/common'
+import {Image} from '@prisma/client'
+import {FileUpload} from 'graphql-upload'
+import {ImageTransformation, UploadImage} from './image-upload'
+import {ArrayBufferUpload, MediaAdapter} from './media-adapter'
+
+@Injectable()
+export class MediaAdapterService implements MediaAdapter {
```
tomaszdurka, line 8: Is it actually possible to use abstract class here? I would prefer that.

```
@Injectable()
export abstract class MediaAdapterService implements MediaAdapter {
  abstract uploadImage(fileUpload: Promise<FileUpload>): Promise<UploadImage>;

  abstract uploadImageFromArrayBuffer(arrayBufferUpload: Promise<ArrayBufferUpload>): Promise<UploadImage>;

  abstract deleteImage(id: string): Promise<boolean>;

  abstract getImageURL(image: Image, transformation?: ImageTransformation | undefined): Promise<string>;
}
```

tomaszdurka (approved):

PR #924, opened 2023-05-15
Allow same variables on deployment and app-specific envs

tomaszdurka:
- Allow merging "envs" with the same variables
- Bump chart to 2.0.1


tomaszdurka:
@Itrulia fyi
So lack of this was blocking our automatic deployments. In default helm values there was NDOE_ENV for website, when you set the NODE_ENV for deployment, this resulted in duplicate parameters for ENV, resulting in error.

PR #928, opened 2023-05-16
f/WPC-1195-add-designation-field

michael-scheurer:


libs/api/src/lib/graphql/user/user.public-mutation.ts
```diff
@@ -148,9 +148,15 @@ export const updatePublicUser = async (
       name,
       firstName,
       preferredName,
-      address: {
-        update: address
-      }
+      address: address
```
michael-scheurer, line 151: in case a new user has no address, it would throw an error when trying to update a non-existing address database entry.

Itrulia:
Wouldn't this make more sense to use the user properties for?

Itrulia:
Deleted my old comment because I saw Laurent was fine with adding the field to the user.
Designation seems a bit too specific to politicians, should we use `flair` maybe?

michael-scheurer:
@elias-summermatter @Itrulia May I get your review again? Thank you so much!

elias-summermatter (approved):

Itrulia (approved):

PR #931, opened 2023-05-18
WPC-1198-fixing-error-in-migration-to-convert-jsonb_array-to-jsonb

elias-summermatter:


Itrulia (dismissed):

Itrulia:
(haven't tested it yet, but the SQL looks a lot better)

elias-summermatter:
@Itrulia but it does not work... and its for me obvious why:
```
UPDATE "table"
SET blocks2 = jsonBlocks
FROM (
    SELECT jsonb_agg(j::jsonb)
    FROM (
        SELECT unnest(field)
        FROM "table"
    ) as x(j) 
) as x(jsonBlocks); => is for every entry the same  object  (list with all blocks from the hole db, multi mb in size)
```

Itrulia (approved):

michael-scheurer (approved):

PR #932, opened 2023-05-18
WPC-1124: Event Teaser

antkiewiczk:


Itrulia (approved):

penk85 (approved):
Looks good. 
(Only maybe small ui fix, I'd swap the position of custom tab and events tab so the custom is last on the list)

antkiewiczk:
> Looks good. (Only maybe small ui fix, I'd swap the position of custom tab and events tab so the custom is last on the list)

Let's add this to list of improvements to Agenda Basel in general :)

PR #934, opened 2023-05-22
fix/create-poll-without-info-text

michael-scheurer:
If an editor leaves info text field empty, the application throws an error. In case of an empty info text field we have to pass undefined, not null.


Itrulia:
Hey hey, whats the problem here?

`poll.infoText || undefined` would only do something different if `infoText` is 0 or ''. Or is there a problem when infoText is null?

antkiewiczk:
> Hey hey, whats the problem here?
> 
> `poll.infoText || undefined` would only do something different if `infoText` is 0 or ''. Or is there a problem when infoText is null?

yep also asked Michael on slack to specify the issue/fix 😆

michael-scheurer:
Now, providing some information about this pr.

Itrulia:
We should fix the root cause (accepting null too) IMO.

Itrulia:
Hey hey, it's more that the API should handle this instead of the client as the interface allows to use null.
We don't handle this anywhere unfortunately, but we handle the difference between null and undefined in other cases already IIRC (undefined in prisma means it won't be set/changed at all, even if it already has a value)

PR #935, opened 2023-05-23
timepicker alignment

penk85:
WPC-1165: align timepicker with datepicker.


antkiewiczk (approved):

Itrulia:
image

Nitpick: causes the right arrow to be misaligned, but that probably doesn't bother anyone but me :D

Good improvement!

Itrulia (approved):

PR #941, opened 2023-05-24
Fix metadata panel add new author

penk85:
WPC-1128: add new author from metadata panel was not working. Change bio field in Author to be optional.


apps/editor/src/app/panel/authorCheckPicker.tsx
```diff
@@ -50,7 +50,8 @@ export function AuthorCheckPicker({list, disabled, onChange}: AuthorCheckPickerP
       variables: {
         input: {
           name: authorsFilter,
-          slug: slugify(authorsFilter)
+          slug: slugify(authorsFilter),
+          bio: []
```
Itrulia, line 54: Should be done on the API level, not the frontend

penk85: Make optional field or add default?

Itrulia: Make it optional

Itrulia (approved):

antkiewiczk (approved):

PR #945, opened 2023-05-24
feat(website): session context implementation

Itrulia:


libs/authentication/website/src/lib/session.context.ts
Itrulia, line 1: Allows operators to choose what client storage they want to use and the website builder to have access to user data / login information

penk85 (dismissed):

antkiewiczk (approved):

PR #954, opened 2023-05-31
fix(api): unique constraint violation when publishing article pending

Itrulia:
Also some small refactoring to make the code smaller.

Delete happened after the update (this is something we learned, 2nd time this happened).
Changed it so we delete the revision and recreate it, ideally we would flip the schema a bit but this is overkill for this.

### DIFF
--- a/libs/api/src/lib/graphql/article/article.private-mutation.ts
+++ b/libs/api/src/lib/graphql/article/article.private-mutation.ts
@@ -5,6 +5,30 @@ import {DuplicateArticleSlugError, NotFound} from '../../error'
 import {authorise} from '../permissions'
 import {CanCreateArticle, CanDeleteArticle, CanPublishArticle} from '@wepublish/permissions/api'
 
+const fullArticleInclude = {
+  draft: {
+    include: {
+      properties: true,
+      authors: true,
+      socialMediaAuthors: true
+    }
+  },
+  pending: {
+    include: {
+      properties: true,
+      authors: true,
+      socialMediaAuthors: true
+    }
+  },
+  published: {
+    include: {
+      properties: true,
+      authors: true,
+      socialMediaAuthors: true
+    }
+  }
+}
+
 export const deleteArticleById = async (
   id: string,
   authenticate: Context['authenticate'],
@@ -17,29 +41,7 @@ export const deleteArticleById = async (
     where: {
       id
     },
-    include: {
-      draft: {
-        include: {
-          properties: true,
-          authors: true,
-          socialMediaAuthors: true
-        }
-      },
-      pending: {
-        include: {
-          properties: true,
-          authors: true,
-          socialMediaAuthors: true
-        }
-      },
-      published: {
-        include: {
-          properties: true,
-          authors: true,
-          socialMediaAuthors: true
-        }
-      }
-    }
+    include: fullArticleInclude
   })
 
   if (!article) {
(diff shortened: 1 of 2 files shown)


libs/api/src/lib/graphql/article/article.private-mutation.ts
```diff
@@ -468,11 +324,20 @@ export const publishArticle = async (
   }
 
   if (publishAt > new Date()) {
-    return articleClient.update({
-      where: {id},
-      data: {
-        pending: {
-          upsert: {
+    const deleOldRevisions = prisma.articleRevision.deleteMany({
```
michael-scheurer, line 327: typo? deleteOldRevision, no?

libs/api/src/lib/graphql/article/article.private-mutation.ts
```diff
@@ -468,11 +324,20 @@ export const publishArticle = async (
...
+      where: {
+        id: {
+          in: [article.pending?.id].filter((id): id is string => Boolean(id))
+        }
+      }
+    })
+
+    const [, updatedArticle] = await prisma.$transaction([
+      deleOldRevisions,
```
michael-scheurer, line 336: idem

libs/api/src/lib/graphql/article/article.private-mutation.ts
```diff
@@ -225,47 +183,7 @@ export const unpublishArticle = async (
...
-              public: true
-            }
-          },
-          authors: true,
-          socialMediaAuthors: true
-        }
-      },
-      published: {
-        include: {
-          properties: {
```
michael-scheurer, line 257: What you delete here, seems to be different from fullArticleInclude, no?

libs/api/src/lib/graphql/article/article.private-mutation.ts
```diff
@@ -384,57 +280,17 @@ export const publishArticle = async (
...
-            select: {
```
michael-scheurer, line 419: idem

libs/api/src/lib/graphql/article/article.private-mutation.ts
```diff
@@ -493,74 +358,33 @@ export const publishArticle = async (
...
+  const deleOldRevisions = prisma.articleRevision.deleteMany({
+    where: {
+      id: {
+        in: [article.pending?.id, article.published?.id].filter((id): id is string => Boolean(id))
+      }
+    }
+  })
+
+  const [, updatedArticle] = await prisma.$transaction([
+    deleOldRevisions,
```
michael-scheurer, line 383: idem

libs/api/src/lib/graphql/article/article.private-mutation.ts
Itrulia: Yeah it's a bit different but `GraphQLMetadataProperty` already does this. We fetch a bit more data from the DB (but the same rows) but the code is a bit cleaner. The API should be unaffected by this.

Itrulia: Actually this caused an issue! Going to revert this

michael-scheurer (dismissed):

antkiewiczk:
Is this supposed to be a draft? LGTM

Itrulia:
@michael-scheurer / @antkiewiczk updated with tests

PR #956, opened 2023-06-01
Add basic JobsModule with JobsService based on PgBoss

tomaszdurka:


tomaszdurka:
> what's the exact purpose of this module?

- Ability to run scheduled jobs without concurrency problems.
- Ability to queue async jobs if needed
- Using Postgree for that

libs/jobs/src/lib/jobs.module.ts
```diff
@@ -0,0 +1,34 @@
...
+  exports: []
+})
+export class JobsModule {
+  static forRoot(options?: JobsModuleOptions): DynamicModule {
+    return {
+      module: JobsModule,
+      providers: [
+        {
+          provide: PgBossService,
+          useValue: new PgBossService(options?.databaseUrl ?? process.env['DATABASE_URL']!)
```
Itrulia, line 28: should probably use `useFactory` here and instead of accessing process.env directly, could use the `ConfigService` https://docs.nestjs.com/techniques/configuration

Itrulia (changes requested):
Looks good! Just one small thing about the creating of the service 👍🏼

libs/jobs/src/lib/jobs.module.ts
Itrulia: JobsModule would be dependent on it, but not the service. Imo if you access the process.env it should use the ConfigService.

Itrulia (approved):

PR #959, opened 2023-06-02
fix(membership-api): active subscribers was ignoring paidUntil

Itrulia:


libs/membership/api/src/lib/dashboard/subscriptions/dashboard-subscription.resolver.spec.ts
```diff
@@ -324,11 +345,7 @@ describe('DashboardSubscriptionResolver', () => {
     await request(app.getHttpServer())
       .post('')
       .send({
-        query: activeSubscribersQuery,
-        variables: {
```
Itrulia, line 328: query didn't have any args

antkiewiczk (approved):

PR #964, opened 2023-06-09
fix(api): updatePaymentProviderCustomers prisma mutation was broken

Itrulia:


libs/api/__tests__/specs/user.test.ts
```diff
@@ -48,185 +52,245 @@ describe('Users', () => {
...
-          createUser: {
-            id: expect.any(String),
-            email: expect.any(String),
-            emailVerifiedAt: expect.any(Date)
-          }
-        }
-      })
-      ids.unshift(res.data?.createUser.id)
-    })
+        test('can update payment provider customers', async () => {
```
Itrulia, line 77: This test makes sure we are not running into the same issues as with the authorIds. This makes sure we can update an already existing value

antkiewiczk (approved):

penk85 (approved):

PR #966, opened 2023-06-13
chore(editor): upgrade i18next

Itrulia:


antkiewiczk:
from what I see in docs, in the freshest version (23) of i18next `returnNull` is set to `false` by default so there should be no need for a rewrite like this IMO, at least if I understand this correctly. Would prefer not to rewrite all translations to return `xxx || undefined` if it isn't absolutely necessary

antkiewiczk:
Screenshot 2023-06-16 at 10 33 20

https://github.com/i18next/i18next/issues/1884

Itrulia:
That release came 1 day ago! God damnit.

Itrulia:
Upgraded to v23, discarded most of the changes

antkiewiczk (approved):

PR #973, opened 2023-06-16
feat(event/import/api): make event providers async

Itrulia:


Itrulia:
@antkiewiczk feel free to merge when you think it's good!

PR #978, opened 2023-06-19
refactor(storybook): move from tools to libs to satisify IDEs

Itrulia:


antkiewiczk (commented):
looks good, but if it's moved from tools to libs, then why can't I see deleted files in tools?

Itrulia:
Because it's moved, not deleted

antkiewiczk (approved):

PR #982, opened 2023-06-26
feat(block-content-website): implemented teaser grid block

Itrulia:
Also cleaned up schema and auto generate possibleTypes

- [x] Add mocks for article, pages & blocks stories


codegen.yml
```diff
@@ -60,6 +60,7 @@ generates:
       - typescript
       - typescript-operations
       - typescript-react-apollo
+      - fragment-matcher
```
Itrulia, line 63: generates possibleTypes so that we can use fragments and nesting.

Prob can move editor to this at some point too, then the startup of the editor is a bit faster, and in case the API is down we are not getting a blank page.

libs/article/website/src/lib/__snapshots__/article-container.spec.tsx.snap
```diff
@@ -5,552 +5,6 @@ exports[`Article Container should render ById 1`] = `
   <article
     class="css-1thzeso"
   >
-    <div
```
Itrulia, line 8: this shouldn't be, hm.

libs/page/website/src/lib/page.stories.tsx
```diff
@@ -1,12 +1,90 @@
...
+  largeURL: 'https://unsplash.it/500/300',
+  mediumURL: 'https://unsplash.it/300/200',
+  smallURL: 'https://unsplash.it/200/100',
+  squareBigURL: 'https://unsplash.it/800/800',
+  squareLargeURL: 'https://unsplash.it/500/500',
+  squareMediumURL: 'https://unsplash.it/300/300',
+  squareSmallURL: 'https://unsplash.it/200/200'
+} as FullImageFragment
+
+const author = {
```
antkiewiczk, line 39: same as with my previous comment

libs/website/api-v2/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './lib/client'
 export * from './lib/graphql'
+export {default as possibleTypes} from './lib/graphql'
```
antkiewiczk, line 3: that's an interesting syntax

antkiewiczk (approved):

libs/block-content/website/src/lib/teaser-grid-block.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+  TeaserGridBlock as TeaserGridBlockType
+} from '@wepublish/website/api'
+import {TeaserGridBlock} from './teaser-grid-block'
+
+export default {
+  component: TeaserGridBlock,
+  title: 'Blocks/Teaser Grid'
+} as Meta
+
+const image = {
```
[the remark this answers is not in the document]
Itrulia: Theres a ticket for it already open

PR #989, opened 2023-07-05
WPC-1240: hide GraphQLPublicUser properties by session id

antkiewiczk:


libs/api/src/lib/graphql/utils.ts
```diff
@@ -1,8 +1,17 @@
...
 ): keyof Prisma.DateTimeFilter => {
   return comparison === DateFilterComparison.Equal ? 'equals' : comparison
 }
+
+type IsMeBySessionArgs = {
+  id: string
+  session?: AuthSession
+}
+
+export const isMeBySession = ({id, session}: IsMeBySessionArgs) =>
```
Itrulia, line 16: can we use multiple args instead of an object here? Doesn't improve anything here

Itrulia (commented):
Could we add some tests for this? Would be be amazing!

Itrulia (approved):

libs/api/src/lib/graphql/user.ts
```diff
@@ -170,20 +171,58 @@ export const GraphQLPublicUser = new GraphQLObjectType<UserWithRelations, Contex
     id: {type: GraphQLNonNull(GraphQLString)},
     name: {type: GraphQLNonNull(GraphQLString)},
     firstName: {type: GraphQLString},
-    email: {type: GraphQLNonNull(GraphQLString)},
-    preferredName: {type: GraphQLString},
-    address: {type: GraphQLUserAddress},
-    flair: {type: GraphQLString},
+    email: {
+      type: GraphQLNonNull(GraphQLString),
```
Itrulia, line 175: I guess returns empty string to not have a breaking change? Could also make it nullable I guess, I don't mind either way.

libs/api/src/lib/graphql/user.ts
```diff
@@ -170,20 +171,58 @@ export const GraphQLPublicUser = new GraphQLObjectType<UserWithRelations, Contex
...
+        preferredName && isMeBySession(id, session) ? preferredName : ''
+      )
+    },
+    address: {
+      type: GraphQLUserAddress,
+      resolve: createProxyingResolver(({address, id}, _, {session}) =>
+        address && isMeBySession(id, session) ? address : ''
+      )
+    },
+    flair: {
```
Itrulia, line 192: Flair is fine to expose, even needed

libs/api/src/lib/graphql/user.ts
```diff
@@ -170,20 +171,58 @@ export const GraphQLPublicUser = new GraphQLObjectType<UserWithRelations, Contex
...
+            })
+          : []
+      )
     },
     image: {
       type: GraphQLImage,
-      resolve: createProxyingResolver(({userImageID}, _, {prisma: {image}}) =>
-        userImageID
+      resolve: createProxyingResolver(({userImageID, id}, _, {prisma: {image}, session}) =>
+        userImageID && isMeBySession(id, session)
```
Itrulia, line 225: Image is needed for comments

PR #994, opened 2023-07-11
feat(block-content-website): implemented teaser grid block

Itrulia:


libs/comments/website/src/lib/comment-list/comment-list-item.tsx
```diff
@@ -33,21 +37,10 @@ export const CommentListItemName = styled('div')`
   font-weight: ${({theme}) => theme.typography.fontWeightBold};
 `
 
-export const CommentListItemVerifiedBadge = styled('div')<{isVerified: boolean; isTeam: boolean}>`
+export const CommentListItemVerifiedBadge = styled('div')`
   display: grid;
   align-items: center;
-
-  ${({theme, isVerified, isTeam}) => css`
```
Itrulia, line 40: Removed staff badge as that type is when a staff creates a comment in the editor and not for a comment made by that staff.

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -7,20 +7,21 @@ import {Reducer, useReducer} from 'react'
 import {MdClose, MdMenu} from 'react-icons/md'
 import {navigationLinkToUrl} from '../link-to-url'
 
-export const NavbarWrapper = styled('nav')<{isMenuOpen: boolean}>`
+export const NavbarWrapper = styled('nav')`
```
Itrulia, line 10: Fixes not being able to set a background-color without breaking the open menu background-color

apps/website-example/pages/a/[id]/[slug].tsx
```diff
@@ -22,7 +23,7 @@ export function ArticleBySlug() {
 
           <ArticleWrapper>
             <H5 component={'h2'}>Kommentare</H5>
-            <CommentListContainer id={id as string} />
+            <CommentListContainer id={id as string} type={ApiV1.CommentItemType.Article} />
```
Comment (Itrulia) [inline apps/website-example/pages/a/[id]/[slug].tsx:26]: Type is needed for adding a comment

libs/api/__tests__/api/private/index.ts
```diff
@@ -2156,7 +2156,7 @@ export type QueryRemotePeerProfileArgs = {
 
 
 export type QuerySettingArgs = {
-  name?: InputMaybe<Scalars['String']>;
+  name: Scalars['String'];
```
Itrulia, line 2159: Used to be optional in graphql and then verified to be non optional in the resolver. Changed it to be non optional by graphql

libs/api/__tests__/specs/comment.test.ts
```diff
@@ -128,19 +128,7 @@ describe('Comments', () => {
       }
     })
 
-    expect(getComments).toMatchSnapshot({
```
Itrulia, line 131: Was a race condition as more comments could be available.

antkiewiczk (approved):

PR #1000, opened 2023-07-18
feat(block-content-website): implemented poll block

Itrulia:
Also

1. Moved block-content files around for better navigation
2. Removed caching from mockprovider
3. Added immutability assumption to apollo for speed
4. Fixed GraphQL nullable properties that are never nullable

Anonymous poll voting comes with a different PR

When not logged in the highest vote will be highlighted
image

When logged in and having cast a vote, your vote will be highlighted
image

When logged in, poll is still open and haven't cast a vote yet:
image

When logged in but poll is closed, immediately go to result view:
image


libs/comments/website/src/lib/comment-list/comment-list.tsx
```diff
@@ -23,7 +24,7 @@ export const CommentList = ({data, className}: BuilderCommentListProps) => {
         />
       ))}
 
-      {!data?.comments.length && <Alert severity="info">Keine Kommentare vorhanden.</Alert>}
```
Itrulia, line 26: Was buggy

libs/article/website/src/lib/article.spec.tsx
```diff
@@ -1,13 +1,19 @@
...
+import {MockedProvider} from '@apollo/client/testing'
 
 const storiesCmp = composeStories(stories)
 
 describe('Article', () => {
   Object.entries(storiesCmp).forEach(([story, Component]) => {
     it(`should render ${story}`, () => {
-      const {asFragment} = render(<Component />)
+      const {asFragment} = render(
+        <MockedProvider {...Component.parameters?.apolloClient}>
```
Itrulia, line 12: Needed now as the PollBlock uses Apollo directly

libs/events/import/api/jest.config.ts
```diff
@@ -2,14 +2,15 @@
 export default {
   displayName: 'events/import-api',
   preset: '../../../../jest.preset.js',
-  globals: {
```
Itrulia, line 5: This way was deprecated, I guess this library was generated with an older nx version.

Itrulia:
@antkiewiczk please note that the PR is a lot smaller than it seems, as a lot of files have been moved.
This is indicated by an icon and the diff tells you so.

libs/block-content/website/src/lib/poll/poll-block.tsx
```diff
@@ -0,0 +1,149 @@
...
+  gap: ${({theme}) => theme.spacing(1)};
+`
+
+export const PollBlockMeta = styled('div')`
+  font-size: 0.875em;
+  color: ${({theme}) => theme.palette.text.disabled};
+`
+
+export const PollBlock = ({poll, className}: BuilderPollBlockProps) => {
+  const [fetchUserVote, loggedInVote] = useUserPollVoteLazyQuery()
```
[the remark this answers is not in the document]
Itrulia: We can, but we need to pass it down multiple layers, thats the discussion I want to open up here

Itrulia:
@antkiewiczk I know it has megre conflicts :D The discussion here is more important, there's no reason to keep it updated as it's blocked until that discussion is resolved.

libs/article/website/src/lib/article-container.tsx
```diff
@@ -24,5 +30,24 @@ export function ArticleContainer({onQuery, id, slug, className}: ArticleContaine
...
 
-  return <Article data={data} loading={loading} error={error} className={className} />
+  return (
+    <PollBlockProvider>
+      <Article data={data} loading={loading} error={error} className={className} />
+    </PollBlockProvider>
+  )
+}
+
+function PollBlockProvider({children}: PropsWithChildren) {
```
antkiewiczk, line 40: prob. not a big deal, but it's defined after using

libs/block-content/website/src/lib/poll/poll-block.context.tsx
```diff
@@ -0,0 +1,29 @@
...
+import {useUser} from '@wepublish/authentication/website'
+import {
+  PollVoteMutation,
+  PollVoteMutationVariables,
+  UserPollVoteQuery,
+  UserPollVoteQueryVariables
+} from '@wepublish/website/api'
+import {createContext, useContext} from 'react'
+
+export const PollBlockContext = createContext<
```
antkiewiczk, line 11: 👍

antkiewiczk (approved):

libs/article/website/src/lib/article-container.tsx
Itrulia: Hoisting does the trick because it's used inside a function :D

PR #1002, opened 2023-07-21
WPC-1068: Embed block + other embeds blocks

antkiewiczk:


libs/block-content/website/src/lib/tik-tok-video-block.tsx
```diff
@@ -7,6 +7,27 @@ export const isTikTokVideoBlock = (block: Block): block is TikTokVideoBlockType
 
 export const TikTokVideoBlockWrapper = styled('div')``
 
+export const TikTokVideoBlockPlayer = styled('div')`
```
Itrulia, line 10: You can use react-player for this. See youtube or vimeo

Itrulia (commented):
Is this in draft or ready for review?

antkiewiczk:
> Is this in draft or ready for review?

Marked it as ready, but the tests are failing because react-player is changing the id every time when writing snapshot. I am now looking at solution for this. If you have anything else, let me know

libs/block-content/website/src/lib/tik-tok-video-block.tsx
antkiewiczk: I used iframe directly, as react-player wouldn't run (some cross-origin issues)

Itrulia:
@antkiewiczk many embeds are missing, hence im asking.

libs/block-content/website/src/lib/tik-tok-video-block.tsx
Itrulia: My bad, didn't see Tiktok was not supported by react-player as it TikTok doesn't have a non iframe API.

Can we atleast use a CSS specifity of 1, not hardcode widths/heights and give it a proper title?

libs/block-content/website/src/lib/facebook-video-block.tsx
```diff
@@ -24,6 +24,7 @@ export const FacebookVideoBlock = ({
       height={'auto'}
       url={`https://www.facebook.com/${userID}/videos/${videoID}/`}
       controls={true}
+      config={{facebook: {playerId: 'facebook-player'}}}
```
Itrulia, line 27: this now causes that you can not have the multiple facebook videos as they have the same Id. 
Please combine this with `const id = useId()`

libs/block-content/website/src/lib/bildwurf-ad-block.tsx
```diff
@@ -0,0 +1,92 @@
...
+    load()
+    try {
+      window._ASO.loadAd('bildwurf-injection-wrapper', zoneID)
+    } catch (error) {
+      console.warn('could not call _ASO.loadAd()')
+    }
+  }, [])
+
+  return (
+    <BildwurfBlockWrapper className={className} id="bildwurf-injection-wrapper">
```
antkiewiczk, line 88: I tried using `const id = useId()`, but it wouldn't load

libs/block-content/website/src/lib/bildwurf-ad-block.tsx
```diff
@@ -0,0 +1,92 @@
...
+import styled from '@emotion/styled'
+import {BuilderBildwurfAdBlockProps} from '@wepublish/website/builder'
+
+function useScript(src: string, checkIfLoaded: () => boolean, crossOrigin = false) {
+  const scriptRef = useRef<HTMLScriptElement | null>(null)
+
+  const [isLoading, setLoading] = useState(false)
+  const [isLoaded, setLoaded] = useState(() => checkIfLoaded())
+
+  useEffect(() => {
```
Itrulia, line 13: Can we use `<Script>` instead? (It's in the website builder, probably won't work on storybook but should in the next app as the storybook mock is very basic)

libs/block-content/website/src/lib/bildwurf-ad-block.tsx
Itrulia: Just checked, while the script doesn't include it and even gives a way to specify it, another script that is loaded hardcodes it.

Itrulia (approved):

PR #1003, opened 2023-07-21
feat(website): add apollo cache for SSR

Itrulia:
Also:

1. Removed onVariablesChange from author list as it was not used
2. Added sort and order that was missing from author list
3. Added SSR to single author view
4. Changed fallback: true to fallback: 'blocking', as the router params were undefined to start with

### DIFF
--- a/apps/website-example/src/next-wepublish-link.tsx
+++ b/apps/website-example/src/next-wepublish-link.tsx
@@ -3,11 +3,8 @@ import {BuilderLinkProps} from '@wepublish/website'
 import NextLink from 'next/link'
 
 export const NextWepublishLink = ({children, href, ...props}: BuilderLinkProps) => {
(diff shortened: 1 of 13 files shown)


apps/website-example/src/next-wepublish-link.tsx
```diff
@@ -3,11 +3,8 @@ import {BuilderLinkProps} from '@wepublish/website'
 import NextLink from 'next/link'
 
 export const NextWepublishLink = ({children, href, ...props}: BuilderLinkProps) => {
-  const hrefWithoutOrigin =
```
Itrulia, line 6: Not needed it seems, it doesn't do a full reload.
Would have caused problems in SSR.

antkiewiczk (approved):

PR #1010, opened 2023-07-27
fix(api): facebook video block missing in public API

Itrulia:


libs/api/src/lib/graphql/blocks.ts
```diff
@@ -1257,6 +1257,7 @@ export const GraphQLPublicBlock: GraphQLUnionType = new GraphQLUnionType({
     GraphQLImageGalleryBlock,
     GraphQLListicleBlock,
     GraphQLFacebookPostBlock,
+    GraphQLFacebookVideoBlock,
```
antkiewiczk, line 1260: that's really the whole change? 🤔

Itrulia: Yes :D

PR #1014, opened 2023-08-01
updated some deps and integrated config

mandarini:
* Removed root `.storybook` config folder, since it's the new approach that's suggested by Storybook and Nx
* Integrated settings from root `.storybook` config folder into the project-level configuration
* Removed duplicate babel emotion plugin (from storybook config and project config)
* Added `NODE_ENV: 'production'` environment variable to Storybook
* Updated some dependencies.

If issue persists, use `NODE_ENV: 'production'` in an `.env` file.


Itrulia:
Thank you!

PR #1018, opened 2023-08-07
Add code coverage reports

tomaszdurka:


.github/workflows/test.yml
```diff
@@ -23,7 +23,14 @@ jobs:
         with:
           node-version: '16'
       - run: npm ci
-      - run: npm test
+      - run: npm test-ci
```
Itrulia, line 26: couldnt we just add -- --code-coverage instead of adding more scripts to npm?

Itrulia (changes requested):

.github/workflows/test.yml
tomaszdurka: We could indeed.

Itrulia: Would prefer that as too many scripts in package.json gets confusing fast

Itrulia (approved):

PR #1021, opened 2023-08-08
WPC-1067: Break Block (website builder)

antkiewiczk:


Itrulia (changes requested):
Article (and Container), Page (and Container) are missing the updated mock for the page break.

Also made a PR for CSS changes: https://github.com/wepublish/wepublish/pull/1022

Itrulia (approved):

PR #1022, opened 2023-08-09
Break Block reverse improvements & smaller screen design improvements

Itrulia:
Made some small improvements to the break block, as the square image was huge on mobile/tablet, spacing was too big on smaller screens etc.

Also changed the reverse to use `order` instead of multiple properties at once

### DIFF
--- a/libs/block-content/website/src/lib/break/break-block.tsx
+++ b/libs/block-content/website/src/lib/break/break-block.tsx
@@ -5,7 +5,7 @@ import {Block, LinkPageBreakBlock as LinkPageBreakBlockType} from '@wepublish/we
 export const isBreakBlock = (block: Block): block is LinkPageBreakBlockType =>
   block.__typename === 'LinkPageBreakBlock'
 
-export const BreakBlockWrapper = styled('div')`
+export const BreakBlockWrapper = styled('div')<{reverse?: boolean}>`
   display: grid;
   gap: ${({theme}) => theme.spacing(2)};
   justify-content: center;
@@ -13,6 +13,14 @@ export const BreakBlockWrapper = styled('div')`
   ${({theme}) => theme.breakpoints.up('md')} {
     grid-template-columns: 1fr 2fr;
   }
+
+  ${({theme, reverse}) =>
+    reverse &&
+    css`
+      ${theme.breakpoints.up('md')} {
+        grid-template-columns: 2fr 1fr;
+      }
+    `}
 `
 
 export const BreakBlockSegment = styled('div')<{reverse?: boolean}>`
@@ -46,7 +54,7 @@ export const BreakBlock = ({
   const reverse = layoutOption === 'image-right'
 
   return (
-    <BreakBlockWrapper className={className}>
+    <BreakBlockWrapper className={className} reverse={reverse}>
       <BreakBlockSegment reverse={reverse}>
         {image && <Image image={image} square={squareImage} />}
       </BreakBlockSegment>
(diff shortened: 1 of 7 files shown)


libs/block-content/website/src/lib/break/break-block.tsx
```diff
@@ -5,42 +5,36 @@ import {Block, LinkPageBreakBlock as LinkPageBreakBlockType} from '@wepublish/we
...
 
   ${({theme}) => theme.breakpoints.up('md')} {
-    grid-auto-flow: ${props => props.reverse && 'column'};
-    gap: ${({theme}) => theme.spacing(20)};
-    grid-template-columns: repeat(2, 1fr);
-  }
-
-  ${({theme}) => theme.breakpoints.up('lg')} {
-    gap: ${({theme}) => theme.spacing(30)};
+    grid-template-columns: 1fr 2fr;
```
Itrulia, line 14: having the text 2x the size of the image made it look a lot better

libs/block-content/website/src/lib/break/break-block.tsx
```diff
@@ -5,42 +5,36 @@ import {Block, LinkPageBreakBlock as LinkPageBreakBlockType} from '@wepublish/we
...
   display: grid;
-  width: 100%;
   align-items: center;
-  ${({theme}) => theme.breakpoints.up('md')} {
-    grid-column-end: ${props => props.reverse && '-1'};
-  }
-`
+  gap: ${({theme}) => theme.spacing(2)};
 
-const headingStyles = css`
```
Itrulia, line 34: Is now done via grid gap and using the theme

libs/block-content/website/src/lib/break/break-block.tsx
```diff
@@ -5,42 +5,36 @@ import {Block, LinkPageBreakBlock as LinkPageBreakBlockType} from '@wepublish/we
...
-  margin-bottom: 24px;
+  ${({theme, reverse}) =>
+    reverse &&
+    css`
+      ${theme.breakpoints.up('md')} {
+        order: 1;
```
Itrulia, line 27: Reversing can be done by using order

libs/block-content/website/src/lib/break/break-block.tsx
```diff
@@ -50,12 +44,12 @@ export const BreakBlock = ({
...
   return (
-    <BreakBlockWrapper className={className} reverse={reverse}>
-      <BreakBlockSegment reverse={reverse}>
-        {image && <Image image={image} square />}
-      </BreakBlockSegment>
+    <BreakBlockWrapper className={className}>
+      <BreakBlockSegment reverse={reverse}>{image && <Image image={image} />}</BreakBlockSegment>
+
       <BreakBlockSegment>
-        <H4 css={headingStyles}>{text}</H4>
```
Itrulia, line 58: H4 was semtantically wrong there, as it can't be an article semantically, i added the component `div`.

Itrulia:
It's just to keep things more "consistent" in a way, was that Tsri design or Gruppetto?

The problem with square was, that the image is bigger than the entire viewport: 
image

The problem with the big gaps were that on smaller resolutions, the gap was half as big as the content itself:
image


Now with these changes the image isn't as overpowering for smaller screens:
image

The gap isn't as bad anymore, but the image looks a bit out of place. this could be fixed by using the `useMedia` I guess and  changing it to be square once it hits a certain threshold. Should I change that @antkiewiczk ?

image

libs/block-content/website/src/lib/break/break-block.tsx
antkiewiczk: I couldn't make it work for some reason, and how I did it I found somewhere on stack overflow. And actually many people wrote that the way I did it is the only way 😆

antkiewiczk:
@Itrulia yeah I think square on smaller resolutions would actually look better. Thanks!

PR #1027, opened 2023-08-16
Improved periodicjob debugging

elias-summermatter:


michael-scheurer (approved):

libs/mails/src/lib/mail-context.ts
```diff
@@ -69,8 +69,9 @@ export class MailContext implements MailContextInterface {
       templateData: data
     })
   }
-
-  async getUserTemplateName(event: UserEvent): Promise<string> {
+  async getUserTemplateName(event: UserEvent, throwOnMissing: true): Promise<string>
```
Itrulia, line 72: dont think this one is needed as the other one is identical but can also return Promise<null>

PR #1030, opened 2023-08-22
feat(website-api): setup all membership/subscription/invoice queries/…

Itrulia:
Also

    1. Removed unnecessary cache operations in comment list
    2. Adjusted apollo parameter for storybook to use network only but still updates on cache update
    3. Removed useEffects that don't add anything valueable as one can use apollo with cache-only option
    4. Added an auth guard with intended routing
    5. Implemented url on public subscription
    6. Added new endpoint to pay by a subscription, not invoice


libs/api/src/lib/graphql/invoice.ts
```diff
@@ -70,9 +71,16 @@ export const GraphQLPublicInvoice = new GraphQLObjectType<InvoiceWithItems, Cont
 
     description: {type: GraphQLString},
     paidAt: {type: GraphQLDateTime},
+    dueAt: {type: new GraphQLNonNull(GraphQLDateTime)},
```
Itrulia, line 74: its set as non nullable in the DB schema

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -925,6 +925,51 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
       }
     },
 
+    createPaymentFromSubscription: {
```
Itrulia, line 928: To mimic bajour dashboard functionality

libs/api/src/lib/graphql/query.public.ts
```diff
@@ -73,7 +73,7 @@ import {getPoll, userPollVote} from './poll/poll.public-queries'
 import {GraphQLSetting} from './setting'
 import {getSetting, getSettings} from './setting/setting.public-queries'
 import {GraphQLSlug} from './slug'
-import {GraphQLPublicSubscription} from './subscription'
```
Itrulia, line 76: split into its own file as else it would have been a circular dependency

libs/article/website/src/lib/article-container.stories.tsx
```diff
@@ -2297,7 +2297,6 @@ export default {
 
 export const ById = {
   args: {
-    onQuery: action('onQuery'),
```
Itrulia, line 2300: removed all `onQuery` as they didn't work as intended, idea was to move to render props in the future, but I don't know if that even makes sense. One can get all that information by using apollo in cache only fetch mode.

libs/storybook/src/lib/parameters.tsx
```diff
@@ -5,23 +5,23 @@ import {possibleTypes} from '@wepublish/website/api'
 
 const defaultOptions: DefaultOptions = {
   watchQuery: {
-    fetchPolicy: 'no-cache',
```
Itrulia, line 8: no-cache means it never uses the cache, network-only means the initial data is from the network and then uses the cache.

We update the cache, to test this I switched to network-only which makes the most sense to be begin with in my opinion.

libs/storybook/src/lib/parameters.tsx
```diff
@@ -5,23 +5,23 @@ import {possibleTypes} from '@wepublish/website/api'
...
+    fetchPolicy: 'network-only',
     errorPolicy: 'all'
   }
 }
 
 const cache = new InMemoryCache({
   possibleTypes: possibleTypes.possibleTypes,
   resultCaching: false,
-  addTypename: false,
```
Itrulia, line 24: set to true so that the cache has the __typename, else it won't work in mocks.

libs/comments/website/src/lib/comment-list/comment-list-container.tsx
```diff
@@ -275,42 +212,3 @@ const useAddCommentMutationWithCacheUpdate = (
       })
     }
   })
-
-const useEditCommentMutationWithCacheUpdate = (
```
Itrulia, line 279: Apollo does this automatically as they are using the same fragment with the same id.

antkiewiczk:
isn't it possible to somehow split it into multiple, smaller PRs? 🤔 😅

antkiewiczk (dismissed):

libs/website/builder/src/lib/website-builder.context.tsx
```diff
@@ -84,14 +87,11 @@ import {
 
 const NoComponent = () => null
 
-export type WebsiteBuilderComponents = {
+export type WebsiteBuilderProps = {
```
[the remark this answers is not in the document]
Itrulia: Felt weird to have it called Components when there's now more than just components there.

apps/website-example/pages/_app.tsx
```diff
@@ -89,12 +89,16 @@ const NavBar = styled(NavbarContainer)`
   background-color: ${({theme}) => theme.palette.common.white};
 `
 
-function CustomApp({Component, pageProps}: AppProps) {
+type CustonAppProps = AppProps<{
```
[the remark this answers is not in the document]
Itrulia: Thank you!

libs/membership/website/src/lib/formatters/format-currency.ts
```diff
@@ -0,0 +1,20 @@
+import {fixCHNumberDelimiter} from './format-number'
+
+export const monthlyToQuarterly = (value: number) => Math.ceil((value / 100) * 4)
+export const monthlyToBiAnual = (value: number) => Math.ceil((value / 100) * 6)
```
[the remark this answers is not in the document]
Itrulia: Thank you!

PR #1036, opened 2023-08-29
Streamline dockerfile, use cache

tomaszdurka:


Dockerfile
```diff
@@ -1,18 +1,29 @@
...
-RUN npm run build
+FROM node:16-alpine3.17 as builder
+WORKDIR /build
+COPY . .
+RUN rm -rf .env*
+COPY --from=dependencies /dependencies .
+RUN npm run build --prod
 
+
+FROM node:16-alpine3.17
```
Itrulia, line 22: couldnt we use a slim package here?

Itrulia: We also miss cleaning up the apt-get cache, should be a few hundred MBs too, no?

Dockerfile
```diff
@@ -1,18 +1,29 @@
...
-RUN npm run build
+FROM node:16-alpine3.17 as builder
+WORKDIR /build
+COPY . .
+RUN rm -rf .env*
+COPY --from=dependencies /dependencies .
+RUN npm run build --prod
 
+
+FROM node:16-alpine3.17 as runtime
```
Itrulia, line 22: Using `FROM gcr.io/distroless/nodejs:16` here would save 80MB

FAQ.md
Itrulia, line 1: Hm those changes look not quite right

Dockerfile
tomaszdurka: > couldnt we use a slim package here?

Size was 2mb diff and actually npm ci command took longer 250 slim vs 150 alpine secs locally...

Dockerfile
tomaszdurka: > Using `FROM gcr.io/distroless/nodejs:16` here would save 80MB

```
#12 0.276 container_linux.go:380: starting container process caused: exec: "/bin/sh": stat /bin/sh: no such file or directory
------
executor failed running [/bin/sh -c npm ci]: exit code: 1
```

I get this on distroless so wont't go there atm.

FAQ.md
tomaszdurka: That was 1 liner change to trigger build ...
Reformatted whole thing too - reverting.

elias-summermatter:
@tomaszdurka  Do we still need this? Merge or close.

Itrulia:
@elias-summermatter  still something we want to do and got stuck because of `bcrypt` as `bcrypt` doesn't allow us to bundle the whole application. I have a few other things I wanna try but didn't get to do it yet.

PR #1041, opened 2023-09-11
feat(example-api/api): added api example seeding for PRs, local etc.

Itrulia:
Also fixed issue where anonymous commenting was broken


apps/api-example/project.json
```diff
@@ -67,7 +67,7 @@
       "options": {
         "commands": [
           {
-            "command": "npx ts-node apps/api-example/prisma/seed.ts"
```
Itrulia, line 70: tsx is better than node-ts

libs/api/src/lib/graphql/comment/comment.public-mutation.ts
```diff
@@ -49,7 +49,7 @@ export const addPublicComment = async (
     throw new CommentLengthError(+maxCommentLength)
   }
 
-  const canSkipApproval = hasPermission(CanCreateApprovedComment, user.roles)
```
Itrulia, line 52: fixes anonymous commenting as user was undefined

antkiewiczk (approved):

PR #1042, opened 2023-09-11
refactor(website-example): performance, image and seo improvements

Itrulia:
A few bug fixes too


apps/website-example/next.config.js
```diff
@@ -1,7 +1,10 @@
 //@ts-check
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
 const {composePlugins, withNx} = require('@nx/next')
+const withBundleAnalyzer = require('@next/bundle-analyzer')({
```
Itrulia, line 4: Allows us to see the bundle analyzer for prod builds

apps/website-example/public/robots.txt
Itrulia, line 1: for SEO

libs/author/website/src/lib/author-list/author-list-item.tsx
```diff
@@ -37,7 +37,7 @@ export function AuthorListItem({
   } = useWebsiteBuilder()
 
   return (
-    <Link css={wrapperStyles(theme)} className={className} href={slug}>
```
Itrulia, line 40: Bug fix as it was not pointing to the correct url

libs/block-content/website/src/lib/facebook/facebook-video-block.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled'
 import {Block, FacebookVideoBlock as FacebookVideoBlockType} from '@wepublish/website/api'
 import {BuilderFacebookVideoBlockProps} from '@wepublish/website/builder'
-import ReactPlayer from 'react-player'
```
Itrulia, line 4: smaller bundle

libs/feed/website/src/lib/feed-generator.ts
```diff
@@ -48,14 +48,15 @@ export const generateFeed =
       } as Item
     })
 
+    const Feed = (await import('feed')).Feed
```
Itrulia, line 51: The feed library is quite big and somehow webpack isn't able to properly treshake it out on the client. This fixes it.

libs/image/website/src/lib/image.tsx
```diff
@@ -24,10 +24,11 @@ export const imageToSquareImageItems = (image: FullImageFragment): ImageItem[] =
   {url: image.squareSmallURL, size: 200}
 ]
 
-export const ImageWrapper = styled('img')`
+export const ImageWrapper = styled('img')<{aspectRatio: number}>`
   max-width: 100%;
   width: 100%;
   height: auto;
+  aspect-ratio: auto ${({aspectRatio}) => aspectRatio};
```
Itrulia, line 31: Fixes image going haywire when the image can't load. As aspect-ratio by the user agent is only set when the image is loaded.

libs/richtext/src/lib/to-html.tsx
```diff
@@ -1,11 +1,12 @@
 import {Node} from 'slate'
-import {slateToDomConfig, slateToHtml} from 'slate-serializers'
 import {BlockFormat} from './block-format'
 import {InlineFormat} from './inline-format'
 import {TextFormat} from './text-format'
 
-export const toHtml = (nodes: Node[]) =>
-  slateToHtml(nodes, {
+export const toHtml = async (nodes: Node[]) => {
+  const {slateToHtml, slateToDomConfig} = await import('slate-serializers')
```
Itrulia, line 7: slate-serializers is quite a big package, no need to load it on every page even when this function isn't called.

libs/image/website/src/lib/image.tsx
antkiewiczk: shouldn't there be a default?

libs/block-content/website/src/lib/facebook/facebook-video-block.tsx
antkiewiczk: 👍

libs/website/builder/src/lib/website-builder.context.tsx
```diff
@@ -173,7 +173,7 @@ export type WebsiteBuilderProps = {
   }
 
   date: {
-    format: (date: Date) => string
+    format: (date: Date, includeTime?: boolean) => string
```
antkiewiczk, line 176: 👍

antkiewiczk (approved):

libs/image/website/src/lib/image.tsx
Itrulia: This is the default

PR #1052, opened 2023-09-22
WPC-1347, WPC-1350: Events bug fixes

antkiewiczk:


apps/api-example/schema-v2.graphql
Itrulia, line 1: You probably want to regenerate the client

Itrulia (approved):

PR #1053, opened 2023-09-26
WPC-1352: Events filters

antkiewiczk:


libs/api/src/lib/graphql/event/event.query.ts
```diff
@@ -129,12 +131,48 @@ const createToFilter = (filter?: Partial<EventFilter>): Prisma.EventWhereInput =
   return {}
 }
 
+const createNameFilter = (filter?: Partial<EventFilter>): Prisma.EventWhereInput => {
+  if (filter?.name) {
+    return {
+      OR: [
+        {
+          name: {
+            contains: filter.name,
```
Itrulia, line 140: Could full text search in these maybe work better than contains?

antkiewiczk: if you mean `search` instead of `contains`, then it didn't seem to work better - it seemed like it searched by the beginning of the word

Itrulia (approved):

PR #1054, opened 2023-09-26
feat(membership-website): implemented subscribe

Itrulia:


libs/navigation/website/src/lib/footer/footer-container.stories.tsx
Itrulia, line 1: Uses now the same data as the navigation instead of duplicating the data

apps/website-example/pages/mitmachen.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {NextPageContext} from 'next'
+import getConfig from 'next/config'
+import {ssrAuthLink} from '../src/auth-link'
+import {getSessionTokenProps} from '../src/get-session-token-props'
+
+export default function Mitmachen() {
+  return <SubscribeContainer failureURL="/" successURL="/" />
+}
+
+Mitmachen.getInitialProps = async (ctx: NextPageContext) => {
```
Itrulia, line 11: Using getInitialProps to preload user data when not already on the site, but not SSRing when the user is already on the site.

apps/website-example/pages/mitmachen.tsx
```diff
@@ -0,0 +1,45 @@
+import {ApiV1, SubscribeContainer} from '@wepublish/website'
+import {NextPageContext} from 'next'
+import getConfig from 'next/config'
+import {ssrAuthLink} from '../src/auth-link'
+import {getSessionTokenProps} from '../src/get-session-token-props'
+
+export default function Mitmachen() {
+  return <SubscribeContainer failureURL="/" successURL="/" />
```
Itrulia, line 8: Should probably make some real pages for these

apps/website-example/src/styles/tsri-article.styles.tsx
Itrulia, line 1: I don't remember making changes here, checking

PR #1055, opened 2023-09-27
Add bexio payment provider

elias-summermatter:


libs/api/src/lib/utility.ts
```diff
@@ -1,4 +1,4 @@
-import {MemberPlan, PaymentMethod} from '@prisma/client'
+import {MemberPlan, PaymentMethod, PaymentPeriodicity} from '@prisma/client'
```
antkiewiczk, line 1: why this change?

libs/payments/src/lib/payment-provider/bexioPaymentProvider.ts
```diff
@@ -0,0 +1,237 @@
+import Bexio, {ContactsStatic, InvoicesStatic} from 'bexio'
+import ContactSearchParameters = ContactsStatic.ContactSearchParameters
```
antkiewiczk, line 2: what's this syntax?

libs/payments/src/lib/payment-provider/bexioPaymentProvider.ts
```diff
@@ -0,0 +1,237 @@
...
+    ])
+    return contacts[0]
+  }
+
+  async createOrUpdateContact(
+    bexio: Bexio,
+    contact: ContactsStatic.ContactSmall,
+    user: User & {address: UserAddress}
+  ) {
+    const uppserContact: ContactsStatic.ContactOverwrite = {
```
antkiewiczk, line 162: do you mean upsert?

antkiewiczk (commented):
Please make sure that tests and linting is passing, also please delete the other bexio PR if it's not relevant anymore

libs/payments/src/lib/payment-provider/bexioPaymentProvider.ts
elias-summermatter: Yes wasn't finished yet but had a call and forgot to set it to do draft. But thanks, changed this.

libs/api/src/lib/utility.ts
elias-summermatter: Because i don't want to do code duplication and its used in multiple payment adapters. To keep it dry.

libs/payments/src/lib/payment-provider/bexioPaymentProvider.ts
elias-summermatter: An import Bexio is default export and ContactsStatic, InvoicesStatic are others. The secound line is a way to import  a type or an interface.

elias-summermatter:
@antkiewiczk now its ready for your review. Thanks :-)

Itrulia (commented):
Hey hey, I don't understand what Bexio is and how it's related to be a payment provider. Looking at the payment provider code this seems to be a hack to integrate bexio, which would be best solved via Nestjs modules.

Do we have a general use case for bexio? Or is it a specific newsroom that wants this?

elias-summermatter:
@Itrulia In my mind it is clearly a payment provider since it creates automatically invoices on renewal and when you get a new subscription. I dont see this as a usecase for a nestjs module or why do you think its better suited as module when we want to react to specific events in subscription process?
For now its just used at Hauptstadt but Bexio is very widely used in Switzerland so i'm sure it come in handy to other medias like Bajour who have members who want to pay by invoice.

Itrulia:
It creates invoices but not payments, clearly a payment provider should create payments?

tomaszdurka:
@antkiewiczk is this closable?

antkiewiczk:
> @antkiewiczk is this closable?

I think so, yes

PR #1060, opened 2023-09-29
Fix editor bugs

antkiewiczk:


package.json
```diff
@@ -283,7 +283,7 @@
     "recharts": "^2.5.0",
     "reflect-metadata": "^0.1.13",
     "regenerator-runtime": "0.13.7",
-    "rsuite": "^5.10.0",
+    "rsuite": "^5.40.0",
```
antkiewiczk, line 286: updating rsuite helped for the SelectPicker not working with async data. I was worried that it may break other stuff though, but I tested it and I didn't find any other issues

PR #1068, opened 2023-10-17
fix(website): missing exports, file case, subscribe filter, personal data form validation

Itrulia:


libs/article/website/src/lib/article-list/article-list-container.tsx
```diff
@@ -21,20 +23,19 @@ export function ArticleListContainer({
     variables
   })
 
+  const filteredArticles = useMemo(
+    () =>
+      produce(data, draftData => {
```
Itrulia, line 28: `produce` allows to mutate draftData and it makes a new deep copy.

libs/image/website/src/lib/image.tsx
```diff
@@ -26,9 +26,12 @@ export const imageToSquareImageItems = (image: FullImageFragment): ImageItem[] =
 
 export const ImageWrapper = styled('img')<{aspectRatio: number}>`
   max-width: 100%;
-  width: 100%;
   height: auto;
   aspect-ratio: auto ${({aspectRatio}) => aspectRatio};
+  // Sets a max height for images so that they will not be too tall
+  // and makes them not distorted
+  max-height: 80lvh;
```
antkiewiczk, line 33: first time I see lv 😮 looks like it might solve some issues, especially on mobile

PR #1069, opened 2023-10-17
fix(api): peering being broken due to urls & graphql update

Itrulia:


libs/api/src/lib/context.ts
```diff
@@ -628,11 +622,11 @@ export async function contextFromRequest(
               )?.value as number) ||
               parseInt(process.env.PEERING_TIMEOUT_IN_MS as string) ||
               10 * 1000 // 10 Seconds timeout in  ms
-            const fetcher = createFetcher(peer.hostURL, peer.token, peerTimeout)
+            const fetcher = createFetcher(`${peer.hostURL}`, peer.token, peerTimeout)
```
Itrulia, line 625:
```suggestion
            const fetcher = createFetcher(`${peer.hostURL}/v1`, peer.token, peerTimeout)
```

PR #1070, opened 2023-10-19
refactor: improved storybook tests, diff snapshots instead, use jest emotion

Itrulia:


tomaszdurka:
@Itrulia Im closing this ticket as it still runs deployment. Probably we should not trigger deployments for drafts. Please reopen if want to continue work. Thanks for understanding ! 👍

PR #1071, opened 2023-10-19
Move settings to v2

antkiewiczk:


libs/settings/editor/src/apiClientv2.ts
Itrulia, line 1: this is now copied a few times, please re-use it.

libs/settings/api/src/lib/settings.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+  Queries
+ */
+  @Query(returns => [Setting], {
+    name: 'settingsList',
+    description: `
+      Returns all settings.
+    `
+  })
+  @Permissions(CanGetSettings)
+  settingsList(@Args('filter', {nullable: true}) filter: SettingFilter) {
```
Itrulia, line 20:
```suggestion
  settingsList(@Args('filter', {nullable: true}) filter?: SettingFilter) {
```

libs/settings/api/src/lib/settings.service.ts
```diff
@@ -0,0 +1,72 @@
...
+export class SettingsService {
+  constructor(private prisma: PrismaClient) {}
+
+  /*
+  Queries
+ */
+  async settingsList(filter?: SettingFilter): Promise<Setting[]> {
+    const data = await this.prisma.setting.findMany({
+      where: {
+        ...filter
```
Itrulia, line 16: Don't think we want to spread here for the filter but create them

libs/website/.storybook/main.ts
```diff
@@ -10,6 +10,7 @@ export default {
   },
   stories: ['../../**/src/lib/**/*.mdx', '../../**/src/lib/**/*.stories.@(js|jsx|ts|tsx)'],
   addons: [
+    // eslint-disable-next-line storybook/no-uninstalled-addons
```
antkiewiczk, line 13: https://github.com/nrwl/nx/issues/16629

libs/settings/api/src/lib/settings.resolver.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+import {GraphQLModule} from '@nestjs/graphql'
+import {PrismaClient} from '@prisma/client'
+import {ApolloDriverConfig, ApolloDriver} from '@nestjs/apollo'
+import {PrismaModule} from '@wepublish/nest-modules'
+import {SettingsResolver} from './settings.resolver'
+import {SettingsService} from './settings.service'
+import {GraphQLSettingValueType} from './settings.model'
+import {SettingName} from './setting'
+
+export const generateRandomString = () => crypto.randomBytes(20).toString('hex')
```
Itrulia, line 14: Move to @wepublish/testing please

libs/settings/api/src/lib/settings.resolver.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+describe('SettingsResolver', () => {
+  let app: INestApplication
+  let prisma: PrismaClient
+
+  beforeAll(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      imports: [AppModule]
+    }).compile()
+
+    prisma = module.get<PrismaClient>(PrismaClient)
```
Itrulia, line 85: I think we said we wanted to mock the DB (unlike what we have now), right @tomaszdurka ?

libs/api/src/lib/utility.ts
```diff
@@ -5,10 +5,8 @@ import {delegateToSchema, IDelegateToSchemaOptions, Transform} from '@graphql-to
 import {ExecutionResult} from '@graphql-tools/utils'
```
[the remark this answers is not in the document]
antkiewiczk: let me know if it's ok now, I moved it to libs/editor

libs/settings/api/src/lib/settings.model.ts
```diff
@@ -0,0 +1,117 @@
...
+  settingRestriction?: SettingRestriction
+}
+
+@InputType()
+export class SettingFilter {
+  @Field({nullable: true})
+  name?: string
+}
+
+@InputType()
```
Itrulia, line 110:
```suggestion
@ArgsType()
```

This would make the API surface flat.

libs/settings/api/src/lib/settings.model.ts
```diff
@@ -0,0 +1,117 @@
...
+      case 'StringValue':
+        return ast.value
+      case 'BooleanValue':
+        return ast.value
+      case 'IntValue':
+        return parseInt(ast.value, 10)
+      case 'FloatValue':
+        return parseFloat(ast.value)
+      default:
+        throw new Error(`Value scalar error: cannot handle kind: ${ast.kind}`)
```
Itrulia, line 41:
```suggestion
        throw new BadRequestException(`Value scalar error: cannot handle kind: ${ast.kind}`)
```

libs/settings/api/src/lib/settings.model.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+}
+// @Scalar('Value', type => GraphQLSettingValueType)
+// export class GraphQLSettingValueType implements CustomScalar<string, any> {
```
Itrulia, line 46: Can be deleted

libs/settings/api/src/lib/settings.resolver.ts
```diff
@@ -0,0 +1,48 @@
+import {Args, Mutation, Query, Resolver} from '@nestjs/graphql'
+import {CanGetSettings, CanUpdateSettings, Permissions} from '@wepublish/permissions/api'
+import {Setting, UpdateSettingInput, SettingFilter} from './settings.model'
+import {SettingsService} from './settings.service'
+
+@Resolver()
+export class SettingsResolver {
+  constructor(private settingsService: SettingsService) {}
+
+  /*
```
Itrulia, line 10: Can we remove these comments? WE see that in the line below already :D

libs/settings/api/src/lib/settings.service.ts
```diff
@@ -0,0 +1,70 @@
+import {Injectable, NotFoundException} from '@nestjs/common'
+import {Prisma, PrismaClient, Setting} from '@prisma/client'
+import {UpdateSettingInput, SettingFilter, SettingRestriction} from './settings.model'
+import {checkSettingRestrictions} from './settings-utils'
+
+@Injectable()
+export class SettingsService {
+  constructor(private prisma: PrismaClient) {}
+
+  /*
```
Itrulia, line 10: idem

libs/settings/api/src/lib/settings.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+  @Permissions(CanGetSettings)
+  setting(@Args('id') id: string) {
+    return this.settingsService.setting(id)
+  }
+
+  /*
+  Mutations
+ */
+  @Mutation(returns => [Setting], {
+    name: 'updateSettings',
```
Itrulia, line 39: Do we want to only support updating 1 setting at the time and let the frontend iterate over it? Makes the API cleaner IMO.

The frontend can batch the requests in GraphQL.

antkiewiczk: that's how it was done in the past, I simply moved it to nest

antkiewiczk: ok @Itrulia I changed the resolver to support 1 setting at the time and I call it in batch on the frontend as suggested. This way I could use the @ArgsType() and it's cleaner

Itrulia (approved):

PR #1076, opened 2023-10-26
WPC-1377: Bexio -> add-bexio-payment-provider, including subscriptions

antkiewiczk:
- [x] - extract utils and errors
- [x] - refactor code, add logger and handle more errors
- [x] - add jsdoc
- [x] - save bexio invoice id as intentId
- [x] - finish checkIntentStatus function
- [x] - write tests
- [ ] - write documentation how to test bexio payment adapter


libs/payments/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "module": "commonjs"
   },
+  "target": "ES2018",
```
Itrulia, line 6: Should be in the tsconfig.lib.json and probably be es2021 looking at the other libraries

libs/payments/src/lib/payment-provider/bexio/bexioUtils.ts
```diff
@@ -0,0 +1,47 @@
+import {MemberPlan, PaymentState, Subscription, User} from '@prisma/client'
+import Bexio, {ContactsStatic} from 'bexio'
+import {MappedReplacer} from 'mapped-replacer/dist/types'
+
+const {ContactSearchParameters} = ContactsStatic
+
+export type BexioInvoiceStatus = 7 | 8 | 9 | 16 | 19 | 31
```
Itrulia, line 7: Could use an Enum and then give them names, makes the status clearer and no need for comments in the switch statements

antkiewiczk: that was my thought, but in enum a member cannot have a numeric name

PR #1093, opened 2023-11-14
refactor(consent): use @ArgType instead of @InputType

Itrulia:
Keeps the API cleaner & more generalized

Also added seeding to not utilize api-example seeding


libs/consent/api/__tests__/prisma/seed.ts
```diff
@@ -0,0 +1,248 @@
...
+      email: 'editor@wepublish.ch',
+      emailVerifiedAt: new Date(),
+      name: 'Editor User',
+      active: true,
+      roleIDs: [editorUserRole.id],
+      password: await hashPassword('123')
+    }
+  })
+
+  const mailTemplates = []
```
Itrulia, line 45: Note to self: Remove everything below

antkiewiczk (approved):

PR #1098, opened 2023-11-14
feat(api-example): subscription changes for api-example

Itrulia:


apps/api-example/src/app/slack-mail-provider.ts
```diff
@@ -20,8 +14,8 @@ export class SlackMailProvider extends BaseMailProvider {
     this.webhookURL = props.webhookURL
   }
 
-  async webhookForSendMail({req}: WebhookForSendMailProps): Promise<MailLogStatus[]> {
-    return Promise.resolve([])
+  async webhookForSendMail() {
+    return []
```
antkiewiczk, line 18: should it stay like this?

apps/api-example/src/nestapp/app.module.ts
```diff
@@ -84,7 +156,7 @@ import bodyParser from 'body-parser'
                 config.get('BEXIO_INVOICE_MAIL_BODY_RENEW') ||
                 'Hello :user.firstname:\n\nThank you for subscribing to :memberPlan.name:.\nYou can view your invoice here: [Network Link]\n\nBest wishes from the Wepublish team',
               markInvoiceAsOpen: false,
-              prisma
+              prisma: prisma
```
antkiewiczk, line 159: just `prisma` should be enough

antkiewiczk (approved):

apps/api-example/src/app/slack-mail-provider.ts
Itrulia: Yeah, slack doesn't give webhooks for sent messages back

PR #1104, opened 2023-11-21
fix(website): use PNG instead of fake WEBP as it breaks safari

Itrulia:
The media server returns a PNG with a WEBP ending and not actually returning a PNG.
Unlike the other browsers, this breaks Safari in weird ways:

1. It gets insanely slow
2. It breaks the layout

### DIFF
--- a/package.json
+++ b/package.json
@@ -256,7 +256,7 @@
     "mapped-replacer": "^1.2.1",
     "matomo-tracker": "^3.0.0-beta.0",
     "nanoid": "^2.1.7",
-    "next": "~13.5.4",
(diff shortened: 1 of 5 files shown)


antkiewiczk (approved):

package.json
```diff
@@ -256,7 +256,7 @@
     "mapped-replacer": "^1.2.1",
     "matomo-tracker": "^3.0.0-beta.0",
     "nanoid": "^2.1.7",
-    "next": "~13.5.4",
```
Itrulia, line 259: Fixes bug of SWC emotion

PR #1092, opened 2023-11-14
WPC-1383-add-invoice-properties-to-mail

elias-summermatter:

PR #1108, opened 2023-11-22
feat(api,mail-api,membership-api): cherry pick from #1092

Itrulia:


libs/mail/api/src/lib/mail-provider/mailchimp-mail-provider.ts
```diff
@@ -52,17 +52,29 @@ function flattenObjForMandrill<T>(ob: T): Record<string, string> {
   for (const i in ob) {
     const nestedObj = ob[i]
 
-    if (nestedObj && typeof nestedObj === 'object' && !Array.isArray(nestedObj)) {
+    if (Array.isArray(nestedObj)) {
+      for (const j in nestedObj) {
+        if (nestedObj[j] && typeof nestedObj[j] === 'object') {
```
Itrulia, line 57: Added this as this fixes a bug of the other PR with `null`, `number` `string` arrays (or rather any non nested arrays or arrays of objects).

libs/membership/api/src/lib/periodic-job/periodic-job.service.ts
```diff
@@ -193,13 +194,18 @@ export class PeriodicJobService {
         daysAwayFromEnding
       })
 
+    const invoices = await this.prismaService.invoice.findMany({
```
Itrulia, line 197: Moved it ouside the loop as it was not affected by it

antkiewiczk (approved):

PR #1117, opened 2023-11-30
WPC:1414: Payrexx - Save transaction token; call Transaction API endpoint 

antkiewiczk:


elias-summermatter (changes requested):

libs/payment/api/src/lib/payrexx/payrexx-client.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  private buildBaseUrl(path) {
+    return this.baseUrl + path + '?instance=' + this.instance
+  }
+
+  async get<Data = any>(path, queryParams = {}): Promise<PayrexxResponse<Data>> {
+    const method = 'GET',
+      queryStrSigned = this.buildSignedQueryString(queryParams),
+      baseUrl = this.buildBaseUrl(path + '&' + queryStrSigned)
+    console.log({message: 'Payrexx GET request issued', path, queryParams})
```
antkiewiczk, line 42: maybe we could use `logger()`

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
```diff
@@ -183,7 +229,8 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
     return {
       state,
       paymentID: gateway.referenceId,
-      paymentData: JSON.stringify(gateway)
+      paymentData: JSON.stringify(gateway),
+      customerID: transaction.preAuthorizationId?.toString()
```
antkiewiczk, line 233: so the idea is still to return the preAuthorizationId as customerID? I guess we would need to rewrite whole payment provider to change the naming convention etc for the tokenization?

antkiewiczk (commented):
looks good 👍 much cleaner

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
tomaszdurka: Yes, that would be good to do soon. There are so many entry points and abstraction is really poor.
Let's talk it over soon.

libs/payment/api/src/lib/payrexx/payrexx-client.ts
tomaszdurka: I don't fully support the logger idea here. It complicates things and keeps logs flat strings. Modern logging should support objects. arrays, etc.
Let's talk about it !

tomaszdurka:
### Summary of today/latest work:
There where was one issue with how the GET Payrexx request URL was constructed - fixed now.

I've tested all the paths I could:
- register member with subscription
- register and subscribe later
- subscribe to multiple memberplans
- creating multiple subscriptions for same memberplans

I've also tested offsite payments, which work now for all possible payment paths, once the customer-id is there. There were couple of wrongdoings around this process. It required passing user/customer through the messy code.

We can probably improve the way UI behaves when user has offsite payments and e.g. extend for another period manually - there is no feedback just redirect to successUrl (which could be the feedback).

### Todo
- Add authentication to webhooks
- Test webhooks
- Discuss if offsite payments should really happen for user actions

tomaszdurka:
## Another update
Looked into web hooks implementation. Couple of things brought my attention e.g. invalid APIKEY would result 200 response to Payment Provider and basically ignore web hook.
Ended up reworking web hooks a bit so implementation support more custom approach for each provider allowing to have more granular response like 400, 403, 500 with custom message. This allows to look up unsuccessful responses in payment provider dashboard too.
image

 

### Rework changes summary:
- customise webhooks responses to better reflect processing result
- not always return 200 on webhook processing (alert/error only on unexpected cases, use payment provider retry mechanism)
- allow returning more complex results with status, optional message and intentStates to process
- implement required apiKey for payrexx webhooks, use query parameter

### Local tests
- Local tests worked fine using ngrok.
- New user test ended up with getting transaction `preAuthorizationId` for future payments and saving it in database in the background from webhook. Before user needed to visit member page - to autorun those checkIntent requests.
- Tested `apiKey` - works/blocks well. This actually started the whole response rework.

## Todo
- Try to secure Payrexx test instance for sandboxing;
- Deploy review, dev, next, demo with proper sandbox payrexx configuration and web hooks setup;
- Think about limiting processing of the web hooks to target instance - would require some arbitrary data in the payment itself.

tomaszdurka:
## Update 9-01-2024

### Webhooks

It took moment to understand what needs to be configured to enable sandbox subscriptions on payrexx. Connecting our stripe account worked out. I've created paylink for subscriptions to initiate transaction with subscriptions. Tested out with two active web hooks. After applying below changes everything works well.

Differentiate events for payrexx web hooks
- Skip non-transaction events (in case payrexx admin sends all events)
- Process only transactions with no subscription in PayrexxProvider
- Process only transactions with subscription in PayrexxSubscriptionProvider
- Always pass informative message in webhook response body

image



### Setting up Payrexx for all instances

- I've async contacted Payrexx to keep the instance I am using for all the tests. Waiting for reply.
- Prepared configuration to the subject instance for:
  - Review deployments
  - Dev instance
  - Next instance
  - Demo instance

### Multiple instances + webhooks
Investigated what can be done to execute webhooks only on the origin instance. So webhooks originated from transactions from `next` instance should be processed only on `next` instance. In order to do that we would need to pass some arbitrary data into the transaction. Once choice is prefixing/suffixing order in referenceId. Did that in other projects - bit hacky.

## Todo
- Get opinions on limiting processing of the web hooks to target instance - would require some arbitrary data in the payment itself;
- Continue deployments changes - setup webhooks for all instances, test payment providers;
- Seed payment providers
- Look into failing Review Database deployment (not related, but blocking).

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -618,26 +619,25 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
...
-              paymentMethodID: subscription.paymentMethodID,
-              customer
-            })
-            if (payment) {
-              return payment
+          } else {
+            // Charge customer
+            try {
+              const payment = await memberContext.chargeInvoice({
+                user,
```
antkiewiczk, line 626: where was the user taken from previously? 🤔 but looks like a good change

libs/api/src/lib/payments.ts
```diff
@@ -14,31 +14,41 @@ export function setupPaymentProvider(opts: WepublishServerOpts): Router {
...
-              paymentsByID: context.loaders.paymentsByID,
-              invoicesByID: context.loaders.invoicesByID,
-              subscriptionClient: context.prisma.subscription,
-              userClient: context.prisma.user,
-              invoiceClient: context.prisma.invoice,
-              subscriptionPeriodClient: context.prisma.subscriptionPeriod,
-              invoiceItemClient: context.prisma.invoiceItem
-            })
+                for (const paymentStatus of webhookResponse.paymentStates) {
+                  // TODO: handle errors properly
```
antkiewiczk, line 30: does this comment still need to be here?

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
```diff
@@ -5,170 +5,237 @@ import {
...
-    }
-
     return {
-      intentID: payrexxResponse.data?.[0].id,
-      intentSecret: payrexxResponse.data?.[0].link,
-      intentData: JSON.stringify(payrexxResponse.data),
+      intentID: gateway.id.toString(),
+      intentSecret: gateway.link,
+      intentData: JSON.stringify(gateway),
       state: PaymentState.submitted
```
antkiewiczk, line 188: no need to return preauthorizationId aka customerID here?

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
```diff
@@ -183,7 +250,27 @@ export class PayrexxPaymentProvider extends BasePaymentProvider {
...
       paymentID: gateway.referenceId,
-      paymentData: JSON.stringify(gateway)
+      paymentData: JSON.stringify(gateway),
+      customerID: transaction?.preAuthorizationId
+        ? transaction.preAuthorizationId.toString()
+        : undefined
+    }
+  }
+
+  private mapPayrexxEventToPaymentStatus(
```
antkiewiczk, line 260: these kind of private helper methods should be at the top IMO, before the actual logic

libs/payment/api/src/lib/payment-provider/payrexx-subscription-payment-provider.ts
```diff
@@ -369,15 +370,32 @@ export class PayrexxSubscriptionPaymentProvider extends BasePaymentProvider {
     }
   }
 
-  async webhookForPaymentIntent(props: WebhookForPaymentIntentProps): Promise<IntentState[]> {
+  async webhookForPaymentIntent(props: WebhookForPaymentIntentProps): Promise<WebhookResponse> {
     const intentStates: IntentState[] = []
 
     // Protect endpoint
     const apiKey = props.req.query.apiKey as string
-    if (!timeConstantCompare(apiKey, this.webhookSecret)) throw new Error('Invalid api key!')
```
antkiewiczk, line 377: this function is now reused in multiple files, can be extracted to utils

libs/payment/api/src/lib/payment-provider/stripe-payment-provider.ts
```diff
@@ -64,17 +65,20 @@ export class StripePaymentProvider extends BasePaymentProvider {
     return this.stripe.webhooks.constructEvent(body, signature, this.webhookEndpointSecret)
   }
 
-  async webhookForPaymentIntent(props: WebhookForPaymentIntentProps): Promise<IntentState[]> {
+  async webhookForPaymentIntent(props: WebhookForPaymentIntentProps): Promise<WebhookResponse> {
```
antkiewiczk, line 68: 👌

libs/payment/api/src/lib/payrexx/gateway-client.ts
```diff
@@ -0,0 +1,73 @@
...
+type CreateGatewayResponseData = Array<Gateway>
+type GetGatewayResponseData = Array<Gateway>
+
+export class GatewayClient {
+  constructor(private readonly client: PayrexxClient) {}
+
+  async createGateway(requestData: CreateGatewayRequestData) {
+    const response = await this.client.post<CreateGatewayResponseData>('Gateway', requestData)
+    if (response.status === 'error') {
+      throw new Error(`Payrexx request has error status with message: ${response.message}`)
```
antkiewiczk, line 50: maybe we can specify that the error is about the payrexx gateway, not just payrexx in general

libs/payment/api/src/lib/payrexx/gateway-client.ts
```diff
@@ -0,0 +1,73 @@
...
+  async getGateway(gatewayId: string) {
+    const response = await this.client.get<GetGatewayResponseData>(
+      'Gateway/' + encodeURIComponent(gatewayId)
+    )
+
+    if (response.status === 'error') {
+      if (response.message === 'An error occurred: No Gateway found with id ' + gatewayId) {
+        return null
+      } else {
+        throw new Error(`Payrexx request has error status with message: ${response.message}`)
```
antkiewiczk, line 68: maybe we can specify that the error is about the payrexx gateway, not just payrexx in general

libs/payment/api/src/lib/payrexx/transaction-client.ts
```diff
@@ -0,0 +1,76 @@
...
+  subscription: null | Subscription
+}
+
+type ChargeTransactionResponseData = Array<Transaction>
+type RetrieveTransactionResponseData = Array<Transaction>
+
+export class TransactionClient {
+  constructor(private readonly client: PayrexxClient) {}
+
+  async chargePreAuthorizedTransaction(
```
antkiewiczk, line 44: 👍

elias-summermatter:
@tomaszdurka and as discussed:
- If transcationId is invalid to create new transactionId and not just fail.

Itrulia (approved):

tomaszdurka:
Added tests for Payrexx payment provider.

```
  PayrexxPaymentProvider
    Webhooks
      ✓ should reject unauthorized (36 ms)
      ✓ should ignore non-transaction payloads (1 ms)
      ✓ should ignore payload with subscription (1 ms)
      ✓ should process correct payload
      ✓ should ignore transactions with unknown payment status
    Payments
      ✓ should create gateway for user without customerId (17 ms)
      ✓ should charge transaction for user with customerId (50 ms)
      ✓ should create gateway for user with customer id and unsuccessful transaction
    Payment status checks
      ✓ should throw on unmappable transaction status (37 ms)
      ✓ should throw on empty transaction referenceId (1 ms)
      ✓ should get intent state based on transaction id (1 ms)
      ✓ should throw on unmappable gateway status (1 ms)
      ✓ should throw on empty gateway referenceId (4 ms)
      ✓ should get intent state based on gateway id
      ✓ should throw if inten is not related to transaction or gateway (1 ms)
```

apps/api-example/src/nestapp/app.module.ts
```diff
@@ -191,23 +192,23 @@ import {SlackMailProvider} from '../app/slack-mail-provider'
...
               id: 'payrexx',
               name: 'Payrexx',
-              offSessionPayments: false,
-              instanceName: config.get('PAYREXX_INSTANCE_NAME'),
-              instanceAPISecret: config.get('PAYREXX_API_SECRET'),
+              offSessionPayments: true,
+              transactionClient: payrexxFactory.transactionClient,
+              gatewayClient: payrexxFactory.gatewayClient,
+              webhookApiKey: config.getOrThrow('PAYREXX_WEBHOOK_SECRET'),
               psp: [0, 15, 17, 2, 3, 36],
```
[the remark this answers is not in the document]
tomaszdurka: This is just example api. Didn't touch that tbh.

libs/api/src/lib/payments.ts
tomaszdurka: I don't know tbh, I don't think we handle payment errors here at all. Another work to do imo.

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
tomaszdurka: There is no customerId here in create gateway response.
This is creation of the gateway. Only after the gateway is paid we can check gateway transaction or use webhook information. Which we do.

libs/payment/api/src/lib/payrexx/gateway-client.ts
tomaszdurka: That's usually in the `response.message`. Might add it.

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
```diff
@@ -5,170 +5,237 @@ import {
...
+        'Payrexx gateway with ID: %s for paymentProvider %s returned with an unmappable status %s',
         gateway.id,
         this.id,
         gateway.status
       )
-      throw new Error('unknown gateway state')
+      throw new Error('Unmappable Payrexx gateway status')
+    }
+
+    const transaction = gateway.invoices[0]?.transactions[0]
```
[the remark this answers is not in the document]
tomaszdurka: I am not 100% sure with all the cases.

Actually there are two invoices with a transaction. Both transaction seem to have the same crucial data.
Tbh whole Payrexx API is based on arrays. All APIs like getGateway, getTransaciton or webhooks has array of things as payload, but can only have one.

Here is the jq of gateway I just paid.
```
node get-gateway.js | jq '.data[].invoices[] | {number, transaction: .transactions[] | {id, amount, preAuthorizationId}}'
{
  "number": "Sub 1",
  "transaction": {
    "id": 12792651,
    "amount": 2000,
    "preAuthorizationId": 12792651
  }
}
{
  "number": "Sub 1",
  "transaction": {
    "id": 12792652,
    "amount": 2000,
    "preAuthorizationId": 12792651
  }
}
```

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
tomaszdurka: So this is private method in the class, Im used to add private methods below public ones.
I agree helper functions in front of file.
Any thoughts @Itrulia @elias-summermatter?

libs/payment/api/src/lib/payment-provider/payrexx-subscription-payment-provider.ts
tomaszdurka: Ah, let me get back to it if we rework it.

tomaszdurka:
> @tomaszdurka and as discussed:
> 
> * If transcationId is invalid to create new transactionId and not just fail.

I don't understand what you mean here.

libs/payment/api/src/lib/payment-provider/payrexx-payment-provider.ts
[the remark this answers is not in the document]
elias-summermatter: Ok thats fine just wondered :-)

elias-summermatter:
@tomaszdurka what I meant is our discussion about what happens if the credit card doesn't work (no money etc.) that we get a new gateway to change the credit card.

elias-summermatter (approved):

PR #1123, opened 2023-12-08
WPC-1375: Fix remove image

antkiewiczk:


libs/membership/editor/src/lib/memberplan/memberplan-edit.tsx
```diff
@@ -50,6 +50,7 @@ function MemberPlanEdit() {
       onError: showErrors
     })
 
+  console.log('memberPlanData', memberPlanData)
```
Itrulia, line 53: probably want to remove that

antkiewiczk: removed, the other one too

PR #1127, opened 2024-01-02
WPC-1014: Linting refactor

antkiewiczk:
many warnings are still there, I've created this doc: https://docs.google.com/document/d/1iOjjYLCovEqaRTSQDyRqK8rz2TBbIclHpDWrOFtLHz8/edit


elias-summermatter:
@antkiewiczk is this ready for review? If no convert it to draft :-)

antkiewiczk:
> @antkiewiczk is this ready for review? If no convert it to draft :-)



> @antkiewiczk is this ready for review? If no convert it to draft :-)



> @antkiewiczk is this ready for review? If no convert it to draft :-)

yes, the failing tests are not related and I think there's some problem with deployment that's also not related to my linting changes

libs/image/website/src/lib/image.tsx
```diff
@@ -3,6 +3,7 @@ import {FullImageFragment} from '@wepublish/website/api'
 import {BuilderImageProps} from '@wepublish/website/builder'
 
 declare module 'react' {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
elias-summermatter, line 6: Here too

libs/ui/editor/src/lib/atoms/comment/commentDeleteBtn.tsx
```diff
@@ -37,6 +37,7 @@ export function CommentDeleteBtn({comment, onCommentDeleted}: CommentDeleteBtnPr
   })
 
   if (!comment) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
```
elias-summermatter, line 40: return null?

libs/ui/editor/src/lib/atoms/comment/commentHistory.tsx
```diff
@@ -27,6 +27,7 @@ function ChildComments({
   setRevision
 }: ChildCommentsProps) {
   if (!comments) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
```
elias-summermatter, line 30: here too

libs/ui/editor/src/lib/atoms/poll/pollExternalVotes.tsx
```diff
@@ -46,6 +46,7 @@ export function ExternalVoteTable({
 }: ExternalVoteTableProps) {
   const {t} = useTranslation()
   if (!poll?.externalVoteSources?.length) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
```
elias-summermatter, line 49: Here too

libs/ui/editor/src/lib/atoms/searchAndFilter/subscriptionListFilter.tsx
```diff
@@ -132,6 +132,7 @@ export function SubscriptionListFilter({
    */
   function resetFilterView() {
     if (!isAnyFilterSet()) {
+      // eslint-disable-next-line react/jsx-no-useless-fragment
```
elias-summermatter, line 135: Also here

libs/ui/editor/src/lib/atoms/user/editUserPassword.tsx
```diff
@@ -111,7 +111,10 @@ export function EditUserPassword({
 
   function resetPasswordModal() {
     const userId = user?.id
-    if (!userId) return <></>
+    if (!userId) {
+      // eslint-disable-next-line react/jsx-no-useless-fragment
```
elias-summermatter, line 115: Here as well

elias-summermatter (changes requested):

libs/api/__tests__/utility.ts
```diff
@@ -108,12 +108,6 @@ export async function createGraphQLTestClient(overwriteRequest?: any): Promise<T
   const prisma = new PrismaClient()
   await prisma.$connect()
 
-  const adminUser = await prisma.user.findUnique({
```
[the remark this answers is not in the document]
antkiewiczk: yes, it was unused

libs/image/website/src/lib/image.tsx
antkiewiczk: this comment is about <T> in react type definition. the T meant for generic type can be removed, but it's recommended to still use it while defining type. so two ways are possible here: so what I did, or just remove it. I prefer the way I did it because I feel it aligns better with typescript's best practices. If you wish, I can just remove it, both approaches are okay

libs/ui/editor/src/lib/atoms/comment/commentHistory.tsx
antkiewiczk: for fragments vs null - both approaches are okay and in the end the difference is marginal. But I feel that in places where we mean to return html, but we don't for some reason, it's better to return empty fragment than null as null is a different type than html element. 

In React with TypeScript, when deciding between returning null or an empty fragment (<></> or <React.Fragment></React.Fragment>) in the default case of a switch statement where you don't want to render anything, both approaches are technically correct, but they have subtle differences:

Returning null:

This is a common pattern in React for conditional rendering. When you return null, nothing gets rendered to the DOM in that part of the component tree. It signals that no component should be rendered.
It's more explicit in indicating that nothing should be rendered.
It can be more readable and clear, especially for developers who are new to React or reviewing your code.
Returning an empty fragment (<></> or <React.Fragment></React.Fragment>):

An empty fragment renders nothing to the DOM, similar to null. However, it maintains the component's presence in the component tree.
It can be useful if you are tracking the lifecycle of a component, as the component technically still exists even though it renders nothing.
It can also be useful if you anticipate that the component might need to render something in the future based on state changes or props.
Which one to choose? It largely depends on your specific use case and coding style preference. If you want to explicitly show that nothing should be rendered, use null. If you prefer keeping the component's presence or anticipate future changes, an empty fragment might be more suitable.

again, both approaches are okay and there's no bad answer. If you prefer for me to change it to null, I will.

libs/image/website/src/lib/image.tsx
elias-summermatter: No No you are more used to frontend stuff. If you (and the others reviewing it) think it's cleaner, I'm fine with that. :-)

libs/ui/editor/src/lib/atoms/comment/commentHistory.tsx
elias-summermatter: Here also whatever you and the rest of the Frontend devs think it's better :-)  But if we decide to leave it like this wouldn't it be clearer to disable the linting rule? Or do you think it gives us some extra value to explicit define "// eslint-disable-next-line react/jsx-no-useless-fragment"?
With the follwoing rule:
  "rules": {
    "react/jsx-no-useless-fragment": "off"
  }

apps/editor/src/app/panel/peerEditPanel.tsx
```diff
@@ -282,7 +282,10 @@ function PeerEditPanel({id, hostURL, onClose, onSave}: PeerEditPanelProps) {
                   {profile?.callToActionURL}
                 </DescriptionListItem>
                 <DescriptionListItem label={t('peerList.panels.callToActionImage')}>
-                  <img src={profile?.callToActionImage?.thumbURL || undefined} />
+                  <img
+                    src={profile?.callToActionImage?.thumbURL || undefined}
+                    alt="callToActionImage"
```
Itrulia, line 287:
```suggestion
                    alt="call to action image"
```

Ideally would be translated

libs/ui/editor/src/lib/atoms/comment/commentHistory.tsx
antkiewiczk: I would still keep the rule because often during development someone might omit that, For example. someone creates fragment, adds multiple elements inside, and then during development some stuff changes and he removes those multiple items inside fragment and just leaves one - it's easy to omit that there is still a fragment that shouldn't be there

apps/editor/src/app/panel/peerEditPanel.tsx
antkiewiczk: I changed it to translation

libs/author/website/src/lib/author.tsx
```diff
@@ -6,6 +6,7 @@ import {
 } from '@wepublish/website/builder'
 
 declare module 'react' {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
Itrulia: Would cause an error here:
image

libs/image/website/src/lib/image.tsx
Itrulia: image

It's required

antkiewiczk:
@elias-summermatter I added return type JSX.Element to all the functions so that Fragment now makes much more sense 😉

libs/membership/editor/src/lib/subscription-flow/timeline/timeline-body.tsx
```diff
@@ -46,7 +46,7 @@ export function TimelineBody({days, subscriptionFlow, eventIcons, eventColors}:
   }
 
   return (
-    <>
```
Itrulia, line 49: This fragment seems correct here and looks more like eslint unable to understand we return multiple here

libs/consent/api/.eslintrc.json
```diff
@@ -4,7 +4,7 @@
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
-      "rules": {}
+      "rules": {"@typescript-eslint/no-unused-vars": ["error", {"args": "none"}]}
```
Itrulia, line 7: Can we set this globally?

Itrulia (dismissed):

libs/membership/editor/src/lib/subscription-flow/timeline/timeline-body.tsx
antkiewiczk: yeah I know :D but I would still keep it as it is

libs/consent/api/.eslintrc.json
antkiewiczk: just did

elias-summermatter:
@antkiewiczk what is the state here? Do you need something here?

antkiewiczk:
> @antkiewiczk what is the state here? Do you need something here?

yes, re-review/approve 😄

Itrulia (approved):

PR #1128, opened 2024-01-04
feat: media server replacement idea

Itrulia:
Drop in replacement for new instances for the media server.

- Does not support file formats on purpose (Set to WebP right now, will be Avif when supported by all browsers) as there's no need to deliver outdated formats.
- Focus points is calculated into regions

Implements:

- [x] Existing image transformations
- [x] HTTP cache headers
- [x] S3 Bucket
- [x] Seperate bucket for file transformations  
- [ ] 30 day file policy
- [ ] Migration from old media server 
- [x] ~~Twitter card~~ (Twitter now displays the title on the image automatically)


libs/api/src/lib/media/novaMediaAdapter.ts
Itrulia, line 1: nova just because i needed a name

apps/media/src/app/app.service.ts
```diff
@@ -0,0 +1,129 @@
+import {Injectable} from '@nestjs/common'
+import {StorageClient} from '@wepublish/media/api'
+import sharp from 'sharp'
+import {TransformationsDto} from './transformations.dto'
+import {Readable} from 'stream'
+
+const getTransformationKey = (transformations: TransformationsDto) => {
```
Itrulia, line 7: sorts the object keys and turns it into json. Sorting is needed so we don't have the same transformations multiple time just with a different sorting.

apps/media/src/app/app.service.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+@Injectable()
+export class AppService {
+  constructor(private storage: StorageClient) {}
+
+  public async getData(imageId: string, transformations: TransformationsDto) {
+    const transformationsKey = getTransformationKey(transformations)
+    const isAlreadyTransformed = await this.storage.hasFile(
+      'wepublish-transformed',
```
Itrulia, line 27: Prob should be moved to a const or env var.

apps/media/src/app/app.service.ts
```diff
@@ -0,0 +1,129 @@
...
+    ])
+  }
+
+  public async saveImage(imageId: string, image: Buffer) {
+    const sharpInstance = sharp(image, {
+      animated: true
+    })
+
+    await this.storage.saveFile(
+      'wepublish-staff',
```
Itrulia, line 115: Prob should be moved to a const or env var.

apps/media/src/app/app.module.ts
```diff
@@ -0,0 +1,35 @@
...
+@Module({
+  imports: [
+    StorageClientModule.registerAsync({
+      imports: [ConfigModule],
+      useFactory: (config: ConfigService) => ({
+        endPoint: config.getOrThrow('S3_ENDPOINT'),
+        port: +config.getOrThrow('S3_PORT'),
+        accessKey: config.getOrThrow('S3_ACCESS_KEY'),
+        secretKey: config.getOrThrow('S3_SECRET_KEY'),
+        useSSL: Boolean(config.get('S3_SSL'))
```
antkiewiczk, line 19: I think there's no such value (S3_SSL) in the env

antkiewiczk: ah I just noticed the `Boolean()` thing :D

libs/api/src/lib/media/novaMediaAdapter.ts
```diff
@@ -0,0 +1,148 @@
...
+    // Related issue: https://github.com/form-data/form-data/issues/394
+    form.hasKnownLength = () => false
+
+    const response = await fetch(this.internalURL, {
+      method: 'POST',
+      headers: {authorization: `Bearer ${this.token}`},
+      body: form
+    })
+
+    const json = await response.json()
```
antkiewiczk, line 38: maybe some other more descriptive name would be better? like parsed, responseAsJson, parsedResponse or something? it would help to understand better on line 44 as well

apps/media/src/app/app.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+      })
+    )
+    uploadedFile: Express.Multer.File
+  ) {
+    const imageId = uuidv4()
+    const metadata = await this.appService.saveImage(imageId, uploadedFile.buffer)
+
+    res.status(201).send({
+      id: imageId,
+      filename: `${imageId}.webp`,
```
antkiewiczk, line 47: if we want to use webp for now, but move to avif in the future, maybe the value could be parametrized to make it easier and less error prone to switch?

antkiewiczk (commented):
looks good 👍

libs/api/src/lib/media/novaMediaAdapter.ts
Itrulia: Was just c&p from the old one haha :D Can change that and clean that up yeah. Didn't really clean anything up yet in general

apps/media/src/app/app.controller.ts
Itrulia: Avif will be the output for later, I think it makes sense to have everything in webp (or avif to begin with) and only modify the what we return. This is for the DB entry that shows the filename. Not sure we even need this in the editor and maybe can drop it?

apps/media/src/app/app.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  @Get(':imageId')
+  async transformImage(
+    @Res() res: Response,
+    @Param('imageId') imageId: string,
+    @Query() transformations: TransformationsDto
+  ) {
+    const [file, stats] = await this.appService.getData(imageId, transformations)
+    res.setHeader('Content-Type', 'image/webp')
+    // res.setHeader('ETag', stats.etag)
```
Itrulia, line 65: Had that disabled because of debugging an issue :D

PR #1135, opened 2024-01-10
Fix deployment issues with Helm database deploy

tomaszdurka:


tomaszdurka:
## Summary

Actually the reason of database releases failing was too low amount of cpu on the cluster.
- Downgraded lerna to v3
- Made sure uninstall will run for all active releases (app, database, media-server) also if only some releases are present.

PR #1137, opened 2024-01-11
feat(events-api): moved events API to v2

Itrulia:
No deletion of v1 & migration of editor until image API has been fully moved


libs/utils/api/src/lib/dataloader/prime-dataloaders.decorator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export function PrimeDataLoader<T extends Primeable<unknown>>(dataloader: Type<T>) {
+  const decoratorFactory = (
+    target: object,
+    propertyKey: unknown,
+    descriptor: PropertyDescriptor
+  ) => {
+    const origin = descriptor.value
+    const injectDataloader = Inject(dataloader)
+    injectDataloader(target, `__DATALOADER__${dataloader.name}`)
```
Itrulia, line 19: Tells nestjs dependency injector to inject the dataloader into the class.

libs/utils/api/src/lib/dataloader/prime-dataloaders.decorator.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+
+      const results = Array.isArray(resultItem)
+        ? resultItem
+        : Array.isArray(resultItem.nodes)
+        ? resultItem.nodes
+        : [resultItem]
+
+      for (const result of results) {
+        if ('id' in result) {
```
Itrulia, line 35: If we would use validation schemas and generate the graphql model out of it, we could validate against that

apps/api-example/schema-v2.graphql
```diff
@@ -47,29 +47,86 @@ scalar DateTime
...
+input EventFilter {
+  from: DateTime
+  location: String
+  name: String
+  tags: [String!]
+  to: DateTime
+  upcomingOnly: Boolean
+}
+
+type EventFromSource {
```
Itrulia, line 73: Used to be Event

libs/utils/api/src/lib/dataloader/prime-dataloaders.decorator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      if (!resultItem) {
+        return resultItem
+      }
+
+      const results = Array.isArray(resultItem)
+        ? resultItem
+        : Array.isArray(resultItem.nodes)
+        ? resultItem.nodes
+        : [resultItem]
```
Itrulia, line 32: we should update prettier so this correctly gets formated :D

elias-summermatter (approved):

PR #1138, opened 2024-01-15
WPC-1199: move stuff from editor to libs/ui

antkiewiczk:


libs/ui/editor/src/lib/hooks/index.ts
```diff
@@ -0,0 +1 @@
+export * from './unsavedChangesDialog'
```
Itrulia, line 1: Just be aware that barrel files are very slow for bundlers, jest etc (learned this a few weeks ago) to parse which may cause issues in the future. But no need to care for now.

I might check what the impact for us is but I don't think it's too big due to using nx.

Itrulia (approved):

elias-summermatter (approved):

PR #1142, opened 2024-01-17
Yaml configuration for hosting

elias-summermatter:


apps/api-example/src/app/index.ts
```diff
@@ -5,14 +5,16 @@ import {
...
+  URLAdapter,
   WepublishServer
 } from '@wepublish/api'
 import pinoMultiStream from 'pino-multi-stream'
 import {createWriteStream} from 'pino-sentry'
 import pinoStackdriver from 'pino-stackdriver'
 import * as process from 'process'
-import {ExampleURLAdapter} from './url-adapter'
 import {Application} from 'express'
+import {loadAsync} from 'node-yaml-config'
```
Itrulia, line 16: Would https://www.npmjs.com/package/yaml not make more sense as node-yaml-config hasn't had an update in 2 years?

elias-summermatter: Just copied over what we had for now not finished yet :-)  but the advantage of  node-yaml-config is that it supports easy default configuration, and you can just override stuff. But yeah, if we don't use that, we can change to the default yaml package. :-)

elias-summermatter:
@tomaszdurka & @Itrulia please have a look.

Itrulia (approved):

PR #1144, opened 2024-01-18
Fixes Payrexx Adapter

elias-summermatter:


tomaszdurka (approved):
Im ok to keep the logs, but better create some simple logging mechanism.

PR #1149, opened 2024-01-23
fix(block-content-website): empty html block throwing errors

Itrulia:


libs/block-content/website/src/lib/html/html-block.stories.tsx
```diff
@@ -22,6 +22,12 @@ export const WithScript = {
   }
 }
 
+export const Empty = {
```
Itrulia, line 25: Added test to make sure it works in the future

antkiewiczk (approved):

PR #1150, opened 2024-01-23
F/wpc 1145 improved docker build

elias-summermatter:


antkiewiczk (dismissed):

package.json
```diff
@@ -11,6 +11,8 @@
     "clean:media": "docker-compose exec media rm -rf ./.media",
     "lint": "nx run-many --target=lint",
     "build": "nx run-many --targets=build",
+    "docker:build:api": "npx nx run-many --targets=build --projects=api-example",
```
Itrulia, line 14: Imo don't think these are needed and we could just run `npx nx build api-example` in the dockerfile. I don't like adding simple commands to the package.json. What do others think?

There's a few other scripts I would get rid of too :D They are mostly there bc of historic reasons (yarn time and workspaces)

Itrulia (dismissed):

Itrulia (approved):

Itrulia (approved):

tomaszdurka (changes requested):
@elias-summermatter Looks good!

As discussed let's change:
1. Allow to pass multiple tags for building/pushing api, editor and migration images
2. Authenticate to both ghcr and dockerhub in build-push jobs so we can push to both or any of them by specifying tags

Dockerfile
```diff
@@ -1,3 +1,117 @@
...
+RUN groupadd -r wepublish && \
+    useradd -r -g wepublish -d /wepublish wepublish && \
+    chown -R wepublish:wepublish /wepublish && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends openssl && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+COPY --chown=wepublish:wepublish apps/api-example/src/default.yaml /wepublish/config/default.yaml
+COPY --chown=wepublish:wepublish --from=build-api /wepublish/api /wepublish
+COPY --chown=wepublish:wepublish  --from=build-api /wepublish/node_modules/bcrypt node_modules/bcrypt
```
tomaszdurka: whitespace

tomaszdurka (approved):

Itrulia (approved):

PR #1162, opened 2024-02-06
Fixe mailgun overflow bug

elias-summermatter:


libs/mail/api/src/lib/mail-provider/mailgun-mail-provider.ts
```diff
@@ -129,16 +129,21 @@ export class MailgunMailProvider extends BaseMailProvider {
...
-            ? `${value}`
-            : JSON.stringify(value)
+        // Enforce max length of 16kb per key => https://documentation.mailgun.com/en/latest/api-sending.html
+        let serializedValue: string | number = ''
+        if (typeof value === 'number') {
+          serializedValue = value
+        } else if (typeof value === 'string') {
+          serializedValue = value.substring(0, 15000)
+        } else {
+          serializedValue = JSON.stringify(value).substring(0, 15000)
```
michael-scheurer, line 139: the sub-string function is called within a for loop. this means, that the entire form data could still be heavier than 16 kb in total.

michael-scheurer: 16 kb is per field.

michael-scheurer (approved):

michael-scheurer (approved):

PR #1165, opened 2024-02-08
feat(editor): implemented block styles

Itrulia:
* Changed output of blockStyle to be the name so that the frontend can target it


apps/editor/src/app/locales/de.json
```diff
@@ -715,6 +715,7 @@
       "authors": "Autor*innen",
       "blocks": {
         "polls": "Umfrage",
+        "blockStyles": "Block Styling",
```
antkiewiczk, line 718: i think some translations are missing here

antkiewiczk: also, no french translations

apps/editor/src/app/routes/blockStyles/blockStyleList.tsx
```diff
@@ -0,0 +1,379 @@
...
+  | {type: BlockStyleListActionType.Delete; payload: {id: string}}
+
+const mapBlockStyleToFormValue = (blockStyles: BlockStyle[] | null | undefined) =>
+  blockStyles?.reduce((obj, node) => {
+    obj[node.id] = node
+
+    return obj
+  }, {} as Record<string, BlockStyle>) ?? {}
+
+const blockstyleFormValueReducer = (
```
antkiewiczk, line 87: blockStyle <-- capital S

apps/editor/src/app/routes/events/eventForm.tsx
```diff
@@ -101,7 +101,7 @@ export const EventForm = ({event, onChange, create}: EventFormProps) => {
               <Form.Control
                 name="description"
                 value={event.description || []}
-                onChange={(description: RichTextBlockValue) => onChange({description})}
+                onChange={(description: RichTextBlockValue['richText']) => onChange({description})}
```
antkiewiczk, line 104: 👌

antkiewiczk (approved):
left some non-blocking comments

apps/editor/src/app/locales/de.json
Itrulia: Thats intended, it's like the translation I copied. We should define a workflow for this as it used to be that we just do english and Laurent does the rest.

We have a few that we only have in english.

apps/editor/src/app/routes/blockStyles/blockStyleList.tsx
Itrulia: Will be done in the next PR! Thanks for finding

PR #1167, opened 2024-02-13
BAJ-642:  Best of WePublish

antkiewiczk:


apps/bajour/components/bajour/best-of-wepublish.tsx
```diff
@@ -0,0 +1,264 @@
...
+    padding-top: 2rem;
+    padding-left: 2rem;
+  }
+`
+
+const BestOfWePublishHeader = styled('div')`
+  width: 100%;
+  background-color: #ffbaba;
+
+  > span {
```
Itrulia, line 78: Don't use direct elements please and ideally avoid nesting unless not really possible.

apps/bajour/components/bajour/best-of-wepublish.tsx
```diff
@@ -0,0 +1,264 @@
...
+            peerName: teaser?.peer?.name,
+            peerLogo: teaser?.peer?.profile?.logo.squareSmallURL || ''
+          }
+
+          return (
+            <LinkWrapper color="inherit" underline="none" href={values.href} key={values.title}>
+              <PeerLogoWrapper>
+                {/* not possible to use next/image with dynamic domains */}
+                {/* eslint-disable-next-line @next/next/no-img-element */}
+                <img
```
Itrulia, line 224: Why are we not using the Image component from website builder?

apps/bajour/components/bajour/best-of-wepublish.tsx
```diff
@@ -0,0 +1,264 @@
...
+`
+
+const BestOfWePublish = ({numColumns, teasers, blockStyle, className}: BestOfWePublishProps) => {
+  return (
+    <BestOfWePublishWrapper>
+      <BestOfWePublishHeader>
+        <span>Best of We.Publish</span>
+      </BestOfWePublishHeader>
+      <Wrapper>
+        {teasers.map(teaser => {
```
Itrulia, line 210: This should be a seperate component

apps/bajour/components/bajour/best-of-wepublish.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+const InnerContainer = styled('div')`
+  position: relative;
+  display: grid;
+  align-content: space-between;
+`
+
+const Content = styled('div')`
+  display: inline-block;
+  padding: 0.2rem;
```
Itrulia, line 137: Use spacing please

Itrulia (changes requested):
1. Use spacing for paddings etc please and not rem
2. Don't use raw HTML components as it's making styling and portability harder
3. No `<br />`s for this as again it makes styling harder as you would have to override the component
4. There's no need for the div hack as you can simply have the Title & Peer have their own white background based on their size

apps/bajour/components/bajour/best-of-wepublish.tsx
antkiewiczk: for some reason nextjs or rather next/image requires to have a list of supported domains defined in next config. But in our case, it's not possible as the images defined by media can be on any domain so it's impossible to list them all

https://nextjs.org/docs/app/api-reference/components/image#remotepatterns

apps/bajour/components/bajour/best-of-wepublish.tsx
antkiewiczk: is there a reason that nesting should be avoided? is there some best practice in styled-components/emotion or is it just simply because of specificity?

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish-teasers.tsx
```diff
@@ -0,0 +1,146 @@
...
+const TeaserBackground = styled(BuilderImage)`
+  width: 100%;
+  object-fit: cover;
+`
+
+const PeerLogoWrapper = styled('div')`
+  position: absolute;
+  top: 0;
+  left: 0;
+  padding-right: 0.4rem;
```
Itrulia, line 19: Using spacing from the theme and not rems please!

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
```diff
@@ -0,0 +1,129 @@
...
+`
+
+const WePublishEcosystem = styled('b')`
+  font-weight: 400;
+`
+
+const MoreButton = styled('a')`
+  padding: 0.5rem 1.5rem;
+  background-color: white;
+  border: 3px solid #ffbaba;
```
Itrulia, line 89:
```suggestion
  border: 3px solid currentColor;
```

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish-teasers.tsx
```diff
@@ -0,0 +1,146 @@
+import {styled} from '@mui/material'
+import {ApiV1, Image as BuilderImage} from '@wepublish/website'
+
+import {NextWepublishLink} from '../../should-be-website-builder/next-wepublish-link'
+
+interface Props {
```
Itrulia, line 6: Better naming please!

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish-teasers.tsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+`
+
+const LinkWrapper = styled(NextWepublishLink)`
+  display: grid;
+  grid-template-columns: repeat(24, 1fr);
+  grid-template-rows: auto;
+  align-items: center;
+  grid-template-columns: 1fr;
+  spacing: ${({theme}) => theme.spacing(2)};
```
Itrulia, line 36: don't think that CSS property exists

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish-teasers.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  ${({theme}) => theme.breakpoints.up('md')} {
+    font-size: 16px;
+  }
+`
+
+const BestOfWePublishTeasers = ({teasers}: BestOfWePublishTeasersProps) => {
+  return (
+    <>
+      {teasers?.map(teaser => {
```
Itrulia, line 103: this should only render a single teaser

Itrulia:
eaGS4jr

completely breaks the UI FYI

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
```diff
@@ -0,0 +1,131 @@
...
+  margin-left: 20%;
+  padding: 0 ${({theme}) => theme.spacing(2)};
+  display: inline-block;
+
+  ${({theme}) => theme.breakpoints.up('md')} {
+    font-size: 2rem;
+  }
+`
+
+const BestOfWePublishFooter = styled('div')`
```
Itrulia, line 52: There's a margin bottom missing as the button is now hanging into neighbouring elements. While the margin exists on the parent, that's not really portable and prone to errors.

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
```diff
@@ -0,0 +1,131 @@
...
+  height: 100%;
+
+  ${({theme}) => theme.breakpoints.up('md')} {
+    padding: 0 ${({theme}) => theme.spacing(2)};
+  }
+`
+
+const FooterContent = styled('span')`
+  font-weight: 300;
+  padding: ${({theme}) => theme.spacing(1)} 0 ${({theme}) => theme.spacing(4)}
```
Itrulia, line 74: padding: ${({theme}) => `${theme.spacing(1)} 0 ${theme.spacing(4)} ${theme.spacing(2)}`};

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
```diff
@@ -0,0 +1,131 @@
...
+    ${({theme}) => theme.spacing(2)};
+  display: inline-block;
+
+  ${({theme}) => theme.breakpoints.up('md')} {
+    padding: ${({theme}) => theme.spacing(1)} 0 ${({theme}) => theme.spacing(6)}
```
Itrulia, line 79: padding: ${({theme}) => `${theme.spacing(1)} 0 ${theme.spacing(6)} ${theme.spacing(3)}`};

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
```diff
@@ -0,0 +1,131 @@
+import {styled} from '@mui/material'
+import {ApiV1, BuilderTeaserGridBlockProps} from '@wepublish/website'
+
+import {BestOfWePublishTeasers} from './best-of-wepublish-teasers'
+
+export type BestOfWePublishProps = Omit<BuilderTeaserGridBlockProps, 'teasers'> & {
```
Itrulia, line 6:
```suggestion
export type BestOfWePublishProps = BuilderTeaserGridBlockProps & {
```

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish-teasers.tsx
```diff
@@ -0,0 +1,144 @@
...
+          peerName: teaser?.peer?.name,
+          peerLogo: (teaser?.peer?.profile?.logo as ApiV1.FullImageFragment)?.squareSmallURL || ''
+        }
+
+        return (
+          <LinkWrapper color="inherit" underline="none" href={values.href} key={values.title}>
+            <PeerLogoWrapper>
+              {/* not possible to use next/image with dynamic domains */}
+              {/* eslint-disable-next-line @next/next/no-img-element */}
+              <img
```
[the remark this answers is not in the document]
Itrulia: Also: 
image

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
antkiewiczk: why this change? because it causes some problems with typings when I do that

antkiewiczk:
> eaGS4jr
> 
> completely breaks the UI FYI

I noticed that the usage of <TeaserWrapper {...alignment}> broke the ui, also other elements don't use it and they just render stuff like div etc, instead of article, so I removed it and now it looks like it's not breaking anything anymore

apps/bajour/components/bajour/best-of-wepublish/best-of-wepublish.tsx
Itrulia: I don't have those issues. That change is as there's no need to omit it and then override it.

Itrulia:
@antkiewiczk switching from grid 1/13 to -1/1 also fixes the issue

antkiewiczk:
> issue

didn't help, only getting rid of <TeaserWrapper {...alignment}>  worked for me 🤷

Itrulia (approved):

PR #1169, opened 2024-02-19
WPC-1470: Stats lib

antkiewiczk:


libs/stats/api/src/lib/stats.service.ts
```diff
@@ -0,0 +1,32 @@
+import {Injectable} from '@nestjs/common'
+import {PrismaClient} from '@prisma/client'
+
+@Injectable()
+export class StatsService {
+  constructor(private prisma: PrismaClient) {}
+
+  async getStats() {
```
Itrulia, line 8: There's a few issues here:

1. All data is loaded aftereachother
2. All articles are counted, even non published ones

libs/stats/api/src/lib/stats.service.ts
Itrulia, line 1: Tests missing

libs/stats/api/src/lib/stats.resolver.ts
```diff
@@ -0,0 +1,15 @@
+import {Query, Resolver} from '@nestjs/graphql'
+import {Stats} from './stats.model'
+import {StatsService} from './stats.service'
+
+@Resolver(() => Stats)
+export class StatsResolver {
```
Itrulia, line 6: Do we maybe want to make all stats loaded independently from each other?
Even if I would for example just need the article count, the rest is still loaded. We could use resolvers here instead (search for `@Resolver`)

Itrulia (changes requested):

libs/stats/api/src/lib/stats.model.ts
```diff
@@ -0,0 +1,13 @@
+import {Field, ObjectType} from '@nestjs/graphql'
+
+@ObjectType()
+export class Stats {
+  @Field()
```
Itrulia, line 5: Probably want to make it an int here

libs/stats/api/src/lib/stats.resolver.ts
antkiewiczk: are you thinking about making separate methods in service that would return each value (e.g. getArticleCount, getAuthorCount) or continue using a single method (getStats), but add parameters representing values?

antkiewiczk: also, is there a point to do it now as we will only be using this api in a single bajour component?

Itrulia: Would be 3 separate methods and would also fix the issue of them running after each other instead of the same time.

I think it would serve as a good base, but there's not much of a *need* right now. But when we want to extend this, every added stat would make the endpoint slower etc.

apps/api-example/schema-v2.graphql
```diff
@@ -307,6 +307,8 @@ type PeriodicJob {
 type Query {
   "\n      Returns all active subscribers.\n      Includes subscribers with a cancelled but not run out subscription.\n    "
   activeSubscribers: [DashboardSubscription!]!
+  articlesCount: Float!
```
Itrulia, line 310: Those 2 probably should be ints

libs/stats/api/src/lib/stats.resolver.ts
```diff
@@ -0,0 +1,23 @@
+import {Query, Resolver} from '@nestjs/graphql'
+import {StatsService} from './stats.service'
+import {GraphQLISODateTime} from '@nestjs/graphql'
+
+@Resolver()
+export class StatsResolver {
+  constructor(private statsService: StatsService) {}
+
+  @Query(() => Number)
```
Itrulia, line 9: Should probably be using `ResolveField` here and not `Query`, else you need to fire multiple queries. `@Resolver` + `@ResolveField` allows you to lazyload fields.

libs/stats/api/src/lib/stats.service.ts
```diff
@@ -0,0 +1,33 @@
...
+@Injectable()
+export class StatsService {
+  constructor(private prisma: PrismaClient) {}
+
+  async getAuthorsCount(): Promise<number> {
+    return await this.prisma.author.count()
+  }
+
+  async getArticlesCount(): Promise<number> {
+    return await this.prisma.article.count()
```
Itrulia, line 13: this counts all articles and not just published ones.

michael-scheurer (commented):
General question: @Itrulia you implemented some time ago a stats api for memberships and subscriptions. Should we probably use the same lib and same graphql endpoint? Because it's both stats. How do you see this?

Itrulia (approved):

PR #1176, opened 2024-02-21
feat(bajour): redirect old urls to new urls

Itrulia:
Also removed mail chimp as not used anymore


apps/bajour/next.config.js
```diff
@@ -43,7 +43,57 @@ const nextConfig = {
       }
     }
   },
-  transpilePackages: ['@wepublish/ui', '@wepublish/website', 'react-tweet']
+  transpilePackages: ['@wepublish/ui', '@wepublish/website', 'react-tweet'],
+  async redirects() {
+    return [
+      {
+        source: '/home',
```
Itrulia, line 50: Because the home page has a slug of /home, we probably should change that in the future

apps/bajour/next.config.js
```diff
@@ -43,7 +43,57 @@ const nextConfig = {
...
+          }
+        ]
+      },
+      {
+        source: '/profile/dashboard',
+        destination: '/profile',
+        permanent: false
+      },
+      {
+        source: '/member-uebersicht',
```
Itrulia, line 81: Those 3 probably should be split into 3 pages to stay that way, right @michael-scheurer ?

antkiewiczk (approved):

PR #1208, opened 2024-03-05
GRUP-8: update navbar in website builder + in bajour and gruppetto

antkiewiczk:


apps/bajour/pages/_app.tsx
```diff
@@ -84,6 +84,9 @@ const ButtonLink = styled('a')`
 function CustomApp({Component, pageProps}: CustomAppProps) {
   const siteTitle = 'Bajour'
 
+  const {data: peerInfoData} = ApiV1.usePeerProfileQuery({fetchPolicy: 'network-only'})
```
Itrulia, line 87: Any reason this is network only? Can't SSR it now.

antkiewiczk:
> Thanks for your work. I found some details, we should correct.
> 
> * The logo has no space to the top nor to the bottom of the header. Please compare it with the Design of Ron (Default Design) and current Gruppetto implementation.
>   image
> * The header navbar should have background color white. Please compare it with Gruppetto.
> * Are the links in the header implemented? How can I see an example? How can Bajour or any other media add those links?

so the space (or rather no space) is because it really depends on the logo - when the logo is more or less narrow than on the designs I still wanted for it to fit - and if the logo is much wider - there would still be bottom/top space as on designs. But I'll look into it and check if it can be improved

Regarding the links implementation and  `How can Bajour or any other media add those links?` - I was actually planning to talk with you and suggest making a proper documentation for all of the components/features so that media can be sure that for example here they need to pass 'header' as key, that for example bajour archive has to be 6col with "Archive" blockStyle, that basel briefing should have "BaselBriefing" blockStyle, FCB briefing FCBBriefing etc

Itrulia:
Missing SSR of the data loaded, search for 

```
client.query({
      query: ApiV1.NavigationListDocument
    })
```

and add the new query too!

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -23,45 +25,172 @@ const appBarStyles = (theme: Theme, isMenuOpen: boolean) =>
...
+    height: ${({theme}) => theme.spacing(12.5)};
+
+    svg {
+      width: ${({theme}) => theme.spacing(6.5)};
+      height: ${({theme}) => theme.spacing(6.5)};
+    }
+  }
+`
+
+const LogoLink = styled(Link)<{isMenuOpen?: boolean}>`
```
Itrulia, line 141: Missing exports and prob want to rename because of that. Same for the others.

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -23,45 +25,172 @@ const appBarStyles = (theme: Theme, isMenuOpen: boolean) =>
...
+    css`
+      z-index: -1;
+    `}
+`
+
+const LogoWrapper = styled('div')`
+  fill: currentColor;
+  width: auto;
+
+  img {
```
Itrulia, line 160: can't we pass down the styling as `css={..}` on the image?

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -85,12 +214,39 @@ export function Navbar({className, children, categorySlugs, slug, data}: Builder
         <NavbarInnerWrapper>
           <NavbarMain>
             <NavbarIconButtonWrapper>
-              <IconButton size="large" color="secondary" aria-label="Menu" onClick={toggleMenu}>
+              <IconButton size="large" aria-label="Menu" onClick={toggleMenu} sx={{color: 'white'}}>
```
Itrulia, line 217: Probably don't want to use inline styles here

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -23,45 +25,172 @@ const appBarStyles = (theme: Theme, isMenuOpen: boolean) =>
...
   justify-content: center;
   align-items: center;
+  width: ${({theme}) => theme.spacing(6.5)};
+  height: ${({theme}) => theme.spacing(6.5)};
+
+  ${({theme}) => theme.breakpoints.up('md')} {
+    width: ${({theme}) => theme.spacing(7.5)};
+    height: ${({theme}) => theme.spacing(7.5)};
+
+    svg {
```
Itrulia, line 124: Any reason we don't pass that down via `css={...}`? Makes it easier to override as it's not using a specificity of 3

Itrulia (approved):

PR #1211, opened 2024-03-06
feat(seeding): Ensure admin user while seeding

tomaszdurka:
Create admin user with generated password if not exists
Alternatively if OVERRIDE_ADMIN_PW passed set the password to the value

### DIFF
--- a/libs/api/prisma/run-seed.ts
+++ b/libs/api/prisma/run-seed.ts
@@ -2,6 +2,7 @@ import {PrismaClient} from '@prisma/client'
 import {seed as rootSeed} from './seed'
 import bcrypt from 'bcrypt'
 import {randomBytes} from 'crypto'
+import {DefaultBcryptHashCostFactor} from '../src/lib/db/common'
 
 const generateSecureRandomPassword = (length: number) => {
   const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-'
@@ -18,6 +19,7 @@ const generateSecureRandomPassword = (length: number) => {
   }
   return password
 }
+
 export async function runSeed() {
   const prisma = new PrismaClient()
   await prisma.$connect()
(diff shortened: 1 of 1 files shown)


tomaszdurka:
@elias-summermatter the main change is that it will always change the password to `OVERRIDE_ADMIN_PW` - not only when admin user does not exist.

libs/api/prisma/run-seed.ts
```diff
@@ -27,47 +28,52 @@ export async function runSeed() {
...
     }
-  })
-  if (!admin) {
-    const email = 'admin@wepublish.ch'
-    const overriddenPw = process.env.OVERRIDE_ADMIN_PW
-    let password = generateSecureRandomPassword(20)
-    if (overriddenPw) {
-      console.log(
-        '\x1b[31m\x1b[1m%s\x1b[0m',
-        `!!!!!!!!!!!!!!!!!! WARNING: UNSECURE PASSWORD IS OVERRIDDEN WITH ENV VARIABLE !!!!!!!!!!!!!!!!!! `
```
elias-summermatter, line 42: Maybe keep the warning? What do you think?

libs/api/prisma/run-seed.ts
```diff
@@ -27,47 +28,52 @@ export async function runSeed() {
...
-      `Bootstrapping initial admin user ${email} with password: ${password}`
-    )
+    console.log('\x1b[31m\x1b[1m%s\x1b[0m', `Ensuring admin user`)
     await prisma.user.upsert({
-      where: {
-        id: 'admin'
+      where: {id},
+      update: {
+        ...userBase,
+        password: await bcrypt.hash(password, 11)
```
elias-summermatter, line 44: Store password hash in a variable and use hash function just once.  And maybe increase the weight from 11 to 12

elias-summermatter (changes requested):
Besides, this is looks nice, thanks  :+1:

libs/api/prisma/run-seed.ts
tomaszdurka: OK

libs/api/prisma/run-seed.ts
elias-summermatter: That does not really matter you can change the text, but it's important to notice that it's an unsecure setup.

PR #1215, opened 2024-03-07
feat(editor/api/website): implement teaser list block

Itrulia:
also saw 2 small issues with the docker-compose


libs/api/src/lib/db/block.ts
```diff
@@ -198,7 +199,6 @@ export enum TeaserType {
   PeerArticle = 'peerArticle',
   Page = 'page',
   Event = 'event',
-  External = 'external',
```
Itrulia, line 201: ? never seen this one so deleted it

libs/api/src/lib/graphql/blocks.ts
```diff
@@ -487,6 +496,269 @@ export const GraphQLPublicTeaser = new GraphQLUnionType({
...
+              },
+              ArticleSort.PublishedAt,
+              SortOrder.Descending,
+              undefined,
+              skip,
+              take,
+              prisma.article
+            )
+
+            articles.nodes.forEach(article => loaders.articles.prime(article.id, article))
```
Itrulia, line 555: Priming it means it wont have to be fetched again from dataloaders

libs/api/src/lib/graphql/blocks.ts
```diff
@@ -487,6 +496,269 @@ export const GraphQLPublicTeaser = new GraphQLUnionType({
...
+        }
+      )
+    }
+  },
+  isTypeOf: createProxyingIsTypeOf(value => {
+    return value.type === BlockType.TeaserList
+  })
+})
+
+export const GraphQLPublicTeaserListBlock = new GraphQLObjectType<TeaserListBlock, Context>({
```
Itrulia, line 636: Requires a public seperate one because we use the same name for public & private. Which means it will have a naming conflict and doesn't allow that.

libs/api/src/lib/graphql/blocks.ts
```diff
@@ -487,6 +496,269 @@ export const GraphQLPublicTeaser = new GraphQLUnionType({
...
+      resolve: resolveBlockStyleIdToName
+    },
+    teaserType: {
+      type: GraphQLTeaserType
+    },
+    filter: {type: new GraphQLNonNull(GraphQLTeaserListBlockFilter)},
+    take: {type: GraphQLInt},
+    skip: {type: GraphQLInt},
+    teasers: {
+      type: new GraphQLNonNull(new GraphQLList(GraphQLTeaser)),
```
Itrulia, line 538: Reusing the same interface for teasers for convenience and code consistency. Had to keep the optional on teasers else I would need to rename the property in the client schema due to naming conflicts.

I tried to make all teasers optional, but the editor is a bit fucked up how blocks are rendered and reuses the output types for the input and that causes issues. It's quite a refactor and abandoned it halfway through.

libs/ui/editor/src/lib/blocks/types.ts
```diff
@@ -190,14 +191,6 @@ export type EmbedBlockValue =
   | BildwurfAdEmbed
   | OtherEmbed
 
-export enum TeaserType {
```
Itrulia, line 193: Now exposed from the generated schema.

libs/ui/editor/src/lib/blocks/types.ts
```diff
@@ -687,7 +706,8 @@ export function unionMapForBlock(block: BlockValue): BlockInput {
                   blockStyle: block.value.blockStyle
                 }
             }
-          })
+          }),
+          blockStyle: block.value.blockStyle
```
Itrulia, line 710: Was missing, oops.

libs/ui/editor/src/lib/blocks/types.ts
```diff
@@ -977,6 +997,35 @@ export function blockForQueryBlock(block: FullBlockFragment | null): BlockValue
...
+          blockStyle: block.blockStyle,
+          filter: block.filter,
+          skip: block.skip ?? 0,
+          take: block.take ?? 6,
+          teaserType: block.teaserType ?? TeaserType.Article,
+          teasers: block.teasers.map((teaser, index) => [
+            `${index}`,
+            {
+              ...teaser,
+              type:
```
Itrulia, line 1014: No clue why we rewrite the __typename to type, they essentially do the same thing. Had to do this as the typing for the input teasers uses this.

libs/ui/editor/src/lib/panel/teaserListConfigPanel.tsx
```diff
@@ -0,0 +1,164 @@
...
+  width: 200px !important;
+`
+
+export type TeaserListConfigPanelProps = {
+  value: TeaserListBlockValue
+  onClose(): void
+  onSelect(newValue: TeaserListBlockValue): void
+}
+
+export const useTagTypeText = () => {
```
Itrulia, line 25: Eh this should be useTeaserTypeText, too much tag handling in this file I guess :D

libs/website/api/src/lib/absolute-url-to-relative.ts
```diff
@@ -22,7 +22,13 @@ export const absoluteUrlToRelative: InMemoryCacheConfig['typePolicies'] = {
   Article: {
     fields: {
       url: {
-        merge: (_, url: string) => absoluteToRelative(url)
+        merge: (_, url: string, options) => {
```
Itrulia, line 25: readded it as the link to the original article was broken

libs/api/src/lib/graphql/blocks.ts
antkiewiczk: :( unnecessary doubling of code, hopefully there will be time this year to move everything to nest

libs/ui/editor/src/lib/blocks/teaserListBlock.tsx
```diff
@@ -0,0 +1,107 @@
...
+`
+
+export const TeaserListBlock = ({value, onChange, autofocus}: BlockProps<TeaserListBlockValue>) => {
+  const {filter, teasers, skip, take, teaserType} = value
+  const [isDialogOpen, setIsDialogOpen] = useState(false)
+  const {t} = useTranslation()
+  const teaserTypeText = useTeaserTypeText()
+
+  const isEmpty = !teasers.length
+  const teasersToDisplay = teasers.slice(0, 6)
```
antkiewiczk, line 54: should this value be hardcoded? 6 I mean

libs/ui/editor/src/lib/panel/teaserListConfigPanel.tsx
```diff
@@ -0,0 +1,164 @@
...
+import {TeaserListBlockValue} from '../blocks/types'
+import styled from '@emotion/styled'
+import {css} from '@emotion/react'
+
+const DrawerBody = styled(Drawer.Body)`
+  padding: 24px;
+`
+
+const inputStyles = css`
+  width: 200px !important;
```
antkiewiczk, line 16: not perfect 😆

antkiewiczk (approved):

libs/ui/editor/src/lib/panel/teaserListConfigPanel.tsx
Itrulia: Yeah it's because I use it for multiple different inputs and they have like 3-4 levels of specificity :(

libs/ui/editor/src/lib/blocks/teaserListBlock.tsx
Itrulia: Yeah it's just how much to preview

PR #1232, opened 2024-03-12
f/WP-49: trial subscription

michael-scheurer:


libs/api/prisma/migrations/20240313100544_add_extendable_subscription_flag_and_max_subscription_count/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- AlterTable
+ALTER TABLE "member.plans" ADD COLUMN     "extendable" BOOLEAN NOT NULL DEFAULT true,
+ADD COLUMN     "maxCount" INTEGER;
+
+-- AlterTable
+ALTER TABLE "subscriptions" ADD COLUMN     "extendable" BOOLEAN NOT NULL DEFAULT true;
+
+-- Add constraint
+ALTER TABLE "subscriptions"
+    ADD CONSTRAINT "check_extendable_autorenew"
```
michael-scheurer, line 10: I would like to see a test for this.

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -554,6 +557,11 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
           throw new SubscriptionNotFound()
         }
 
+        // Prevent user from extending not extendable subscription
+        if (!subscription.extendable) {
```
michael-scheurer, line 561: A test for this extend-function as well as for the update function would be perfect. Could catch missing checks from other programmers, not involved with this functionality.

libs/api/src/lib/memberContext.ts
```diff
@@ -622,17 +622,38 @@ export class MemberContext implements MemberContext {
...
+
+    const memberPlan = await prisma.memberPlan.findUnique({where: {id: memberPlanId}})
+    const memberPlanSubscriptionCount = await prisma.subscription.count({
+      where: {
+        userID,
+        memberPlanID: memberPlanId
+      }
+    })
+
+    if (memberPlan.maxCount && memberPlan.maxCount <= memberPlanSubscriptionCount) {
```
michael-scheurer, line 648: In case maxCount === 0, this error would never hit. From a users perspective, this would probably not be intended.
I would test this functionality with unit tests using undefined, null, 0 and 1 as input values.

libs/api/prisma/migrations/20240313100544_add_extendable_subscription_flag_and_max_subscription_count/migration.sql
```diff
@@ -0,0 +1,11 @@
...
+        CHECK ("extendable" OR NOT "autoRenew");
```
michael-scheurer, line 11: For now, it is possible to create non-extendable member-plans, with a forced auto-renew setting. We should catch this also on api level.

michael-scheurer: image

libs/api/src/lib/memberContext.ts
```diff
@@ -286,7 +286,9 @@ export class MemberContext implements MemberContext {
               amount,
               quantity: 1
             }
-          }
+          },
+          // mark invoices with amount 0 as paid
+          paidAt: amount === 0 ? new Date() : undefined
```
michael-scheurer, line 291: @elias-summermatter can you please double-check

apps/api-example/schema-v2.graphql
```diff
@@ -30,7 +30,6 @@ enum BlockType {
   TeaserGrid1
   TeaserGrid6
   TeaserGridFlex
-  TeaserList
```
michael-scheurer, line 33: mistakenly removed?

elias-summermatter (approved):

PR #1241, opened 2024-03-15
fix(api): multiple open connections

Itrulia:
Tuned it down from like 10 to 1

### DIFF
--- a/libs/mail/api/src/lib/mail.controller.spec.ts
+++ b/libs/mail/api/src/lib/mail.controller.spec.ts
@@ -9,7 +9,7 @@ import {
 } from '@wepublish/testing'
 import {Test, TestingModule} from '@nestjs/testing'
 import {forwardRef} from '@nestjs/common'
-import {PrismaModule, PrismaService} from '@wepublish/nest-modules'
+import {PrismaModule} from '@wepublish/nest-modules'
 import {matches} from 'lodash'
 import bodyParser from 'body-parser'
 import Mailgun from 'mailgun.js'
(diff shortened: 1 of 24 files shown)


libs/mail/api/src/lib/mail.controller.spec.ts
```diff
@@ -164,7 +164,7 @@ describe('MailController', () => {
       root3: 'ok'
     }
 
-    await new MailController(prismaClient as PrismaService, mailContext, {
```
antkiewiczk, line 167: 🤣

libs/membership/api/src/lib/mail-template/mail-template.resolver.spec.ts
```diff
@@ -112,7 +110,7 @@ describe('MailTemplatesResolver', () => {
     const module: TestingModule = await Test.createTestingModule({
       providers: [
         MailTemplatesResolver,
-        {provide: PrismaService, useValue: prismaServiceMock},
+        {provide: PrismaClient, useValue: prismaServiceMock},
```
antkiewiczk, line 113: probably should be called prismaClientMock

libs/membership/api/src/lib/mail-template/mail-template.resolver.ts
```diff
@@ -1,14 +1,14 @@
...
 import {CanGetMailTemplates, CanSyncMailTemplates, Permissions} from '@wepublish/permissions/api'
 import {MailTemplateSyncService} from './mail-template-sync.service'
 import {MailProviderModel, MailTemplateWithUrlAndStatusModel} from './mail-template.model'
+import {PrismaClient} from '@prisma/client'
 
 @Resolver(() => MailTemplateWithUrlAndStatusModel)
 export class MailTemplatesResolver {
   constructor(
-    private prismaService: PrismaService,
+    private prismaService: PrismaClient,
```
antkiewiczk, line 11: also here probably should be  `private prismaClient: PrismaClient,`

antkiewiczk (approved):
added some non-blocking comments

PR #1244, opened 2024-03-19
fix(navigation-website): links not using client router & not needed styles

Itrulia:


libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -67,16 +74,23 @@ export const NavbarLinks = styled('div')<{isMenuOpen?: boolean}>`
   }
 `
 
-export const NavbarLink = styled('a')`
```
Itrulia, line 70: this used to cause reloads

PR #1246, opened 2024-03-20
fix navbar

antkiewiczk:


libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -129,20 +129,24 @@ export const NavbarIconButtonWrapper = styled('div')`
...
     }
   }
 
   ${({theme}) => theme.breakpoints.up('lg')} {
+    width: ${({theme}) => theme.spacing(12.5)};
     height: ${({theme}) => theme.spacing(12.5)};
 
     svg {
+      width: ${({theme}) => theme.spacing(6.5)};
       height: ${({theme}) => theme.spacing(6.5)};
```
Itrulia, line 150: dont think we need those reverts

antkiewiczk: yeah I didn't notice the color, it should be ${({theme}) => theme.palette.common.white};

but for width, for me the hamburger is squeezed

Itrulia: Ahhh! This should probably be font-size then and no need to set width/height. As the width & height is set to 1em. Good catch!

PR #1267, opened 2024-04-09
BABA-25: Add pagination to article list

antkiewiczk:


libs/article/website/src/lib/article-list/article-list.tsx
Itrulia, line 1: Article list should not have pagination. This is a job for the parent of article list.

Itrulia (changes requested):

libs/article/website/src/lib/article-list/article-list.tsx
```diff
@@ -21,14 +28,35 @@ export const ArticleListWrapper = styled('article')`
...
+  }
+
+  const paginationCount = useMemo(() => {
+    if (data?.articles.totalCount && data?.articles.totalCount > limit) {
+      return Math.ceil(data.articles.totalCount / limit)
+    }
+    return 1
+  }, [data?.articles.totalCount, limit])
+
+  useEffect(() => {
```
Itrulia, line 48: There's like 3 bugs in this useEffect, it's also not needed

libs/article/website/src/lib/article-list/article-list.tsx
```diff
@@ -1,5 +1,12 @@
-import {css, styled} from '@mui/material'
+import {Pagination, css, styled} from '@mui/material'
 import {BuilderArticleListProps, useWebsiteBuilder} from '@wepublish/website/builder'
+import {useEffect, useMemo, useState} from 'react'
+
+export const PaginationStyled = styled(Pagination)`
```
Itrulia, line 5: No styled post fix please. **Ever** when it gets exported.

libs/article/website/src/lib/article-list/article-list.tsx
antkiewiczk: why is onVariablesChange passed to article-list? (why is it a prop?)

Itrulia: So that you can implement pagination.

libs/article/website/src/lib/article-list/article-list.tsx
Itrulia: Yes, but inside the website builder.

Itrulia: It's part of the library but not the builder itself. So it can not be overridden. Checkout the WebsiteBuilderContext

libs/article/website/src/lib/article-list/article-list.tsx
antkiewiczk: so you mean article-list-container? or should I create an override for article-list in babanews?

apps/babanews/pages/a/index.tsx
```diff
@@ -13,11 +25,38 @@ export default function ArticleList() {
...
+  })
+
+  const paginationCount = useMemo(() => {
+    if (data?.articles.totalCount && data?.articles.totalCount > limit) {
+      return Math.ceil(data.articles.totalCount / limit)
+    }
+    return 1
+  }, [data?.articles.totalCount, limit])
+
+  useEffect(() => {
```
Itrulia, line 46: Not needed and causes a double render, call onVariablesChange inside the onChange

Itrulia: Actually no, the way this was implemented we need the setPage. But we can move this also into the onChange to not have a double render.

Itrulia (approved):

PR #1270, opened 2024-04-11
feat(bajour/article-website): use teaser slider for article list & automatic iframe sizing

Itrulia:
Changed ArticleList to use the TeaserGrid implementation
Fixed teasers using css instead of styled

PR #1273, opened 2024-04-11
fix bajour build

antkiewiczk:


Itrulia:
Fixed with https://github.com/wepublish/wepublish/pull/1270

PR #1275, opened 2024-04-15
chore(nx): Upgrade nx to v18

tomaszdurka:


Itrulia (approved):

Itrulia (changes requested):
Error: Component selectors can only be used in conjunction with @emotion/babel-plugin, the swc Emotion plugin, or another Emotion-aware compiler transform.


when starting bajour

tomaszdurka:
@Itrulia seems fine. 
However the single build on Vercel is delaying CI heavily.

Itrulia (approved):

PR #1276, opened 2024-04-15
fix(tsri): build error

Itrulia:


apps/tsri/pages/index.tsx
```diff
@@ -1,37 +1,13 @@
 import {css, styled} from '@mui/material'
 import {
   ApiV1,
-  PageContainer,
```
antkiewiczk, line 4: looks good, please fix lint

PR #1277, opened 2024-04-15
BABA-28: babanews various fixes

antkiewiczk:


apps/babanews/src/components/babanews/fullwidth-banner/fullwidth-banner.tsx
```diff
@@ -3,10 +3,11 @@ import {styled} from '@mui/material'
 import {BuilderBreakBlockProps, useWebsiteBuilder} from '@wepublish/website'
 import {useMemo} from 'react'
 
-const Banner = styled('div')`
+const Banner = styled('a')`
```
Itrulia, line 6: Does this link to an internal url or always outwards?

libs/website/builder/src/lib/navbar.interface.ts
```diff
@@ -12,5 +12,6 @@ export type BuilderNavbarProps = PropsWithChildren<
     loginUrl?: string
     profileUrl?: string
     subscriptionsUrl?: string
+    showSubscriptionsUrl?: boolean
```
Itrulia, line 15: I guess okay for now, we need to find a better way to also add stuff etc to it in the future

Itrulia (approved):

PR #1278, opened 2024-04-15
bugfix(tsri): Bring back teaser date

tomaszdurka:
@Itrulia Hey, I see your changes removed the `TeaserDate` and also used it in `tsri` website. 
Can you confirm this PR or fix it differently?


tomaszdurka:
nvm I see you fixed it already

PR #1281, opened 2024-04-16
WP-158: editor various fixes and improvements

antkiewiczk:


libs/ui/editor/src/lib/atoms/addBlockInput.tsx
```diff
@@ -26,6 +26,7 @@ export interface AddBlockInputProps {
 
 const Wrapper = styled.div`
   position: relative;
+  left: 41.5%;
```
Itrulia, line 29: Whats this achieving?

antkiewiczk: basically Michael wanted to align the + button more or less with the box that renders the teaser. I know the solution is not perfect, but I don't think that it really makes sense and still, after my fix it looks better than before

https://wepublish.atlassian.net/jira/software/c/projects/WP/boards/25?search=158&selectedIssue=WP-158

Screenshot 2024-04-17 at 14 55 00

Itrulia (approved):

PR #1282, opened 2024-04-16
GRUP-2: Add Google Analytics script

LukasSkywalker:
This adds the Google Analytics script to all pages.

### DIFF
--- a/apps/gruppetto/pages/_app.tsx
+++ b/apps/gruppetto/pages/_app.tsx
@@ -23,6 +23,7 @@ import i18next from 'i18next'
 import LanguageDetector from 'i18next-browser-languagedetector'
 import {AppProps} from 'next/app'
 import getConfig from 'next/config'
+import {GoogleAnalytics} from '@next/third-parties/google'
 import Head from 'next/head'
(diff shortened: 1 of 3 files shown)


apps/gruppetto/pages/_app.tsx
```diff
@@ -142,6 +143,8 @@ function CustomApp({Component, pageProps}: CustomAppProps) {
               <title key="title">{siteTitle}</title>
               <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
+              <GoogleAnalytics gaId="G-FL5CRKBXV4" />
```
Itrulia, line 146: Probably shouldn't be in the <Head> according to documentation. But I guess it doesn't hurt either.

Itrulia (dismissed):

antkiewiczk (approved):

apps/gruppetto/pages/_app.tsx
LukasSkywalker: You are probably right. Moved it below the content.

Itrulia (approved):

PR #1290, opened 2024-04-22
fix(bajour,image-website): various fixes found in the bajour review

Itrulia:


apps/bajour/pages/_app.tsx
```diff
@@ -116,25 +117,25 @@ function CustomApp({Component, pageProps}: CustomAppProps) {
             <ThemeProvider theme={theme}>
               <CssBaseline />
 
+              <GoogleAnalytics gaId="391346139" />
```
Itrulia, line 120: TODO: use env variables

PR #1292, opened 2024-04-22
fix/BA-62-minor-ui-enh.

michael-scheurer:


apps/bajour/src/components/website-builder-styled/blocks/break-block-styled.tsx
```diff
@@ -0,0 +1,17 @@
+import {styled} from '@mui/material'
+import {BreakBlock} from '@wepublish/website'
+
+export const BajourBreakBlock = styled(BreakBlock)`
+  background-color: ${({theme}) => theme.palette.secondary.cta};
+  color: white;
+
+  ${({theme}) => theme.breakpoints.up('sm')} {
```
Itrulia, line 8: We should be able to solve this without importants.

Itrulia (approved):

PR #1302, opened 2024-04-24
feat(bajour, membership-website, payment-website): integrated stripe

Itrulia:
Also parametized GA for Bajour & Gruppetto


libs/membership/website/src/lib/memberplan-picker/memberplan-picker.tsx
```diff
@@ -19,6 +19,8 @@ export const MemberPlanPickerRadios = styled(RadioGroup)`
   // hide unwanted label
   label {
     margin: 0;
+    display: grid;
+    align-items: stretch;
```
Itrulia, line 23: makes them the same height

PR #1305, opened 2024-04-24
fix image aspect ratio

michael-scheurer:


antkiewiczk (commented):
I think this can be closed after merging Karl's PR

PR #1307, opened 2024-04-27
GRUP-12: Allow linking image blocks to a URL

LukasSkywalker:
This change allows adding the "linkUrl" to image blocks, which are then turned into hyperlinks to the specified URL. This allows turning images into ads that point to an external site.
The new linkUrl attribute is a bit of special case. All other content blocks allow changing the block attributes with the corresponding EditPanel. The image block is a bit different in that regard, and doesn't open the block attributes, but those of the underlying image.
This PR modifies the EditPanel to also expose the new linkUrl attribute in addition to the existing image attributes.


michael-scheurer (approved):

antkiewiczk (approved):

antkiewiczk (commented):
Please add tests

Itrulia (changes requested):
Please add a test case to the image-block story & run tests via `nx test image-website` to generate the snapshot.

antkiewiczk (approved):

michael-scheurer (dismissed):

tomaszdurka (dismissed):

PR #1308, opened 2024-04-29
HOTFIX: Prevent creation of multiple invoices

elias-summermatter:


libs/membership/api/src/lib/subscription/subscription.service.ts
```diff
@@ -75,9 +75,25 @@ export class SubscriptionService {
...
+
+    // Ensure that no invoice gets created if there is a open invoice
+    const subscriptions = []
+    for (const subscription of allSubscriptions) {
+      if (subscription.invoices.length === 0) {
+        subscriptions.push(subscription)
+      }
+    }
+
+    return subscriptions
```
michael-scheurer, line 96: Make it more readable
`return allSubscriptions.filter(subscription => !subscription.invoices.length)`

tomaszdurka (approved):

PR #1312, opened 2024-04-30
WPC-1419: Add page info to full text search

LukasSkywalker:
This adds the values required for pagination to the result of the phrase query. In order to display a pagination component, at least the total number of results is required. A first query is run without any pagination settings applied. This produces a list of ids, and indirectly the total count of matches. For loading the content, another query is run with the pagination and sort orders applied.

### DIFF
--- a/libs/api/src/lib/graphql/phrase/phrase.ts
+++ b/libs/api/src/lib/graphql/phrase/phrase.ts
@@ -1,20 +1,40 @@
-import {GraphQLObjectType, GraphQLNonNull, GraphQLList} from 'graphql'
+import {GraphQLObjectType, GraphQLNonNull, GraphQLList, GraphQLInt} from 'graphql'
 import {Context} from '../../context'
 import {GraphQLPublicArticle} from '../article'
 import {PublicArticle} from '../../db/article'
 import {PublicPage} from '../../db/page'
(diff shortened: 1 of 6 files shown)


elias-summermatter (dismissed):

libs/api/src/lib/graphql/phrase/phrase.ts
```diff
@@ -1,20 +1,42 @@
...
+import {GraphQLPageInfo} from '../common'
 
-export const GraphQLPublicPhrase = new GraphQLObjectType<
-  {
-    articles: PublicArticle[]
-    pages: PublicPage[]
-  },
-  Context
->({
+interface PhraseResult {
```
Itrulia, line 9: Unused I think and not added to GraphQLObjectType.

libs/api/src/lib/graphql/phrase/phrase.ts
```diff
@@ -1,20 +1,42 @@
...
+  articles: {
```
Itrulia, line 10: Can use ConnectionResult I think here

Itrulia (approved):

Itrulia (approved):

michael-scheurer (approved):

PR #1317, opened 2024-05-06
chore(v2-migration): Migrate navigation to v2

tomaszdurka:


libs/navigation/api/src/lib/navigation.service.ts
Itrulia, line 1: Missing dataloader & decorators

libs/navigation/api/src/lib/navigation.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  async getNavigations() {
+    return this.prisma.navigation.findMany({
+      orderBy: {createdAt: 'desc'},
+      include: {links: true}
+    })
+  }
+
+  async createNavigation(input: CreateNavigationInput) {
+    const {links, ...data} = input
```
Itrulia, line 31:
```suggestion
    const {links, ...data} = input
    
```

libs/navigation/api/src/lib/navigation.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      providers: [NavigationService, {provide: PrismaClient, useValue: prismaMock}]
+    }).compile()
+
+    service = module.get<NavigationService>(NavigationService)
+  })
+
+  it('should get a navigation by id', async () => {
+    prismaMock.navigation.findUnique.mockResolvedValueOnce({})
+    await service.getNavigationById('123')
+    expect(prismaMock.navigation.findUnique).toHaveBeenCalled()
```
Itrulia, line 33: Already tested by line below (same as the other tests)

libs/navigation/api/src/lib/navigation.resolver.ts
```diff
@@ -0,0 +1,56 @@
...
+@Resolver(() => Navigation)
+export class NavigationResolver {
+  constructor(private readonly navigationService: NavigationService) {}
+
+  @Query(() => Navigation, {nullable: true})
+  getNavigation(@Args() {id, key}: NavigationArgs) {
+    if (id) {
+      return this.navigationService.getNavigationById(id)
+    }
+    if (key) {
```
Itrulia, line 27:
```suggestion

        if (key) {
```

libs/navigation/api/src/lib/navigation.resolver.ts
```diff
@@ -0,0 +1,56 @@
...
+      return this.navigationService.getNavigationByKey(key)
+    }
+    throw new BadRequestException('Need to provide id or key')
```
Itrulia, line 30:
```suggestion

    throw new BadRequestException('Need to provide id or key')
```

libs/navigation/api/src/lib/navigation.resolver.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  @Query(() => [Navigation])
+  @Permissions(CanGetNavigations)
+  getNavigations() {
+    return this.navigationService.getNavigations()
+  }
+
+  @Mutation(() => Navigation)
+  @Permissions(CanCreateNavigation)
+  createNavigation(@Args('input') input: CreateNavigationInput) {
```
Itrulia, line 43: Don't use input as an argument please, there's no need to have  `{input: {...}}` and not `{...}`

Itrulia (changes requested):

libs/navigation/api/src/lib/navigation.model.ts
```diff
@@ -0,0 +1,138 @@
...
+  Field,
+  ObjectType,
+  InputType,
+  InterfaceType,
+  ArgsType,
+  OmitType,
+  registerEnumType
+} from '@nestjs/graphql'
+
+export enum NavigationLinkType {
```
Itrulia, line 11: We should probably turn the schema in the DB to be an enum and get that from Prisma.	

instead of `type  String` to be `type NavigationLinkType`

libs/navigation/api/src/lib/navigation.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  async deleteNavigationById(id: string) {
+    return this.prisma.navigation.delete({
+      where: {id},
+      include: {links: true}
+    })
+  }
+
+  async updateNavigation(input: UpdateNavigationArgs) {
+    const {id, links, ...data} = input
+    await this.prisma.navigationLink.deleteMany({
```
Itrulia, line 53: We should fix this that navigation links have to include ids, so we don't have to delete and recreate them everytime.

libs/navigation/api/src/lib/navigation.resolver.ts
Itrulia, line 1: Missing descriptions in queries/mutations

libs/navigation/api/src/lib/navigation.resolver.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        if (key) {
+      return this.navigationService.getNavigationByKey(key)
+    }
+
+    throw new BadRequestException('Need to provide id or key')
+  }
+
+  @Query(() => [Navigation])
+  @Permissions(CanGetNavigations)
```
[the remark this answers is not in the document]
tomaszdurka: Copying from the old behaviour. I agree - it can be removed.

libs/navigation/api/src/lib/navigation.service.spec.ts
tomaszdurka: Its usually to narrow down if it's failed.

Itrulia: But this will tell you it wasn't called or that it has been called, the snapshot tells you if it has been called and with what.

libs/navigation/api/src/lib/navigation.resolver.ts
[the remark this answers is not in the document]
Itrulia: There's a public query for it too, this would break the websites as we rely on that query for the navigation & footer :D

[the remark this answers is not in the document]
tomaszdurka: Ah right this was about getting single navigation (private only).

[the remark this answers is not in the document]
Itrulia: Also exists in public

PR #1318, opened 2024-05-07
fix: various fixes for bajour, pagination for authors/articles

Itrulia:


libs/image/website/src/lib/image.context.tsx
```diff
@@ -0,0 +1,22 @@
+import {BuilderImageProviderProps} from '@wepublish/website/builder'
+import {mergeDeepLeft} from 'ramda'
+import {createContext, useContext, useMemo} from 'react'
+
+export const ImageContext = createContext<BuilderImageProviderProps>({})
```
Itrulia, line 5: Allows now to override props given to the image, this allows one to set square to true and not just set aspect-ratio to 1.

apps/bajour/pages/profile/index.tsx
```diff
@@ -26,7 +26,7 @@ function Profile() {
     <Container>
       <ProfileWrapper>
         <H4 component={'h1'}>Profil</H4>
-        <PersonalDataFormContainer mediaEmail="info@wepublish.dev" />
+        <PersonalDataFormContainer mediaEmail="info@bajour.ch" />
```
antkiewiczk, line 29: 👍

antkiewiczk (approved):
good that you also changed it in website example

PR #1322, opened 2024-05-08
chore(nestjs): Add prisma-client exception filter

tomaszdurka:


tomaszdurka:
@Itrulia @elias-summermatter @antkiewiczk  Would like to get some feedback from you about this. Good or bad idea? I think we can generalise these exceptions into graphql ones. 

I remember @Itrulia saying about the stacktraces. Indeed you loose stacktrace, but here we should not actually have them as it's user input error. We can always log these original exceptions if needed.

Question is if we should actually generalize these exception as bad user input when triggered from API.

tomaszdurka:
@Itrulia what do you think about this change. 
I am actually not sure if we shouldn't rather explicitly control them.

Itrulia:
I guess its okay. Lets see if it has any downside, it's easy to remove later on :)

Itrulia (approved):

PR #1330, opened 2024-05-14
Added fingerprint function to poll vote

elias-summermatter:


michael-scheurer (changes requested):
Nice code.
To make it definitely useful to the end-user, we should have a UI which is able to
- list the votes
- delete votes from that list

Or what is the advantage of having these hashes in the database for upcoming request from Bajour? We would still need to manually delete in the database, no?

michael-scheurer (approved):

PR #1332, opened 2024-05-18
feat(editor,api): migrate article & pages to new tagging system

Itrulia:
Things to test:

- [x] Related articles in website still works
- [x] Article tag pages for articles in website still works
- [x] Creating, Updating, Deleting, Publishing, Duplicating Pages & Articles

BREAKING CHANGE: Article & Page tag went from GraphqlString to GraphqlTag
BREAKING CHANGE: Article & Page filter by tag is using the id of the tag now not the tag name


libs/api/prisma/migrations/20240513152239_migrate_revision_tags/migration.sql
Itrulia, line 1: Has to be in a seperate migration as the enums have to be committed first before usage

libs/api/prisma/migrations/20240513225723_main_tags/migration.sql
Itrulia, line 1: Allows to set a tag as main, which we can show in the editor UI as separate link or use it in URLs for the website.

libs/api/prisma/schema.prisma
```diff
@@ -121,12 +121,25 @@ model Article {
...
   navigations NavigationLink[]
+  tags        TaggedArticles[]
 
   @@index([createdAt])
   @@index([modifiedAt])
   @@map("articles")
 }
 
+model TaggedArticles {
+  article    Article  @relation(fields: [articleId], references: [id], onDelete: Cascade)
```
Itrulia, line 132: Note that tags are now on the article itself

libs/api/src/lib/graphql/blocks.ts
```diff
@@ -552,8 +552,6 @@ export const GraphQLTeaserListBlock = new GraphQLObjectType<TeaserListBlock, Con
               prisma.article
             )
 
-            articles.nodes.forEach(article => loaders.articles.prime(article.id, article))
```
Itrulia, line 555: Unclean but removed from priming loaders as the tag ids are missing

apps/editor/src/app/routes/articles/articleList.tsx
```diff
@@ -99,27 +107,18 @@ function ArticleList() {
     variables: articleListVariables,
     fetchPolicy: 'network-only'
   })
+  const [createComment] = useCreateCommentMutation()
 
   const articles = useMemo(() => data?.articles?.nodes ?? [], [data])
-
-  useEffect(() => {
```
Itrulia, line 105: Happens automatically as variables change

antkiewiczk (approved):

PR #1336, opened 2024-05-20
fix/BA-140-mail-already-exists-with-snackbar

michael-scheurer:


libs/membership/website/src/lib/subscribe/subscribe-container.tsx
```diff
@@ -37,6 +39,7 @@ export const SubscribeContainer = <T extends OptionalKeysOf<RegisterMutationVari
   schema,
   filter
 }: SubscribeContainerProps<T>) => {
+  const router = useRouter()
```
Itrulia, line 42: This has to be outside the website builder.

libs/ui/src/lib/snackbar/snackbar.tsx
```diff
@@ -0,0 +1,33 @@
...
+    return
+  }
+})
+
+export interface Snack {
+  message: string
+  severity?: AlertColor
+}
+
+export function SnackbarWithContextProvider({children, ...props}: SnackbarProps) {
```
Itrulia, line 17:
```suggestion
export function SnackbarWithContextProvider({children, autoHideDuration = 10000, ...props}: SnackbarProps) {
```

Then you don't need DEFAULT_AUTO_HIDE_DURATION

libs/ui/src/lib/snackbar/snackbar.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const DEFAULT_AUTO_HIDE_DURATION = 10000
+  const [snack, setSnack] = useState<Snack | undefined>(undefined)
+
+  return (
+    <SnackbarContext.Provider value={{setSnack}}>
+      {children}
+      <Snackbar
+        {...props}
+        open={!!snack}
+        autoHideDuration={props.autoHideDuration || DEFAULT_AUTO_HIDE_DURATION}
```
Itrulia, line 27:
```suggestion
        autoHideDuration={autoHideDuration}
```

libs/ui/src/lib/snackbar/snackbar.tsx
Itrulia, line 17: Currently this only supports 1 snackbar, maybe we want to do a push on it so we can have multiple. But not important for now of course. Works great!

libs/membership/website/src/lib/subscribe/subscribe-container.tsx
```diff
@@ -69,14 +72,28 @@ export const SubscribeContainer = <T extends OptionalKeysOf<RegisterMutationVari
   })
 
   const [register] = useRegisterMutation({
-    onError: () => challenge.refetch(),
+    onError: async error => {
+      const emailAlreadyInUseError = error.graphQLErrors.find(
```
Itrulia, line 76: Probably this should be checked globally and not in this container. We can check this via Apollo links or error handlers iirc

libs/membership/website/src/lib/subscribe/subscribe-container.tsx
michael-scheurer: That means, I have to pass it through the properties? Any other idea?

libs/ui/src/lib/snackbar/snackbar.tsx
michael-scheurer: yeah. I wanted to implement that. But I gave up on it because of a really quirky behavior of the snackbar component. It is ok for now.

libs/membership/website/src/lib/subscribe/subscribe-container.tsx
michael-scheurer: Apollo Links => Error Links

michael-scheurer:
Closing it in favor of https://github.com/wepublish/wepublish/pull/1395

PR #1339, opened 2024-05-22
BA-95: Bajour quote block

antkiewiczk:


apps/bajour/src/components/quote/bajour-quote.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <BajourQuoteBlockWrapper className={className} withImage={!!image}>
+      {image && (
+        <QuoteImage>
+          <Image image={image} square />
+        </QuoteImage>
+      )}
+      <BajourQuoteContent>
+        <H4 component="p" css={quoteStyles}>
+          « {quote} »
```
Itrulia, line 111:
```suggestion
          {quote}
```

apps/bajour/src/components/quote/bajour-quote.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return (
+    <BajourQuoteBlockWrapper className={className} withImage={!!image}>
+      {image && (
+        <QuoteImage>
+          <Image image={image} square />
+        </QuoteImage>
+      )}
+      <BajourQuoteContent>
+        <H4 component="p" css={quoteStyles}>
```
Itrulia, line 110: You can make a component above via styled and use component instead of CSS here.

apps/bajour/src/components/quote/bajour-quote.tsx
```diff
@@ -0,0 +1,121 @@
...
+          « {quote} »
+        </H4>
+        {author && (
+          <Paragraph component="cite" css={authorStyles} gutterBottom={false}>
```
Itrulia, line 114: same thing here

apps/bajour/src/components/quote/bajour-quote.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  ${({theme}) => theme.breakpoints.up('md')} {
+    margin: 0;
+    margin-left: ${({theme}) => `-${theme.spacing(8)}`};
+    max-width: calc((100% / 12) * 4 + ${({theme}) => theme.spacing(60)});
+  }
+
+  @media (min-width: 1440px) {
+    // breakpoint needs to be custom
```
Itrulia, line 32: Explain why please

antkiewiczk: that's because the component is pushed a bit to the left via negative margin, at least this is what I understood that is required looking at the designs 
Screenshot 2024-05-23 at 14 10 20

apps/bajour/src/components/quote/bajour-quote.tsx
antkiewiczk: but I would need to make this component inside `BajourQuoteBlock` if I want to use components exported from website builder i.e. 

```
  const {
    elements: {H4, Image, Paragraph}
  } = useWebsiteBuilder()
```

apps/bajour/src/components/quote/bajour-quote.tsx
Itrulia: Because bajour doesn't want it

apps/bajour/src/components/quote/bajour-quote.tsx
Itrulia: No you don't: https://github.com/wepublish/wepublish/blob/3b80f2f386ec4f49d1a17ae4d1c8a4906c0afdae/libs/block-content/website/src/lib/teaser/teaser.tsx#L314

apps/bajour/src/components/quote/bajour-quote.tsx
Itrulia: I still don't understand why the breakpoint needs to be custom

apps/bajour/src/components/website-builder-overwrites/block-renderer/block-renderer.tsx
```diff
@@ -2,6 +2,7 @@ import {
   ApiV1,
   BlockRenderer,
   BuilderBlockRendererProps,
+  isQuoteBlock,
```
Itrulia, line 5: use the website builder provider in _app instead.

michael-scheurer:
As discussed:
@antkiewiczk 
- Fix long author names
- Fix font weight on the quote and author name

michael-scheurer (approved):

PR #1342, opened 2024-05-22
chore(v2-migration): Migrate user-role/permissions api into access-control v2

tomaszdurka:
…ntrol v2


libs/access-control/api/src/lib/user-role.model.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  @Field(() => Boolean)
+  systemRole!: boolean
+
+  @Field(() => [Permission])
+  permissions!: Permission[]
+}
+
+@ObjectType()
+export class PageInfo {
```
Itrulia, line 34: Checkout: https://github.com/wepublish/wepublish/blob/master/libs/event/api/src/lib/event.model.ts#L79

libs/access-control/api/src/lib/user-role.model.ts
```diff
@@ -0,0 +1,139 @@
...
+export class CreateUserRoleInput {
+  @Field()
+  name!: string
+
+  @Field(() => [String])
+  permissionIDs!: string[]
+}
+
+@ArgsType()
+export class CreateUserRoleArgs {
```
Itrulia, line 118: Not needed and we can use CreateUserRoleInput directly as args type.

libs/access-control/api/src/lib/user-role.model.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  @Field({nullable: true})
+  name?: string
+
+  @Field(() => [String], {nullable: true})
+  permissionIDs?: string[]
+}
+
+@ArgsType()
+export class UpdateUserRoleArgs {
```
Itrulia, line 136: Same here as above

libs/access-control/api/src/lib/user-role.service.ts
```diff
@@ -0,0 +1,78 @@
...
+  createUserRole(data: CreateUserRoleInput) {
+    return this.prisma.userRole.create({
+      data: {...data, systemRole: false}
+    })
+  }
+
+  async updateUserRole({id, ...data}: UpdateUserRoleInput) {
+    const role = await this.prisma.userRole.findUnique({where: {id}})
+    if (role?.systemRole) {
+      throw new Error('Cannot change SystemRoles')
```
Itrulia, line 63:
```suggestion
      throw new Error('Cannot change a system role')
```

libs/access-control/api/src/lib/user-role.service.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this.prisma.userRole.update({
+      where: {id},
+      data
+    })
+  }
+
+  async deleteUserRoleById(id: string) {
+    const role = await this.prisma.userRole.findUnique({where: {id}})
+    if (role?.systemRole) {
+      throw new Error('Cannot delete SystemRoles')
```
Itrulia, line 74:
```suggestion
      throw new Error('Cannot delete a system role')
```

libs/access-control/api/src/lib/user-role.resolver.ts
```diff
@@ -0,0 +1,52 @@
...
+import {Permissions} from '@wepublish/permissions/api'
+
+@Resolver(() => UserRole)
+export class UserRoleResolver {
+  constructor(private readonly userRoleService: UserRoleService) {}
+
+  @Query(() => UserRole, {description: 'Returns a user role'})
+  @Permissions(CanGetUserRole)
+  getUserRoleById(@Args() args: UserRoleIdArgs) {
+    return this.userRoleService.getUserRoleById(args.id)
```
Itrulia, line 26: Missing dataloader

libs/access-control/api/src/lib/user-role.service.ts
```diff
@@ -0,0 +1,78 @@
+import {Injectable} from '@nestjs/common'
+import {PrismaService} from '@wepublish/nest-modules'
+import {CreateUserRoleInput, GetUserRolesArgs, UpdateUserRoleInput} from './user-role.model'
+import {getMaxTake} from '@wepublish/utils/api'
+
+@Injectable()
+export class UserRoleService {
```
Itrulia, line 7: Missing dataloader decorators

libs/utils/api/src/lib/graphql/max-limit.ts
```diff
@@ -1,3 +1,3 @@
 const MaxResultsPerPage = 100
 
-export const getMaxTake = (take: number) => Math.min(take, MaxResultsPerPage)
+export const getMaxTake = (take?: number) => Math.min(take ?? MaxResultsPerPage, MaxResultsPerPage)
```
Itrulia, line 3: Don't think this is the default we want or the place to set a default.

libs/access-control/api/src/lib/access-control-api.module.ts
Itrulia, line 1: Don't think the naming of the library/module is consistent with the rest of the application

PR #1344, opened 2024-05-23
BA-25: Show mailchimp popup when popup=true in url

antkiewiczk:


apps/bajour/pages/_app.tsx
Itrulia, line 1: We should use https://nextjs.org/docs/pages/api-reference/components/script here and no need for a useEffect.

apps/bajour/pages/_app.tsx
```diff
@@ -80,8 +82,24 @@ const ButtonLink = styled('a')`
...
+  }
+}
+
 function CustomApp({Component, pageProps}: CustomAppProps) {
   const siteTitle = 'Bajour'
+  const router = useRouter()
+
+  useEffect(() => {
+    const {popup} = router.query
+    if (popup === 'true') {
```
Itrulia, line 99: Can probably just check if the query param exists.

apps/bajour/pages/_app.tsx
antkiewiczk: should be fine now

Itrulia (approved):

PR #1346, opened 2024-05-27
feat(api,editor): add image to payment method

Itrulia:


libs/membership/editor/src/lib/memberplan/memberplan-edit.tsx
```diff
@@ -97,7 +97,10 @@ function MemberPlanEdit() {
     setAvailablePaymentMethods(
       (initMemberPlan?.availablePaymentMethods || []).map(availablePaymentMethod => ({
         id: generateID(),
-        value: availablePaymentMethod
+        value: {
+          ...availablePaymentMethod,
+          paymentMethods: availablePaymentMethod.paymentMethods as PaymentMethod[]
```
Itrulia, line 102: Because we only have an imageRef and not Image.

antkiewiczk (approved):

PR #1353, opened 2024-05-30
chore(field-permissions): Implement FieldPermissions (allow to nullify field if you have no permissions)

tomaszdurka:
…y field if you have no permissions)


libs/member-plan/api/src/lib/member-plan.model.ts
```diff
@@ -32,7 +33,7 @@ export class MemberPlan {
   @Field()
   description!: string
 
-  @Field()
+  @FieldProtected(() => Boolean, {permissions: [CanGetMemberPlan]})
```
Itrulia, line 36: Can probably always be returned and only query active memberplans for users with the perms instead

libs/navigation/api/src/lib/navigation.model.ts
```diff
@@ -33,7 +34,7 @@ export class Navigation {
   @Field()
   key!: string
 
-  @Field(() => [BaseNavigationLink])
+  @FieldProtected(() => [BaseNavigationLink], {permissions: [CanGetNavigations]})
```
Itrulia, line 37: Should probably always be returned :D

libs/member-plan/api/src/lib/member-plan.model.ts
Itrulia: I wonder if we can introspect the type for boolean, number etc, so we don't always have to specify.

libs/permissions/api/src/lib/field-permissions.middleware.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const FieldMiddlewarePermissions: FieldMiddleware = async (ctx, next) => {
+  const {info} = ctx
+  const {extensions} = info.parentType.getFields()[info.fieldName]
+  const permissions = extensions.permissions as Permission[] | undefined
+  const session = ctx.context.req.user
+
+  if (!permissions) {
+    return next()
+  }
```
Itrulia, line 13:
```suggestion
  }
  
```

libs/navigation/api/src/lib/navigation.model.ts
tomaszdurka: Yes indeed, just added this to illustrate.

libs/member-plan/api/src/lib/member-plan.model.ts
tomaszdurka: Agree, it's not really harmful to have it.

PR #1359, opened 2024-06-10
Added cache headers to frontends

elias-summermatter:


Itrulia (commented):
https://github.com/wepublish/wepublish/blob/master/libs/utils/website/src/lib/next.config.js

if you add it there, you only have to do it in 1 place

elias-summermatter:
@Itrulia moved it :-) nice change

Itrulia (approved):

PR #1368, opened 2024-06-13
chore(migrate-v2-block): Introduce block models

tomaszdurka:


libs/block/api/src/lib/model/event.ts
```diff
@@ -0,0 +1,44 @@
...
+  filter!: EventBlockFilter
+
+  @Field(() => [Event])
+  events!: Event[]
+}
+
+// Inputs
+
+@InputType()
+export class EventBlockInputFilter {
```
Itrulia, line 29: The same thing as above, could probably re-use that. 
That's also the case for a lot of the other block models.

PR #1371, opened 2024-06-17
Editor/API changes from Tsri branch

Itrulia:
- fix(editor): poll bug with default answers
- fix(editor): tag picker not working with pagination


libs/ui/editor/src/lib/atoms/poll/pollAnswers.tsx
```diff
@@ -231,7 +233,7 @@ export function PollAnswers({poll, onPollChange}: PollAnswersProps) {
                 content={`${getTotalVotesByAnswerId(poll, answer.id)} ${t('pollAnswer.votes')}`}>
                 <Form.Control
                   name={`answer-${answer.id}`}
-                  value={answer.answer || t('pollEditView.defaultAnswer')}
```
Itrulia, line 234: Was constantly resetting value while typing.

libs/ui/editor/src/lib/atoms/tag/selectTags.tsx
Itrulia, line 1: Fixes that it properly works with paginated data.

elias-summermatter (approved):

PR #1374, opened 2024-06-18
feat(membership-website): redesigned subscribe page

Itrulia:
1. Hide auto renew when forceAutorenew is set
2. Membership picker now fully hides when no content is present and it's the only one
3. Paymentmethod picker now supports images
4. Bonus money is gone and now part of the slider which shows how much you are paying


libs/membership/website/src/lib/memberplan-picker/memberplan-picker.tsx
```diff
@@ -38,6 +39,8 @@ export const MemberPlanPicker = forwardRef<HTMLButtonElement, BuilderMemberPlanP
 
     const showRadioButtons = memberPlans.length > 1
     const selectedMemberPlan = memberPlans.find(({id}) => id === value)
+    const showPicker =
```
Itrulia, line 42: Hide picker fully when no image, no description & only 1 memberplan exists.

Have to use toPlaintext as it can have empty children.

antkiewiczk (approved):

PR #1384, opened 2024-06-22
feat(tsri): change quote block design

Itrulia:


libs/content/website/src/lib/content-wrapper/content-wrapper.tsx
```diff
@@ -57,7 +57,7 @@ export const ContentWrapperStyled = styled('article')<{fullWidth?: boolean}>`
         }
 
         & > ${QuoteBlockWrapper} {
-          grid-column-start: 5;
+          /* grid-column-start: 5; */
```
antkiewiczk, line 60: should this be left like this?

antkiewiczk (approved):

libs/content/website/src/lib/content-wrapper/content-wrapper.tsx
Itrulia: Oh good catch thanks! Was toggling it on and off to check if the margin was correct.

PR #1393, opened 2024-06-24
Migrate FE to selfhosted

elias-summermatter:


.github/workflows/on-pr-review.yml
```diff
@@ -274,6 +275,86 @@ jobs:
...
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up matrix
+        id: set-matrix
+        run: |
+          targets=$(find apps -mindepth 2 -maxdepth 2 -type f -name .frontend_deployment -exec dirname {} \; | xargs -n 1 basename)
```
Itrulia, line 289: probably better to use nx, give the project tags and then you can get all the target that way, no?

elias-summermatter: Also i would need to install nx and node right?

Itrulia: I guess yeah you would need node.


We can add tags in the projects.json and get them like: `nx show projects --projects tag:<TAG_YOU_GAVE>`

.github/workflows/on-demand-deploy-helm-frontend.yml
```diff
@@ -0,0 +1,64 @@
...
+          $(echo -e "${{ steps.get-env.outputs.envvars }}" | awk '{print "    "$0}')
+          frontend:
+            image: ghcr.io/wepublish/frontend-${{ inputs.frontend_project }}:${{ github.sha }}
+            env:
+              NODE_ENV: production
+          EOF
+          cat values.yaml
+          export POD_NAME=review-${{ github.event.pull_request.number }}-${{ inputs.frontend_project }}-frontend
+          if [[ ! -z $(helm ls -a -n reviews |grep ${POD_NAME} |grep -v "deployed ") ]]; then helm uninstall -n reviews ${POD_NAME}; fi
+          helm upgrade --install ${POD_NAME} ./helm/charts/wepublish-frontend -n reviews -f values.yaml --wait
```
tomaszdurka, line 64: missing newline eof :)

tomaszdurka (commented):
@elias-summermatter that's some piece of hacking and seds ! :) GJ

Looks good to me.
A few points:
- I would rename `frontend` to `website` everywhere as it's misleading due to editor also being frontend. And it is really only about the `website`.
- Now when thinking about the whole tag/deployment per customer I see we will have no clue what API version customer is on. Maybe version should include API version, then client-iteration. Like 10.2.3-tsri-20. Or have separate tags, but this misses the point of simple, unified release/deployment.

Best,
Tomasz

.github/workflows/on-demand-deploy-project-database.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Restore Dump to Review Database
+        env:
+          PGPASSWORD: foo
+        run: |
+          psql -U postgres -h 127.0.0.1 -p 5433 -d wepublish -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public;"
+          pg_restore -U postgres -h 127.0.0.1 -p 5433 -d wepublish -v db.dump
+
+      - name: Cleanup
+        run: rm db.dump
```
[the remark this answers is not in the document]
elias-summermatter: yes correct

PR #1394, opened 2024-06-25
fix(website): bildwurf breaking application on navigation

Itrulia:


libs/block-content/website/src/lib/bildwurf-ad/bildwurf-ad-block.tsx
```diff
@@ -40,9 +40,18 @@ export function BildwurfAdBlock({zoneID, className}: BuilderBildwurfAdBlockProps
     <>
       <Script src="https://media.online.bildwurf.ch/js/code.min.js" onLoad={loadAd} />
 
-      <BildwurfBlockWrapper className={className} id="bildwurf-injection-wrapper">
-        <ins className="aso-zone" data-zone={zoneID}></ins>
-      </BildwurfBlockWrapper>
+      <BildwurfBlockWrapper
```
Itrulia, line 43: Similar to this: https://github.com/facebook/react/issues/11538#issuecomment-417504600

React tracks the div & contents and Bildwurf changes the structure. Thus the reference is gone. By injecting it like this, React is not tracking its contents.

antkiewiczk: 👌

antkiewiczk (approved):

PR #1402, opened 2024-06-28
Prompt users to login for commenting

Itrulia:


libs/comments/website/src/lib/comment-editor/comment-editor.tsx
```diff
@@ -1,16 +1,24 @@
...
+import {setCookie} from 'cookies-next'
+import {add} from 'date-fns'
+import {useMemo, useState} from 'react'
 import {Controller, useForm} from 'react-hook-form'
-import {MdSend} from 'react-icons/md'
+import {MdClose, MdLogin, MdSend} from 'react-icons/md'
 import {z} from 'zod'
 
-export const CommentEditorWrapper = styled('form')`
+export const IntendedRouteStorageKey = 'auth.intended'
```
Itrulia, line 14: Why is this c&p? IntendedRouteStorageKey could be moved to authentication-website and then removed from utils-website.

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
```diff
@@ -109,100 +195,147 @@ export const CommentEditor = ({
...
+            <MdClose />
+          </CloseLogin>
+          <LoginWrapper>
+            <LoginFormContainer afterLoginCallback={handleAfterLoginCallback} />
+            <Register>
+              <Button css={buttonStyles} onClick={registerRedirect}>
+                <IconButton css={iconStyles} size="large">
+                  <MdLogin aria-label="Register" />
+                </IconButton>
+                <Link href="/signup" css={aStyles}>
```
Itrulia, line 331: /signup doesn't always exist and we need a way to configure the route. For example Tsri would want to use /mitmachen as they don't have a register url.

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
```diff
@@ -109,100 +195,147 @@ export const CommentEditor = ({
...
+      <Modal open={modalOpen} onClose={handleClose}>
+        <ModalContent>
+          <CloseLogin onClick={handleClose}>
+            <MdClose />
+          </CloseLogin>
+          <LoginWrapper>
+            <LoginFormContainer afterLoginCallback={handleAfterLoginCallback} />
+            <Register>
+              <Button css={buttonStyles} onClick={registerRedirect}>
+                <IconButton css={iconStyles} size="large">
```
Itrulia, line 328: Button is nested in button. Thats not valid HTML.

You can simply use Button and give it a start icon.

libs/website/builder/src/lib/comment.interface.ts
```diff
@@ -135,6 +135,8 @@ export type BuilderCommentEditorProps = {
   maxCommentLength: number
   loading: boolean
   error?: ApolloError
+  canReply?: boolean
```
Itrulia, line 138: Should not be optional

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
Itrulia: image

Itrulia:
image

Layout of this is broken and probably also what it should do. Bajour has anonymous commenting enabled, once the bug is fixed it will not show up anymore.

Also it might make sense to put the register below and not to the side, either as button or as muted text with a link.

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
antkiewiczk: do you mean adding it to env?

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
antkiewiczk: I understand not wanting to copy and paste the same code multiple times, but I don't understand the second part about moving it to authentication-website?

antkiewiczk:
> image
> 
> Layout of this is broken and probably also what it should do. Bajour has anonymous commenting enabled, once the bug is fixed it will not show up anymore.
> 
> Also it might make sense to put the register below and not to the side, either as button or as muted text with a link.

@michael-scheurer what do you think about this?

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
Itrulia: No to make it configureable. Take a look at the WebsiteBuilder Context.

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
Itrulia: How so? Currently it lies within utils-website which you can not use in the website builder since it's for the apps.

libs/comments/website/src/lib/comment-editor/comment-editor.spec.tsx
```diff
@@ -1,13 +1,23 @@
...
 
+// Compose stories
 const storiesCmp = composeStories(stories)
 
 describe('CommentEditor', () => {
   Object.entries(storiesCmp).forEach(([story, Component]) => {
     it(`should render ${story}`, () => {
-      const {asFragment} = render(<Component />)
+      const {asFragment} = render(
+        <MockedProvider>
```
Itrulia, line 15: ThemeProvider is already in component and don't think we need the ApolloProvider here, do we?

Itrulia (commented):
@antkiewiczk comment tests still don't make sense anymore as they don't do what the names say. Please change them to work as they used to.

antkiewiczk:
> @antkiewiczk comment tests still don't make sense anymore as they don't do what the names say. Please change them to work as they used to.

so the rest of the stuff (apart from tests) is ok?

Itrulia:
Yep!

libs/comments/website/src/lib/comment-editor/comment-editor.spec.tsx
antkiewiczk: after removing the tests are failing

libs/comments/website/src/lib/comment-editor/comment-editor.stories.tsx
Itrulia, line 1: No tests where canReply is set or anonymousCanComment is set to false etc.

libs/comments/website/src/lib/comment-editor/comment-editor.stories.tsx
```diff
@@ -134,7 +184,7 @@ export const LoggedIn: StoryObj = {
     ...Anonymous.args,
     challenge: null
   },
-  decorators: [WithUserDecorator({} as any)]
+  decorators: [WithUserDecorator(mockUser)]
```
Itrulia, line 187: Don't think that's needed, only hasUser is used.

Itrulia:
image

Share and editing should probably switch places

**libs/comments/website/src/lib/comment-list/comment-list.stories.tsx**

1. It looks like clicking on a button to answer to a comment doesn't bring up the modal but clicking on the root button does.  I am logged in so it should not bring it up at all. (check Commenting or Commenting Open. Note that the interactions are currently broken in the browser because of an outstanding Nx/Next/Storybook bug)
2. It does the same thing when I am not logged in check `AnonymousCommenting`
3. It looks like there's no way to leave an anonymous comment, check `AnonymousCommenting`  
4. `AnonymousCommentingOpen` does not click on commenting anonymous
5. Why is the share button tied to the user being able to leave a comment?

**libs/comments/website/src/lib/comment-list/comment-list-item.stories.tsx**

^ Same general issues as above (1,2,3)

libs/comments/website/src/lib/comment-editor/comment-editor.stories.tsx
antkiewiczk: when anonymousCanComment is not set to true it behaves as it's set to false

antkiewiczk: anyway, added two more cases for clarity - anonymousCanComment set to true and false

elias-summermatter (approved):

PR #1404, opened 2024-06-28
refactor(comment-website): refactor styles, date & highlight of comment

Itrulia:
use native :target instead of using JS to check if a comment is highlighted
use website builder date format
only apply padding to comment when highlighted so button & text are not misalignedt
fixed bug that if flair does not exist, source is not clickable


antkiewiczk:
Please update the branch with master and update snapshots (there has been quite few changes related with testing from Michael)

PR #1405, opened 2024-07-01
enh/minor-test-improvements

michael-scheurer:


package.json
```diff
@@ -12,15 +12,15 @@
     "lint": "nx run-many --target=lint",
     "build": "nx run-many --targets=build",
     "i18n": "node_modules/i18next-parser/bin/cli.js --config i18next-parser.config.js",
-    "test": "nx affected:test --base master --parallel=false",
-    "test-all": "nx run-many --target=test --parallel=10 --runInBand",
+    "test": "nx run-many --target=test --parallel=false --runInBand",
```
michael-scheurer, line 15: Unfortunately adding `runInBand` to the project.json files didn't solve the problem. Will stick with the argument here. Think this is a pretty straight forward command for newbies or PO :)

Itrulia (approved):

PR #1408, opened 2024-07-02
feat(mannschaft): add main tag to article url

Itrulia:


libs/utils/website/src/lib/prerender-teaser-paths.tsx
```diff
@@ -52,11 +52,16 @@ export const getArticlePathsBasedOnPage =
...
+    const articleSlugs = [] as {slug: string; tag?: string}[]
 
     for (const storeObj of cache) {
       if (storeObj?.__typename === 'Article' && !(storeObj as ApiV1.Article).peeredArticleURL) {
-        articleSlugs.push((storeObj as ApiV1.Article).slug)
+        const article = storeObj as ApiV1.Article
+
+        articleSlugs.push({
+          slug: article.slug,
+          tag: article.tags.find(tag => tag.main)?.tag ?? undefined
```
Itrulia, line 63: image

checked it still properly pre-renders even if only slug is present (for example for tsri)

antkiewiczk (approved):
looks good however some tests are failing and it seems it's connected to your changes

PR #1409, opened 2024-07-02
SHARE-9: Allow injecting complete scripts into the pages

LukasSkywalker:
This change allows users to not only inject script contents (or pure JS code) into the pages, but to inject full scripts, including the wrapping <script> tag. This way, attributes defined on the tag itself are preserved, such as the data-id and the defer attributes of <script data-id="1234" defer></script>.

In order for the scripts to actually be run, the script must be recreated in the frontend and all attributes must be copied. This is a bit ugly, but seems to be the only way to preserve all original element data.


libs/utils/website/src/lib/inject-script.tsx
```diff
@@ -0,0 +1,45 @@
...
+  the copy using appendChild.
+  */
+  useEffect(() => {
+    if (data?.setting?.value) {
+      // Create a temporary div to parse the script tag
+      const temp = document.createElement('div')
+      temp.innerHTML = data.setting.value
+      const children = temp.children
+
+      for (let i = 0; i < children.length; i++) {
```
antkiewiczk, line 24: maybe it would be helpful here to add some additional security measures like sanitization or csp rules?

libs/utils/website/src/lib/inject-script.tsx
```diff
@@ -0,0 +1,45 @@
+import { ApiV1 } from '@wepublish/website'
+import {useEffect} from 'react'
+
+export function InjectScript() {
+  const {data} = ApiV1.useSettingQuery({
```
antkiewiczk, line 5: would also be nice to get error and handle it somehow

libs/utils/website/src/lib/inject-script.tsx
```diff
@@ -0,0 +1,45 @@
...
+  /* Because of the WhatWG HTML specs, article 4.12.1: "When inserted using the
+  innerHTML and outerHTML attributes, they do not execute at all.", the scripts
+  must be injected differently. The below is one such method and works by
+  copying the entire element (tagname, attributes and content) and injecting
+  the copy using appendChild.
+  */
+  useEffect(() => {
+    if (data?.setting?.value) {
+      // Create a temporary div to parse the script tag
+      const temp = document.createElement('div')
```
antkiewiczk, line 20: maybe wrap this with try catch and add some fallback console.error or return null? just in case?

try {
  current code
} catch(e) {
console.error or return null
}

libs/utils/website/src/lib/inject-script.tsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+  })
+
+  /* Because of the WhatWG HTML specs, article 4.12.1: "When inserted using the
+  innerHTML and outerHTML attributes, they do not execute at all.", the scripts
+  must be injected differently. The below is one such method and works by
+  copying the entire element (tagname, attributes and content) and injecting
+  the copy using appendChild.
+  */
+  useEffect(() => {
```
Itrulia, line 17: Can not be server rendered now

Maybe a bit of a crazy idea but could we store a JSON in the database like:

```
[
  {
     "src": "https://example.com/example.js",
     "data-id": "1234",
     "async": true
   }
]
```

But display the HTML in the editor (and parse from pasted HTML)? This would allow the frontend to take the JSON and properly render the script tags on the server (use the `Script` component in the website builder).

I think this is the cleanest in terms of structured data, the frontend then also doesn't rely on browser APIs (like createElement) so we can run it on the server too.

libs/utils/website/src/lib/inject-script.spec.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react'
+import { render, waitFor } from "@testing-library/react";
+import { InjectScript } from './inject-script'
+import '@testing-library/jest-dom'
+import { ApiV1 } from '@wepublish/website';
+
+jest.mock('@wepublish/website', () => ({
```
Itrulia, line 7: Use Apollo MockProvider instead.

libs/utils/website/src/lib/inject-script.tsx
Itrulia, line 1: Should probably live inside the website builder. Good question though in what library. I guess we could make a new one called scripts/website

PR #1415, opened 2024-07-07
fix(api-example): seeding of images

Itrulia:
The upload is now timing out, rest works

### DIFF
--- a/apps/media/src/app/app.controller.ts
+++ b/apps/media/src/app/app.controller.ts
@@ -43,6 +43,7 @@ export class AppController {
     uploadedFile: Express.Multer.File
   ) {
     const imageId = uuidv4()
+    console.log(imageId)
     const metadata = await this.media.saveImage(imageId, uploadedFile.buffer)
 
     res.status(201).send({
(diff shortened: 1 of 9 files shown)


libs/media/api/src/lib/media-service/supported-images-validator.ts
```diff
@@ -6,6 +6,7 @@ const supportedMimeTypes = [
   'image/png',
   'image/gif',
   'image/jpeg',
+  'image/jpg',
```
Itrulia, line 9: Funny enough thats the wrong mimetype for jpegs, but some images use that one

apps/media/src/app/app.controller.ts
```diff
@@ -43,6 +43,7 @@ export class AppController {
     uploadedFile: Express.Multer.File
   ) {
     const imageId = uuidv4()
+    console.log(imageId)
```
Itrulia, line 46: a console.log to see if I am hitting the controller (which I don't right now)

michael-scheurer (approved):
Solves the seeding on my local machine too <3

Itrulia:
@michael-scheurer wait it works for you? This doesn't do it for me as the uploading of images takes ages.
I wanted to try a different way by spawning the server in a child process via `execa` and killing it afterwards.

Itrulia:
@michael-scheurer ping

PR #1416, opened 2024-07-08
BA-43-139: Add importer scripts for baslerin des tages

LukasSkywalker:
This change adds an importer script for different directus contents. It can be run from the editor and is idempotent.


LukasSkywalker:
This has been merged into #1417

PR #1417, opened 2024-07-08
BA-44: Add single view of Baslerin des Tages

LukasSkywalker:
This adds the single view for the Baslerin des Tages page. It loads the required fields from the article and displays them in a custom layout.


apps/bajour/src/components/baslerin-des-tages/baslerin-des-tages.tsx
Itrulia, line 1: Hey hey, could we use the template string style like we have in the rest of the application?

apps/bajour/src/components/baslerin-des-tages/baslerin-des-tages.tsx
```diff
@@ -0,0 +1,206 @@
...
+  fontSize: '1.3em',
+  fontWeight: 700,
+  marginBottom: '1em'
+}))
+
+const ImageWrapperMobile = styled(OnlyMobile)(({theme}) => ({
+  display: 'grid',
+  gridTemplateColumns: '50vw repeat(6, 25vw)',
+  gridTemplateRows: '1fr 1fr',
+  gridGap: '0.8em',
```
Itrulia, line 96: grid-gap is a deprecated css property and superseded by the `gap` property.
We should probably also use the theme spacings here etc

apps/bajour/src/components/baslerin-des-tages/baslerin-des-tages.tsx
Itrulia, line 1: Is there a reason we create a new page URL instead of using the existing article URL? 
The BdT component can mostly be using the Article interface from the website builder and can be conditionally overridden.

apps/bajour/src/components/baslerin-des-tages/baslerin-des-tages.tsx
```diff
@@ -0,0 +1,206 @@
+import {styled} from '@mui/material'
+import {ApiV1, isImageBlock, isRichTextBlock, isTitleBlock, RichTextBlock} from '@wepublish/website'
+import {useWebsiteBuilder} from '@wepublish/website'
+import {format} from 'date-fns'
+import {de} from 'date-fns/locale'
```
Itrulia, line 5: No need to import this as it's already set in _app

PR #1441, opened 2024-07-23
feat: added currency to memberplan & payments

Itrulia:
Bexio isn't supported yet. Probably a biut trickier to do so.


libs/api/prisma/migrations/20240723095847_add_currency_to_memberplan/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- CreateEnum
+CREATE TYPE "Currency" AS ENUM ('CHF', 'EUR');
+
+-- AlterTable
+ALTER TABLE "member.plans" ADD COLUMN "currency" "Currency" NOT NULL DEFAULT 'CHF';
```
Itrulia, line 5: Setting default but removing it later on. This is so that we can't forget to set it but still have all the old memberPlans  set to CHF.

libs/api/prisma/migrations/20240723095847_add_currency_to_memberplan/migration.sql
```diff
@@ -0,0 +1,9 @@
...
+
+-- AlterTable
+ALTER TABLE "member.plans" ALTER COLUMN "currency" DROP DEFAULT;
```
elias-summermatter, line 8: Also add currency to the subscription and invoice table and copy it on subscription creation. This is important to ensure that when a admin changes the currency on member plan level, the currency says the same for existing subscriptions and invoices to keep the record clean. Also you can then remove a lot of database queries you added in the code to get the member plan because the subscription is already available.

libs/api/src/lib/context.ts
```diff
@@ -1009,6 +1009,20 @@ export async function contextFromRequest(
         throw new Error('paymentProvider not found')
       }
 
+      const memberPlan = await prisma.memberPlan.findFirst({
```
elias-summermatter, line 1012: here you can get the currency direct from the invoice

libs/api/src/lib/graphql/payment/payment.private-mutation.ts
```diff
@@ -24,8 +25,17 @@ export const createPaymentFromInvoice = async (
   const paymentProvider = paymentProviders.find(pp => pp.id === paymentMethod?.paymentProviderID)
 
   const invoice = await invoicesByID.load(invoiceID)
+  const memberPlan = await memberPlanClient.findFirst({
```
elias-summermatter, line 28: here you can get the currency direct from the invoice

libs/api/src/lib/memberContext.ts
```diff
@@ -374,9 +377,29 @@ export class MemberContext implements MemberContextInterface {
       }
     })
 
+    const memberPlan = await this.prisma.memberPlan.findFirst({
```
elias-summermatter, line 380: here you can get the currency direct from the invoice

libs/membership/api/src/lib/subscription/subscription.service.ts
```diff
@@ -403,6 +403,10 @@ export class SubscriptionService {
       throw new NotFoundException('Subscription not found!')
     }
 
+    if (!invoice.subscription.memberPlan) {
```
elias-summermatter, line 406: here you know that you have the currency if it is on the invoice

libs/payment/api/src/lib/payment-provider/payrexx-subscription-payment-provider.ts
```diff
@@ -132,17 +139,39 @@ export class PayrexxSubscriptionPaymentProvider extends BasePaymentProvider {
     // Find external id property and fail if subscription has been deactivated
     const properties: MetadataProperty[] = props.subscription.properties
     const isPayrexxExt = properties.find(sub => sub.key === 'payrexx_external_id')
+
     if (!isPayrexxExt) {
       throw new Error(`Payrexx Subscription Id not found on subscription ${props.subscription.id}`)
     }
+
+    const currency = (
```
elias-summermatter, line 147: Here take currency from subscription not from Memberpalan

elias-summermatter (changes requested):

michael-scheurer (changes requested):
In very short: Can't create a new subscription plan in the editor. Saving isn't possible anymore.

image

apps/gruppetto/src/yearly-memberplan-item.tsx
```diff
@@ -20,7 +24,8 @@ export const YearlyMemberPlanItem = forwardRef<HTMLButtonElement, BuilderMemberP
           <MemberPlanItemName>{name}</MemberPlanItemName>
 
           <MemberPlanItemPrice>
-            Ab {formatChf(Math.ceil((amountPerMonthMin / 100) * 12))} pro Jahr
+            Ab {formatCurrency(Math.ceil((amountPerMonthMin / 100) * 12), currency, locale)} pro
+            Jahr
```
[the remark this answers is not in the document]
Itrulia: The other websites don't override this

libs/api/__tests__/api/private/index.ts
```diff
@@ -436,6 +436,11 @@ export type CreatedToken = {
   token: Scalars['String'];
 };
 
+export enum Currency {
+  Chf = 'CHF',
```
tomaszdurka, line 440: Why this is camelcase? After fighting with enums in federation I would prefer all of them are same as values.
Is this only generator?

Itrulia: This is generated and generated enums are always camelcase by the graphql generator

elias-summermatter (approved):

PR #1453, opened 2024-07-29
fix(bajour)/vote-by-url

michael-scheurer:


apps/bajour/src/components/frage-des-tages/fdt-poll-block.tsx
```diff
@@ -140,15 +140,19 @@ export const FdtPollBlock = ({poll}: {poll?: ApiV1.PollBlock['poll']}) => {
 
   const autoVote = useCallback(async () => {
```
michael-scheurer, line 141: Wondering, how we could move this auto-vote functionality to poll block provider to make url voting available to all media by default.
We could add this to the Publishers Documentation. By now the provider would miss the pollId. Any idea?

apps/bajour/src/components/frage-des-tages/fdt-poll-block.tsx
```diff
@@ -140,16 +139,18 @@ export const FdtPollBlock = ({poll}: {poll?: ApiV1.PollBlock['poll']}) => {
...
     const answerId = router.query.answerId as string
-    if (answerId) {
-      callAction(async () => {
-        await vote({
-          variables: {
-            answerId
-          }
-        })
-      })()
+    if (!answerId || !poll?.id) {
```
Itrulia, line 142: No need to check for poll.id

libs/block-content/website/src/lib/poll/poll-block.provider.tsx
```diff
@@ -33,6 +36,29 @@ export function PollBlockProvider({children}: PropsWithChildren) {
     [settings?.settings]
   )
 
+  async function vote(
```
Itrulia, line 39: I am a bit conflicted about this function as IMO, I understand **why** this was created.

It not doing anything when a user can not vote is problematic design as it's impure in this case, but it's inherently impure because it calls an API. It also causes a double check if a user can vote as the UI already does this.

Double checking is a problem because the UI & this function may get out of sync, for example this function isn't correct as it tries to vote on polls that are closed unlike the UI that does check for that.

Imo the canVote probably should be passed down, and this function should **not** safeguard, but the one calling vote should. It's essentially a 

```
if (canVote) {
  await vote(...)
}
```

in the useEffect but not in the UI as that already safeguards against it by not showing the buttons. I think that's a lot cleaner in design and fixes there not being a single source of truth if a user can vote or not.

apps/bajour/src/components/frage-des-tages/fdt-poll-block.tsx
Itrulia: I would not add this to a provider as it's a side-effect and would be very hidden and thus hard to debug or requires knowledge.

I would rather have a central place where we would handle all of these URL shenanigans. Best to wait until we have the app router so we can do these things on the server side not the client. Now that we lost the ability to do proper middleware as we are not using AWS/Vercel.

apps/bajour/src/components/frage-des-tages/fdt-poll-block.tsx
```diff
@@ -140,16 +139,18 @@ export const FdtPollBlock = ({poll}: {poll?: ApiV1.PollBlock['poll']}) => {
...
-  }, [callAction, router.query.answerId, vote])
+    await vote(
+      {
+        variables: {
+          answerId
+        }
+      },
+      poll.id
+    )
+  }, [router.query.answerId])
```
Itrulia, line 153: Pretty sure the linter complains here, please fix this!

libs/block-content/website/src/lib/poll/poll-block.provider.tsx
Itrulia, line 39: Should probably also be memoized.

Itrulia (changes requested):
.

apps/bajour/src/components/frage-des-tages/fdt-poll-block.tsx
michael-scheurer: Linter didn't complain. Just let run again to make sure. Why should it fail IYO?

Itrulia: Linter does complain but throws a warning. The effect has a bug without all dependencies because if a dependency changes, the effect doesnt run

michael-scheurer: ah yeah. having all dependencies lead to endless automatically voting. So, refactoring will be necessary.

michael-scheurer:
As stated by @elias-summermatter I'll address your review @Itrulia. The changes have been deployed as a late night hot-fix for Bajour.

apps/bajour/src/components/frage-des-tages/fdt-poll-block.tsx
Itrulia: memoizing as in the other comment stated would fix that as every render the function is recreated, which causes the effect to run, which causes a rerender which causes the function to be recreated etc.

PR #1458, opened 2024-08-04
fix(BAJ)/fdt-details

michael-scheurer:


libs/comments/website/src/lib/comment/comment.tsx
```diff
@@ -92,7 +91,6 @@ export const Comment = ({
 
   const flair = user?.flair || source
   const image = user?.image ?? guestUserImage
-  const isVerified = authorType === CommentAuthorType.VerifiedUser
```
elias-summermatter, line 95: @michael-scheurer i think the badge is not about if someone has a subscription, it's about that someone is signed in, and I think that's a good feature. I would suggest that we could change this that we can set this on a “per project” bases differently and not just remove it. What do you think?

elias-summermatter (changes requested):

elias-summermatter (approved):

PR #1459, opened 2024-08-04
GRUP-13/change-address-naming

michael-scheurer:


elias-summermatter (changes requested):
Tests ar failing please update snapshots

elias-summermatter (approved):

PR #1461, opened 2024-08-05
f/migrate mannshaft

tomaszdurka:
- Add migrate script
- Add lib for migrate, improve migrate
- Use htmlToSlate, try to migrate richtext properly


Itrulia (commented):
Wouldn't it be easier to use prisma instead of graphql to write data into the DB?

libs/api/src/lib/memberContext.ts
```diff
@@ -694,14 +700,66 @@ export class MemberContext implements MemberContextInterface {
       throw new InternalError()
     }
 
-    const invoice = await this.renewSubscriptionForUser({subscription})
+    if (startsAt < now || paidUntil) {
```
michael-scheurer, line 703: @elias-summermatter could this if/else statement break the special subscriptions at Hauptstadt (subscriptions already paid at the end of this year, but valid from March 2025 on) or the normal subscriptions at Tsüri or elsewhere?

How about adding one or two extra tests? Could it be worth it?

libs/migrate-wordpress/.env
```diff
@@ -0,0 +1,3 @@
+WORDPRESS_URL=https://mannschaft.com
+WEPUBLISH_API_URL=http://localhost:4000
+TOKEN=DDqxIRefm7VxG3ue1czEPbnW66U6FcxJ
```
elias-summermatter, line 3: @tomaszdurka  is this not a secret that you leak?

libs/api/src/lib/memberContext.ts
```diff
@@ -641,12 +641,16 @@ export class MemberContext implements MemberContextInterface {
     properties: Pick<MetadataProperty, 'key' | 'value' | 'public'>[],
     autoRenew: boolean,
     extendable: boolean,
-    startsAt?: Date | string
+    startsAt: Date | string = new Date(),
```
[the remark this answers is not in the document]
elias-summermatter: Its just for new subscriptions not for old one the old ones are handled in the subscription lib. So its not that dangerous

libs/api/src/lib/memberContext.ts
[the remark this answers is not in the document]
elias-summermatter: @michael-scheurer this just impactes new subscriptions. @tomaszdurka I guess you have been carefully testing the change already that everything works fine also with renewal of new created subscripitons?

elias-summermatter: Since it just impactes new subscriptions i would say no

libs/api/src/lib/memberContext.ts
```diff
@@ -694,14 +700,66 @@ export class MemberContext implements MemberContextInterface {
...
 
-    // Send subscribe mail
-    await this.sendMailForSubscriptionEvent(SubscriptionEvent.SUBSCRIBE, subscription, {})
+      const user = await this.prisma.user.findUnique({
+        where: {
+          id: subscription.userID
+        },
+        select: unselectPassword
+      })
+      const invoice = await this.prisma.invoice.create({
```
elias-summermatter, line 712: @tomaszdurka where do you create the periods? And are here not already existing functions for this? Code duplication?

libs/api/src/lib/server.ts
```diff
@@ -158,6 +158,8 @@ export class WepublishServer {
       allowBatchedHttpRequests: true
     })
 
+    await fs.promises.writeFile('./apps/api-example/schema-v1.graphql', printSchema(publicSchema))
```
elias-summermatter, line 161: What about this change? Why is that in this pull request?

libs/api/src/lib/memberContext.ts
elias-summermatter, line 703: @tomaszdurka do you checked that only admins can pass different startAt and paidUntil?

libs/api/src/lib/memberContext.ts
```diff
@@ -663,14 +667,16 @@ export class MemberContext implements MemberContextInterface {
...
+
     const subscription = await prisma.subscription.create({
       data: {
         userID,
-        startsAt: startsAt ? startsAt : new Date(),
+        startsAt,
         modifiedAt: new Date(),
         paymentMethodID: paymentMethodId,
         paymentPeriodicity,
-        paidUntil: null,
```
elias-summermatter, line 673: Why can't that be null? For example if someone buys a subscription then the paidUntil is null at the beginning before it is paid or did you change something there?

libs/api/src/lib/memberContext.ts
```diff
@@ -663,14 +667,16 @@ export class MemberContext implements MemberContextInterface {
       )
     }
 
+    const now = new Date()
+
     const subscription = await prisma.subscription.create({
       data: {
         userID,
-        startsAt: startsAt ? startsAt : new Date(),
+        startsAt,
```
elias-summermatter, line 675: Here same what happens if startsAt is null?

libs/api/src/lib/memberContext.ts
```diff
@@ -694,14 +700,66 @@ export class MemberContext implements MemberContextInterface {
...
 
-    return {
-      subscription,
-      invoice
+      await prisma.subscription.update({
+        where: {id: subscription.id},
+        data: {
+          periods: {
+            create: {
+              startsAt,
```
elias-summermatter, line 740: @tomaszdurka did you ensure that the period you import is always just one times the period (month,year eg.)otherwise you will get a period that is for example 3 years is that intended?

michael-scheurer (approved):

elias-summermatter (changes requested):
Open questions

tomaszdurka:
> Wouldn't it be easier to use prisma instead of graphql to write data into the DB?

I actually thought it will solve some potential issues when using DB directly like article structure, drafts, slate content, blocks. I also don't think it added effort in the end.
Plus I can run it remotely from other machine :)

PR #1465, opened 2024-08-08
enh/tag-filter

michael-scheurer:


michael-scheurer:
Could implement it via slugs later. Have to implement proper polymorphism.

PR #1477, opened 2024-08-14
f/add-cf-turnstile-captcha

elias-summermatter:


libs/api/src/lib/challenges/cfTurnstile.ts
```diff
@@ -0,0 +1,47 @@
...
+  ChallengeValidationProps,
+  ChallengeValidationReturn
+} from './challengeProvider'
+
+export class CfTurnstile implements ChallengeProvider {
+  constructor(private turnstile_secret_key: string) {}
+
+  async generateChallenge(testing = false): Promise<Challenge> {
+    return {
+      challengeID: 'not supported is turnstile!',
```
Itrulia, line 13: Should probably not be set.

Itrulia (changes requested):
Could we add a type or something to the challenge so the frontend knows what to do with it? I understand that in the future the algebraic one will be gone as it's essentially useless, but we could have different ones.

libs/api/src/lib/challenges/cfTurnstile.ts
elias-summermatter: @Itrulia  What do you mean by "should not be set"?

elias-summermatter:
@Itrulia  Where would you return the type? Because for turnstle you don't need to do any request to the api you can just get the solution from  cf....

libs/api/src/lib/challenges/cfTurnstile.ts
Itrulia: The challengeID should not be set, if it's not supported.

elias-summermatter: Is this possible on graphql level if it is required for algebraic but not for turnstile captcha?

Itrulia: Yes, by using Unions (like Blocks). Which would then also require the type like I wrote in another comment. The frontend needs the type to know how to display it.

elias-summermatter:
@Itrulia added a type could you maybe code the fe part that we can finish that off and make the media happy :-)

Itrulia:
roger, will work on it today or tomrrow!

PR #1479, opened 2024-08-17
feat: allow specific articles to disable comments

Itrulia:


apps/babanews/pages/a/[slug].tsx
```diff
@@ -66,10 +66,12 @@ export default function ArticleBySlugIdOrToken() {
             />
           </ArticleWrapper>
 
-          <ArticleWrapper>
-            <H3 component={'h2'}>Kommentare</H3>
-            <CommentListContainer id={data.article.id} type={ApiV1.CommentItemType.Article} />
-          </ArticleWrapper>
+          {!data.article.disableComments && (
```
Comment (michael-scheurer) [inline apps/babanews/pages/a/[slug].tsx:69]: I would prefer having a dedicated component for the comments living in the website builder. This component would take care if the comments are hidden or shown.
Now this is sort of duplicated code. We have to add all over the apps the same line of code. With a bunch of apps that's doable, but not when it comes to 10, 20, 30. That worries me.
What do you think? Is this doable?

michael-scheurer (changes requested):

PR #1480, opened 2024-08-17
feat(mannschaft): added sharing to article date

Itrulia:


michael-scheurer (changes requested):
Doesn't work with Chrome, Firefox nor Brave on my machine. Problem is the share function of navigator.

michael-scheurer (approved):

PR #1481, opened 2024-08-18
feat: added birthday to the user object

Itrulia:
Includes API, Editor & Website Builder implementation

Additional work that has been done:

1. Reworked the personal data form to use the user-form just like the registration form and subscribe form.
2. Added repeat password to the user-form, this now means the registration and subscribe form support it now too.
3. Improved validation handling of repeat password to fire even when not every other validation passes.


build/postInstall.js
Itrulia, line 1: Makes the post install a lot faster, these are already done by the normal `npm install` and are thus duplicated.

apps/babanews/pages/event/index.tsx
Itrulia, line 1: Now done in the website provider

michael-scheurer (changes requested):
Here is my review. Thanks for the work. Looks great.

- I think, it would be good, to also delete the birthdate in the editor. After saving, it seems as it was deleted. But if you re-load the page, there it is again.
image
- The input field of the user form should be in a format, most users of we.publish are used to. Best would be dd.mm.yyyy imo. Other formats could lead to confusion and errors. Perfect if we could align that also in the editor, so that it is consistent for a user.

PR #1483, opened 2024-08-19
feat(mannschaft): added a paywall to Mannschaft+ articles

Itrulia:


michael-scheurer (approved):
Perfect work. I didn't find anything to improve ;)

Next time working on paywall we could move things into editor and start adding settings there. For example I would like to have a boolean on article level to decide whether it is behind paywall or not instead of tags. Would reduce our overuse (abuse) of tags. Also we could make the paywall block customizable etc.

PR #1487, opened 2024-08-21
feat: added country select to website, editor & API with validation

Itrulia:


libs/membership/website/src/lib/formatters/format-currency.spec.ts
Itrulia, line 1: Fixes broken tests from master. The locale is supposed to be de-CH not ch-DE

Itrulia:
Already handled questions via DM:

1. As per UAC Hauptstadt will have to send the correct data. Else we remove the validation on the API side. Please confirm @michael-scheurer 
2. Yes, as it's not shown it's not mandatory in the UI.
3. Merge error, was reverted.

PR #1494, opened 2024-08-26
feat(editor): Manage poll votes

tomaszdurka:
- Add PollVote resolver/service
- Genearte api clients
- Introduce pagination-state-container and pagination helper component
- Add poll answer filter, allow to keep static filter data inside
- Add PollVoteList listing


tomaszdurka:
@Itrulia take a look, please.

I've added some generics as I couldn't stand the way it was repeated for all these things and abused the `useEffect()`.
It's far from perfect as I made quite some compromises - would love to get some feedback.

libs/ui/editor/src/lib/listView/use-selectable-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+type UseSelectableListProps = {
+  ids?: string[]
+}
+
+export function useSelectableList({ids}: UseSelectableListProps) {
+  const [selectedItems, setSelectedItems] = useState<string[]>([])
+  const allSelected = ids && selectedItems.length === ids.length
+  const someSelected = ids && selectedItems.length > 0
+
+  const toggleAll = () => setSelectedItems(someSelected ? [] : [...(ids ?? [])])
```
Itrulia, line 12:
```suggestion
  const toggleAll = () => setSelectedItems(someSelected ? [] : ids ?? []])
```

Should do the trick, no?

libs/ui/editor/src/lib/listView/use-selectable-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const selectItem = (id: string) => {
+    setSelectedItems([...new Set([...selectedItems, id])])
+  }
+  const deselectItem = (id: string) => {
```
Itrulia, line 17:
```suggestion
  const unselectItem = (id: string) => {
```

libs/ui/editor/src/lib/listView/use-selectable-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+    setSelectedItems(selectedItems.filter(item => item !== id))
+  }
+  const toggleItem = (id: string | any, value?: boolean) => {
+    if (value === undefined) {
+      value = !selectedItems.includes(id)
+    }
+    if (value) {
+      selectItem(id)
+    } else {
+      deselectItem(id)
```
Itrulia, line 27:
```suggestion
      unselectItem(id)
```

libs/ui/editor/src/lib/listView/use-selectable-list.tsx
tomaszdurka: Yeah, it should work. Was worried about reference, but it works fine. I will do my own change as there is 1 extra bracket in change 😎

PR #1495, opened 2024-08-26
fix/TSRI-79-remove-white-space-from-bildwurf

michael-scheurer:


Itrulia (changes requested):
Should be done in the bildwurf block and using emotion not adding a CSS file

LukasSkywalker (approved):

Itrulia (approved):

PR #1503, opened 2024-09-04
test(events-import-api): fix flaky tests

Itrulia:


libs/event/import/api/src/lib/import/kulturzueri-parser.ts
Itrulia, line 1: Was 1:1 the same as kulturagenda-parser.ts.

PR #1506, opened 2024-09-08
Allow introspection on v2

tomaszdurka:


apps/api-example/src/nestapp/app.module.ts
```diff
@@ -57,7 +57,8 @@ import {PollModule} from '@wepublish/poll/api'
       path: 'v2',
       cache: 'bounded',
       playground: process.env.NODE_ENV === 'development',
-      allowBatchedHttpRequests: true
+      allowBatchedHttpRequests: true,
+      introspection: true
```
elias-summermatter, line 61: Use the setting from the config file. Otherwise it looks good

elias-summermatter (changes requested):

apps/api-example/src/nestapp/app.module.ts
tomaszdurka: Ah this is about allowBatchedHttpRequests not the introspection i think it's fine with batch. 
Plus this is private API, the decision making and config if at all should be applied on gateway.

PR #1510, opened 2024-09-09
feat(tsri, *): delete preferred name as a feature and hide flair for tsri

Itrulia:


apps/tsri/pages/profile/index.tsx
```diff
@@ -43,7 +43,10 @@ function Profile() {
         </SubscriptionLink>
       </ProfileHeading>
 
-      <PersonalDataFormContainer mediaEmail="info@tsri.ch" />
+      <PersonalDataFormContainer
+        mediaEmail="info@tsri.ch"
+        fields={['firstName', 'flair', 'address', 'password', 'image']}
```
Itrulia, line 48:
```suggestion
        fields={['firstName', 'address', 'password', 'image']}
```

libs/api/prisma/migrations/20240909161100_delete_preferred_name/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `preferredName` on the `users` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+ALTER TABLE "users" DROP COLUMN "preferredName";
```
michael-scheurer, line 8: I will double-check with tsri.

apps/tsri/pages/profile/index.tsx
```diff
@@ -43,7 +43,10 @@ function Profile() {
         </SubscriptionLink>
       </ProfileHeading>
 
-      <PersonalDataFormContainer mediaEmail="info@tsri.ch" />
+      <PersonalDataFormContainer
+        mediaEmail="info@tsri.ch"
+        fields={['firstName', 'address', 'password', 'image']}
```
michael-scheurer, line 48: Missing _name, zip, city, country, email_ compared to now

image

michael-scheurer (changes requested):

libs/api/prisma/migrations/20240909161100_delete_preferred_name/migration.sql
Itrulia: This story was about Tsri not wanting this

apps/tsri/pages/profile/index.tsx
Itrulia: image
image

did you check the wrong forms?

michael-scheurer (approved):

PR #1519, opened 2024-09-17
feat: implemented hot & trending, google analytics

Itrulia:
1. Datasource for hot&trending is configureable
2. Only GA has been implemented for now
3. Can be accessed via API or via TeaserListBlock


apps/mannschaft/pages/a/[slug].tsx
Comment (Itrulia) [inline apps/mannschaft/pages/a/[slug].tsx:1]: Reverted tag in url since I had to touch this file anyway

apps/mannschaft/pages/a/[slug].tsx
```diff
@@ -18,8 +18,8 @@ import {useHasSubscription} from '../../src/paywall/has-subscription'
 import {PaywallBlock} from '../../src/paywall/paywall-block'
 
 const paywallCss = css`
-  // Shows the first 3 blocks and hides the rest
-  & > :nth-child(n + 4):not(:is(${ArticleInfoWrapper})) {
+  // Shows the first 4 blocks (usually title, image, richtext + ad) and hides the rest
```
Comment (Itrulia) [inline apps/mannschaft/pages/a/[slug].tsx:21]: Updated this due to the ad block being the 3rd block now, so no text as teaser was shown.

apps/mannschaft/pages/index.tsx
```diff
@@ -26,6 +26,12 @@ export const getStaticProps: GetStaticProps = async () => {
     }),
     client.query({
       query: ApiV1.PeerProfileDocument
+    }),
+    client.query({
```
Itrulia, line 30: Added for SSR :)

libs/api/src/lib/context.ts
Itrulia, line 1: Added it to context so I have access to it in the teaser list block

michael-scheurer (changes requested):
I really like that you integrated GA in our api. That is very nice an re-usable. Thank you.
I left two comments.
And one more question: would you mind to provide a screenshot how it looks on mobile and desktop for mannschaft?

libs/api/src/lib/graphql/blocks.ts
```diff
@@ -582,23 +590,33 @@ export const GraphQLTeaserListBlock = new GraphQLObjectType<TeaserListBlock, Con
...
+                    },
+                    ArticleSort.PublishedAt,
+                    SortOrder.Descending,
+                    undefined,
+                    skip,
+                    take,
+                    prisma.article
+                  )
+                : {
+                    nodes: await hotAndTrendingDataSource.getMostViewedArticles({skip, take})
```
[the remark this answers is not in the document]
Itrulia: The existing implementation already sorts by publishedAt, this shouldnt change anything

Itrulia:
The UI already exists, you can see it on tsri or on mannschaft. I just filled it with data.

Good catch with the false value, should probably prevent a bug.


If no GA creds are set it will return an empty array and log a warning. Implemented this when I locked myself out of a page due to errors :D

michael-scheurer (approved):

PR #1521, opened 2024-09-18
enh/share-35-extend-subscription

michael-scheurer:


LukasSkywalker (approved):

libs/membership/website/src/lib/subscription-list/subscription-list-item.tsx
```diff
@@ -217,6 +223,22 @@ export function SubscriptionListItem({
...
         </Paragraph>
       </MembershipModal>
+
+      {/* confirm extend */}
+      <MembershipModal
+        open={confirmExtend}
+        onCancel={() => setConfirmExtend(false)}
+        onSubmit={async () => {
+          setConfirmExtend(false)
+          await callAction(extend)()
```
tomaszdurka, line 233: Got me thinking we close modal before we submit. Meaning we assume it always work, right? It's probably general question regarding website-builder. Do we show errors anywhere in case of "connectivity issues etc"?

I'd say it's fine for now here still. Rather. my curiousity

tomaszdurka (approved):

PR #1522, opened 2024-09-18
fix/update invoices after extending subscription

michael-scheurer:
This PR is based on https://github.com/wepublish/wepublish/pull/1521, so please review and close 1521 first.

With this PR, the invoices and subscriptions are fetched again after a subscription is renewed. Also, the pay-btn has been removed from the subscription list item to avoid manually starting to pay an invoice that is actually paid automatically in the background. Also, the pay btn is logically related to an invoice and not a subscription.


michael-scheurer:
This PR is based on https://github.com/wepublish/wepublish/pull/1521, so please review and close 1521 first.

With this PR, the invoices and subscriptions are fetched again after a subscription is renewed. Also, the pay-btn has been removed from the subscription list item to avoid manually starting to pay an invoice that is actually paid automatically in the background. Also, the pay btn is logically related to an invoice and not a subscription.

Itrulia (changes requested):
Why was the ability to pay a subscription removed?

michael-scheurer:
> Why was the ability to pay a subscription removed?

@Itrulia long story. I should have noted it down in detail after coding. I only noted that here:

> Also, the pay-btn has been removed from the subscription list item to avoid manually starting to pay an invoice that is actually paid automatically in the background. Also, the pay btn is logically related to an invoice and not a subscription.

The only solution would have been to move the requests for open invoices and subscriptions up to a parent component and handle the update logic there. So I decided to solve it this way for now. At the moment it is sometimes an issue that users pay multiple times. I was able to reproduce this.

Itrulia (approved):

PR #1523, opened 2024-09-18
feat(website-builder, blocks): Remove the gap between two sibling richtext-blocks

tomaszdurka:


libs/block-content/website/src/lib/richtext/richtext-block.tsx
```diff
@@ -9,7 +9,11 @@ import {Editable, Slate, withReact} from 'slate-react'
 export const isRichTextBlock = (block: Block): block is RichTextBlockType =>
   block.__typename === 'RichTextBlock'
 
-export const RichTextBlockWrapper = styled('div')``
+export const RichTextBlockWrapper = styled('div')`
```
Itrulia, line 12: Probably don't want this generally but for Mannschaft only, no?

tomaszdurka: Depends. Do you think anyone using separate rich-text-blocks to make gaps between?
Probably the paywall could be a feature of its own and then around that we could cut content accordingly. 
Maybe let's talk on the meeting if we should apply that in Mannshaft only and how.

Itrulia: This piece of CSS has quite a side effects so I would rather not have that in for every media. 

If the paywall is implemented in the website builder, we have the possibility to properly think about supporting it in the editor, so we know where and when etc to add it. So this piece of CSS would not help for this.

PR #1526, opened 2024-09-19
fix(users): Alternatively do not allow null for address

tomaszdurka:


tomaszdurka:
@Itrulia please close if https://github.com/wepublish/wepublish/pull/1525 is with you.

PR #1537, opened 2024-09-25
refactor(mannschaft): few improvements, favicons, use gtm instead of ga, etc

Itrulia:


libs/authentication/website/src/lib/registration-form/user-form.tsx
```diff
@@ -1,5 +1,5 @@
 import {IconButton, InputAdornment, Theme, css, styled} from '@mui/material'
-import {DatePicker} from '@mui/x-date-pickers'
```
Itrulia, line 2: makes it always open the datepicker on click

michael-scheurer (approved):

PR #1559, opened 2024-10-01
fix/editor-currency-and-more

michael-scheurer:
In this PR you will find some bug fixes and minor improvements regarding users and subscriptions in the **editor.** These are:

- ask user before extending subscription early.
- show first and last name correctly in the subscription table list.
- show the correct currencies on member plans, subscriptions and invoices
- update the currency of a subscription, when changing the member plan of a subscription.


Itrulia (approved):
Looks good, CHF is probably Chf now in text, right?

PR #1569, opened 2024-10-14
feat(mannschaft,article-api): hot & trending just 24h

Itrulia:
Also improved handling of fetching articles from the google analytics api


codegen.yml
```diff
@@ -121,44 +121,4 @@ generates:
         Color: string
         Slug: string
         RichText: Node[]
-        Upload: File
```
Itrulia, line 124: Threw errors while generating because server is not running etc.

libs/google-analytics/api/src/lib/google-analytics.service.ts
Itrulia, line 1: Turned out the correct articles were not included due to database sorting, so had to slice on the viewMap

libs/google-analytics/api/src/lib/google-analytics.service.ts
```diff
@@ -71,16 +71,25 @@ export class GoogleAnalyticsService implements HotAndTrendingDataSource {
         return object
       }
 
+      // Don't include sub pages of the article prefix
+      if (slug?.split('/').length !== this.config.articlePrefix.split('/').length) {
```
Itrulia, line 75: hides urls like `/a/tag/news` in case they ever have more views than the articles themselves etc.

michael-scheurer (commented):
Looks good. Did you compare the actual hit list in GA with the result on the hot-and-trending teaser? Are the list entries the same?
If this is given, we can merge.

PR #1570, opened 2024-10-14
enh/BAJ-156-frage-des-tages-image

michael-scheurer:


apps/bajour/src/components/frage-des-tages/frage-des-tages.tsx
```diff
@@ -134,6 +135,13 @@ export const StyledComment = styled(Comment)`
...
 `
 
+export const FdtArticleImage = styled(Image)`
+  grid-column-start: 5;
+  grid-column-end: 9;
+  margin-bottom: ${({theme}) => theme.spacing(4)};
+  border-radius: ${({theme}) => theme.spacing(2.5)};
+`
+
 export const FDTLogo = styled(FrageDesTagesLogo)`
```
Itrulia, line 145: can be deleted now, right?

michael-scheurer: good catch. thx. fixed it.

Itrulia (approved):

PR #1577, opened 2024-10-21
feat(permissions): Require guard for each endpoint, introduce @Public+@Authenticated decorator/guard

tomaszdurka:
- fix/Add permissions to system-mail.resolver.ts
- feat/require all endpoints to have permission (GUEST or other)


libs/permissions/api/src/lib/permission.guard.ts
```diff
@@ -18,7 +18,7 @@ export class PermissionsGuard implements CanActivate {
     ])
 
     if (!permissions.length) {
-      return true
+      return false
```
Itrulia, line 21: Can revert this now I guess

Itrulia (approved):

libs/permissions/api/src/lib/permission.guard.ts
tomaszdurka: Hmm, I guess it doesnt make a difference. Basically passing empty `@Permissions()` will act as `@Public` if i return `true` here. I say that is not intended, wdyt?
I would expect if you provider `@Permissions` you wanna restrict, otherwise use `@Public` / `@Authenticated`.

libs/permissions/api/src/lib/authenticated.guard.ts
Itrulia, line 1: Reason we don't use the one we already have and have authentication stuff in authentication/api?

Itrulia (changes requested):

libs/settings/api/src/lib/settings.resolver.ts
```diff
@@ -27,22 +29,22 @@ export class SettingsResolver {
     return this.settingsService.settingByName(name)
   }
 
+  @Permissions(CanGetSettings)
```
Itrulia, line 32: Not needed

libs/authentication/api/src/lib/optional-authentication.guard.ts
```diff
@@ -1,16 +1,10 @@
...
 /**
  * This Guard is to attach the user to the request
  */
-export class OptionalAuthenticationGuard extends AuthGuard('session') {
-  public override getRequest(context: ExecutionContext) {
-    const ctx = GqlExecutionContext.create(context)
-    return ctx.getContext().req
-  }
-
+export class OptionalAuthenticationGuard extends AuthenticationGuard {
```
Itrulia, line 7: Don't think this should extend

libs/settings/api/src/lib/settings.resolver.ts
tomaszdurka: Was here, but can remove.

libs/authentication/api/src/lib/optional-authentication.guard.ts
Itrulia: Yeah but the logic around them still isn't the same, if `AuthenticationGuard` now gets some more authentication features, this one will also get them which shouldn't.

While they have some similar code, they are not the same thing logically.

Itrulia (approved):

PR #1578, opened 2024-10-21
enh/TSRI-98-migrate-to-other-payment-provider

michael-scheurer:
I decided to create a re-usable solution for the other media without the need of hard-coding in the front-end.

In the editor the media can set to migrate to a target payment method within a subscription plan (member plan).

This looks like this:
image

The migration hits as soon as the user calls a public subscription endpoint (e.g. extend a subscription or pay an open invoice etc.).


LukasSkywalker (approved):

libs/api/prisma/migrations/20241022171553_add_migrate_to_target_pm_to_member_plan/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "member.plans" ADD COLUMN     "migrateToTargetPMid" TEXT;
```
elias-summermatter, line 2: Is TEXT here correct shouln't this be VARCHAR? ^^

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -581,8 +589,11 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
         const subscription = (await prisma.subscription.findUnique({
           where: {
             id: subscriptionId
+          },
+          include: {
+            memberPlan: true
           }
-        })) as SubscriptionWithRelations
+        })) as SubscriptionWithRelations & {memberPlan: MemberPlan}
```
elias-summermatter, line 596: Why do we do this explicit casting should prisma not return the type save type?

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -968,10 +981,11 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
         return await createPaymentWithProvider({
           paymentMethodID: paymentMethod.id,
           invoice,
-          saveCustomer: false,
+          saveCustomer: true,
```
elias-summermatter, line 984: Why was this changed?

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -1014,9 +1032,10 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
         return await createPaymentWithProvider({
           paymentMethodID: invoice.subscription.paymentMethodID,
           invoice,
-          saveCustomer: false,
+          saveCustomer: true,
```
elias-summermatter, line 1035: Same here why?

apps/api-example/schema-v1-admin.graphql
```diff
@@ -1926,6 +1927,7 @@ input MemberPlanInput {
   extendable: Boolean!
   maxCount: Int
   availablePaymentMethods: [AvailablePaymentMethodInput!]!
+  migrateToTargetPMid: ID
```
elias-summermatter, line 1930: Should it not be ID= migrateToTargetPMID as it is for all others?

libs/api/__tests__/api/private/index.ts
```diff
@@ -959,6 +959,7 @@ export type MemberPlan = {
   id: Scalars['ID'];
   image?: Maybe<Image>;
   maxCount?: Maybe<Scalars['Int']>;
+  migrateToTargetPMid?: Maybe<Scalars['ID']>;
```
elias-summermatter, line 962: Why is here just the ID field compared to  availablePaymentMethods where the PaymentMethode is passed?

libs/api/prisma/schema.prisma
```diff
@@ -509,6 +509,8 @@ model MemberPlan {
   currency                Currency
   amountPerMonthMin       Float
   availablePaymentMethods AvailablePaymentMethod[]
+  migrateToTargetPM       PaymentMethod?           @relation("PaymentMethodMigration", fields: [migrateToTargetPMid], references: [id])
+  migrateToTargetPMid     String?
```
elias-summermatter, line 513: Keep Naming consistent (ID)

libs/api/src/lib/context.ts
```diff
@@ -225,6 +225,7 @@ export interface CreatePaymentWithProvider {
   successURL?: string
   failureURL?: string
   user?: User
+  migrateToTargetPMid?: string
```
elias-summermatter, line 228: id=>ID

libs/api/src/lib/context.ts
```diff
@@ -1003,9 +1004,12 @@ export async function contextFromRequest(
       saveCustomer,
       failureURL,
       successURL,
-      user
+      user,
+      migrateToTargetPMid
```
elias-summermatter, line 1008: Why not accept here the subscription with memberpaln insted of some id? Like this we can determine the correct payment method from the subscription.

libs/api/src/lib/context.ts
```diff
@@ -1014,6 +1018,26 @@ export async function contextFromRequest(
...
+        const updatedSubscription = await prisma.subscription.update({
+          data: {
+            paymentMethodID: migrateToTargetPMid
+          },
+          where: {
+            id: invoice.subscriptionID
+          }
+        })
+
+        if (!updatedSubscription) {
```
elias-summermatter, line 1036: Since the update methode of prisma throws if update does not work that is not nesseary.

libs/api/src/lib/context.ts
```diff
@@ -1014,6 +1018,26 @@ export async function contextFromRequest(
...
+       * Mainly used in mutation.public.ts
+       * Requirements written down here https://wepublish.atlassian.net/browse/TSRI-98
+       */
+      if (migrateToTargetPMid) {
+        const updatedSubscription = await prisma.subscription.update({
+          data: {
+            paymentMethodID: migrateToTargetPMid
+          },
+          where: {
+            id: invoice.subscriptionID
```
elias-summermatter, line 1032: If you get the subscription passed you can use here subscription.id

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -421,7 +427,8 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
             saveCustomer: true,
             paymentMethodID: paymentMethod.id,
             successURL,
-            failureURL
+            failureURL,
+            migrateToTargetPMid: memberPlan.migrateToTargetPMid
```
elias-summermatter, line 431: Why is this needed in the registerMemberAndReceivePayment function should here not be the payment methode already correct since its a new subscription?

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -693,7 +704,8 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
           saveCustomer: true,
           paymentMethodID: subscription.paymentMethodID,
           successURL,
-          failureURL
+          failureURL,
+          migrateToTargetPMid: subscription.memberPlan.migrateToTargetPMid
```
elias-summermatter, line 708: In the  extendSubscription it is correct.

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -968,10 +981,11 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
         return await createPaymentWithProvider({
           paymentMethodID: paymentMethod.id,
           invoice,
-          saveCustomer: false,
+          saveCustomer: true,
           successURL,
           failureURL,
-          user
+          user,
+          migrateToTargetPMid: subscription.memberPlan.migrateToTargetPMid
```
elias-summermatter, line 988: createPaymentFromInvoice its fine too

libs/api/src/lib/graphql/mutation.public.ts
```diff
@@ -1014,9 +1032,10 @@ export const GraphQLPublicMutation = new GraphQLObjectType<undefined, Context>({
         return await createPaymentWithProvider({
           paymentMethodID: invoice.subscription.paymentMethodID,
           invoice,
-          saveCustomer: false,
+          saveCustomer: true,
           successURL,
-          failureURL
+          failureURL,
+          migrateToTargetPMid: invoice.subscription.memberPlan.migrateToTargetPMid
```
elias-summermatter, line 1038: createPaymentFromSubscription also required

elias-summermatter (changes requested):

libs/api/src/lib/graphql/mutation.public.ts
michael-scheurer: So that the new credit card details would be stored

libs/api/src/lib/graphql/mutation.public.ts
michael-scheurer: same answer as above ;9

libs/api/prisma/migrations/20241022171553_add_migrate_to_target_pm_to_member_plan/migration.sql
michael-scheurer: That's how prisma converts it. This is consistent to the rest of the schema. See schema.prisma

libs/api/src/lib/graphql/mutation.public.ts
michael-scheurer: Typescript forced me to add this line.

apps/api-example/schema-v1-admin.graphql
michael-scheurer: changed it.

libs/api/__tests__/api/private/index.ts
michael-scheurer: the migration property lives on the member plan directly and not on a n:m available payment method relations. that is how available payment methods should also be related some day.

libs/api/src/lib/context.ts
michael-scheurer: done

libs/api/src/lib/context.ts
michael-scheurer: The idea is, that we can tell this method, whether to migrate or not. Would like to keep it that way.

libs/api/src/lib/context.ts
michael-scheurer: idem

libs/api/src/lib/context.ts
michael-scheurer: :heavy_check_mark: fixed

libs/api/src/lib/graphql/mutation.public.ts
michael-scheurer: one could argue so. probably the expected behavior is more concise. 

:heavy_check_mark: done

elias-summermatter (approved):

PR #1597, opened 2024-10-29
feat(navigation-website): add native way to add icons to navigation

Itrulia:


libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -259,7 +262,7 @@ export function Navbar({
         <NavbarInnerWrapper>
           <NavbarMain>
             <NavbarIconButtonWrapper>
-              <IconButton size="large" aria-label="Menu" onClick={toggleMenu}>
+              <IconButton size="large" aria-label="Menu" onClick={toggleMenu} color={'inherit'}>
```
Itrulia, line 265: @tomaszdurka fyi had to be re-added. Caused a lot of websites to be different.

It now inherits the color so you can change `NavbarIconButtonWrapper` in cültür to change the icon color.

michael-scheurer (approved):

PR #1610, opened 2024-10-29
BA-33: Bajour Header

LukasSkywalker:


libs/api/src/lib/pagination.args.ts
```diff
@@ -0,0 +1,10 @@
+import {ArgsType, Field, Int} from '@nestjs/graphql'
+
+@ArgsType()
+export class PaginationArgs {
+  @Field(type => Int)
+  skip = 0
```
Itrulia, line 6: Oh does this work? I usually use {defaultValue: 0} in the @Field, because I think it's now required to include in the GraphqlSchema

elias-summermatter (approved):

libs/banner/api/setup-database.js
Itrulia, line 1: Don't think we want to setup the database for tests anymore and generally mock the database away. I see your tests also use a prisma mock so don't think this is needed

libs/api/prisma/schema.prisma
Itrulia, line 1: Missing `@@map("...")` for the table names, so the tables are now different than the others (also effects the migrations)

PR #1614, opened 2024-10-30
feat(bajour): article charts implemented

Itrulia:


libs/website/api-v2/src/lib/graphql.ts
Itrulia, line 1: Merge error, should not exist anymore

michael-scheurer (approved):

PR #1624, opened 2024-11-04
Mollie payment provider integration

elias-summermatter:


libs/api/src/lib/memberContext.ts
```diff
@@ -273,10 +273,14 @@ export class MemberContext implements MemberContextInterface {
...
 
+      const memberplan = await this.prisma.memberPlan.findUnique({
+        where: {id: subscription.memberPlanID}
+      })
+
       const newInvoice = await this.prisma.invoice.create({
         data: {
           subscriptionID: subscription.id,
-          description: `Membership from ${startDate.toISOString()} for ${user.name || user.email}`,
+          description: `${memberplan.name}: ${format(startDate, 'dd-MM-yyyy')}`,
```
LukasSkywalker, line 283: I see that the user's email or name are no longer present here. Aren't they useful for the financial accounting at the media? Or do they see them anyway?

libs/payment/api/src/lib/payment-provider/mollie-payment-provider.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+}
+
+function calculateAndFormatAmount(invoice: InvoiceWithItems) {
+  return `${(
+    invoice.items.reduce(
+      (prevItem, currentItem) => prevItem + currentItem.amount * currentItem.quantity,
+      0
+    ) / 100
+  ).toFixed(2)}`
```
LukasSkywalker, line 66: \`${}\` should not be needed here, because toFixed already returns a String

libs/payment/api/src/lib/payment-provider/mollie-payment-provider.ts
```diff
@@ -0,0 +1,270 @@
...
+  constructor(props: MolliePaymentProviderProps) {
+    super(props)
+    this.webhookEndpointSecret = props.webhookEndpointSecret
+    this.apiBaseUrl = props.apiBaseUrl
+    this.mollieClient = createMollieClient({apiKey: props.apiKey})
+    this.methods = this.getPaymentMethode(props.methods)
+  }
+
+  getPaymentMethode(methods: string[] | undefined): MaybeArray<PaymentMethod> | undefined {
+    if (!methods) return undefined
```
LukasSkywalker, line 84: I suggest to always add curly brackets, even for single line blocks, so this does not happen: https://www.codecentric.de/wissens-hub/blog/curly-braces

There are also instances in L120 and L208

libs/payment/api/src/lib/payment-provider/mollie-payment-provider.ts
elias-summermatter: True :-)

LukasSkywalker (approved):

PR #1660, opened 2024-11-20
fix(url-adapter): Make sure tag is uri-encoded correctly

tomaszdurka:


libs/api/__tests__/utility.ts
```diff
@@ -154,6 +154,7 @@ export async function createGraphQLTestClient(overwriteRequest?: any): Promise<T
   function mockInstance<Type = unknown>(implementation?: PartialDeep<Type>) {
     return new (jest.fn().mockImplementation(() => implementation) as Mock<Type>)() as Type
   }
+
```
LukasSkywalker, line 157: Nice fix!

LukasSkywalker (approved):

PR #1673, opened 2024-11-21
refactor(website): re-use mitmachen page

Itrulia:
Some getInitialProps was kept due to them loading extra data.
We need a better way to reuse those than we currently have but it's out of scope for this.


libs/utils/website/src/lib/pages/subscribe-page.tsx
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        return preselectedMemberPlan && additionalMemberPlans !== 'all'
+          ? [preselectedMemberPlan]
+          : parentFiltered
+      }}
+      deactivateSubscriptionId={
+        props.deactivateSubscriptionId ?? (deactivateSubscriptionId as string | undefined)
+      }
+      failureURL=""
```
michael-scheurer, line 57: This connection got lost, but is important. Could we solve it by merging first the SHARE-42 story?

michael-scheurer (commented):
That's the kind of PRs I like the most ;) Getting rid of repetitive code. Great job! Found one potential issue.

libs/utils/website/src/lib/pages/subscribe-page.tsx
Itrulia: SHARE-43? That's the point why this is empty, as SHARE-43 will be merged first.

elias-summermatter:
@Itrulia can we merge this now?

Itrulia:
@elias-summermatter yeah just had to resolve merge conflicts

PR #1674, opened 2024-11-21
feat(website): combine and reuse profile pages

Itrulia:


apps/babanews/pages/_app.tsx
```diff
@@ -120,7 +120,6 @@ function CustomApp({Component, pageProps, emotionCache}: CustomAppProps) {
                 slug="main"
                 headerSlug="header"
                 iconSlug="icons"
-                subscriptionsUrl={null}
```
michael-scheurer, line 123: This has been removed everywhere but not replaced. What was this for?

libs/utils/website/src/lib/pages/profile/profile.tsx
```diff
@@ -47,48 +55,55 @@ function Subscriptions() {
...
+            failureURL=""
+            successURL=""
+          />
+        </SubscriptionListWrapper>
+      </SubscriptionsWrapper>
+
+      <ProfileWrapper>
+        <H4 component={'h1'}>Profil</H4>
+
+        <PersonalDataFormContainer mediaEmail="abo@gruppetto-magazin.ch" />
```
michael-scheurer, line 97: looks to me as if every media would get the gruppetto info mail now.

libs/utils/website/src/lib/pages/profile/profile.tsx
```diff
@@ -47,48 +55,55 @@ function Subscriptions() {
...
+        <SubscriptionListWrapper>
+          <H4 component={'h1'}>Offene Rechnungen</H4>
+
+          <InvoiceListContainer
+            filter={invoices =>
+              invoices.filter(
+                invoice => invoice.subscription && !invoice.canceledAt && !invoice.paidAt
+              )
+            }
+            failureURL=""
```
michael-scheurer, line 88: loosing props connection here.

libs/utils/website/src/lib/pages/profile/profile.tsx
```diff
@@ -47,48 +55,55 @@ function Subscriptions() {
...
+    <>
+      <SubscriptionsWrapper>
+        <SubscriptionListWrapper>
+          <H4 component={'h1'}>Aktive Abos</H4>
+
+          <SubscriptionListContainer
+            filter={subscriptions =>
+              subscriptions.filter(subscription => !subscription.deactivation)
+            }
+            failureURL=""
```
michael-scheurer, line 68: loosing props connection here.

apps/mannschaft/pages/profile/index.tsx
```diff
@@ -1,84 +1,3 @@
...
-    elements: {H4}
-  } = useWebsiteBuilder()
-
-  return (
-    <ProfileWrapper>
-      <H4 component={'h1'}>Profil</H4>
-
-      <PersonalDataFormContainer
-        mediaEmail="abo@mannschaft.com"
-        fields={['firstName', 'address', 'birthday', 'image', 'password', 'flair']}
```
michael-scheurer, line 29: This setting is being lost imo.

michael-scheurer (commented):
Again the sort of PRs I most like the most ;) A lot of lines of code gone.

I would like a lot, if we could align the profile form bellow the subscription on the left side. I was even thinking about using tabs (https://mui.com/material-ui/react-tabs/) for subscriptions and profile form. probably out of scope.

image

libs/utils/website/src/lib/pages/profile/profile.tsx
Itrulia: Because they are getting removed

libs/utils/website/src/lib/pages/profile/profile.tsx
Itrulia: Good catch

apps/babanews/pages/_app.tsx
Itrulia: Simply doesn't exist anymore

michael-scheurer (approved):
@Itrulia if you tested yourself and working, let's merge it and hand it over to media to test it.

PR #1681, opened 2024-11-25
feat(website): improve profile and navbar for members

Itrulia:
1. Also fixed issue with re-using apollo client
2. Also changed Gruppetto /abo to /mitmachen
3. Also changed Gruppetto to use the emotion cache


michael-scheurer (changes requested):
Heavily improved, and well done. Here are my findings:
- Given, that I am not logged-in and would like to become a member. If the payment provider throws an error (in this case for testing purposes with wrong credentials), an account as well as a subscription are created. There are two problems though:
a) User doesn't see, that he/she has actually a valid login. Solution imo: log-in user and redirect him/her to the set-up failure page.
b) User will intuitively click again on the subscribe button (probably several times) and each time it creates a new unpaid subscription => Update user state and subscriptions. Also update with "open invoice" within the menu
image

image

image

- Open invoice should come with a colored frame and the same exclamation icon as the menu button. See miro sketches.
image

image

- Found another bug. Given the following settings:
a) Create a stripe payment provider within the editor using standard dev environment.
image
b) create a subscription with this stripe pp
image
c) Create that subscription. You are logged-in but without any error but user stays on the subscription form and will likely hit the btn again and again and pile up his/her subscriptions.
image

- "Member werden"-Button and "Offene Rechnung"-Button should not appear at the same time.
image

- Check mobile. Imo logout btn could or should be placed elsewhere but not in the same line with the other btns.
image

More to come ;)

michael-scheurer:
- Problem with "Offene Rechnung" on mobile
image
=> As a solution I would probably shorten it to "Rechnung"
image

- Also the invoices and subscriptions on mobile are too much spaced around. Also difficulties in reading the text.
image

michael-scheurer:
- If I have a single active subscription on mobile, looks wired.
image

- Inconsistency in naming. It should be "Mein Konto". Instead with a single active subscription it is "Mein Account".
image

libs/api/src/lib/graphql/subscription-public.ts
```diff
@@ -51,6 +52,39 @@ export const GraphQLPublicSubscription = new GraphQLObjectType<SubscriptionWithR
...
+
+        /**
+         * Can only extend when:
+         *   Subscription is extendable
+         *   Subscription is not deactivated
+         *   Subscription is about to run out (less than 1 month)
+         *   All invoices have been paid (or cancelled)
+         *   Not using a deprecated payment method
+         */
+        return (
```
michael-scheurer, line 79: We could make this logic re-usable and refactor it in a separate function and use it among the code.

libs/membership/website/src/lib/has-active-subscription.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const {hasUser} = useUser()
+  const {data} = useSubscriptionsQuery({
+    fetchPolicy: 'cache-first',
+    skip: !hasUser
+  })
+
+  return useMemo(
+    () =>
+      !!data?.subscriptions.find(
+        subscription =>
```
michael-scheurer, line 15: This logic is implemented manually all over the code. We could re-use it in a dedicated function.

libs/membership/website/src/lib/has-unpaid-invoices.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const useHasUnpaidInvoices = () => {
+  const {hasUser} = useUser()
+  const {data} = useInvoicesQuery({
+    fetchPolicy: 'cache-first',
+    skip: !hasUser
+  })
+
+  return useMemo(
+    () => data?.invoices.some(invoice => !invoice.canceledAt && !invoice.paidAt),
```
michael-scheurer, line 13: same logic used all over the code. we could start sharing a re-usable function for that ;)

libs/membership/website/src/lib/subscription-list/__snapshots__/subscription-list-container.spec.tsx.snap
```diff
@@ -912,7 +912,13 @@ exports[`SubscriptionList Container should render WithFilter 1`] = `
         class="emotion-4 emotion-5"
       >
         <strong>
-          Kein aktives Abo
+          Kein has noch kein aktives Abo.
```
michael-scheurer, line 915: Typo. Should be "Du hast noch kein aktives Abo."

libs/membership/website/src/lib/subscription-list/__snapshots__/subscription-list.spec.tsx.snap
```diff
@@ -830,7 +830,13 @@ exports[`SubscriptionList should render Empty 1`] = `
         class="emotion-4 emotion-5"
       >
         <strong>
-          Kein aktives Abo
+          Kein has noch kein aktives Abo.
```
michael-scheurer, line 833: dito

libs/membership/website/src/lib/subscription-list/subscription-list.tsx
```diff
@@ -1,44 +1,35 @@
...
   } = useWebsiteBuilder()
 
   return (
     <SubscriptionListWrapper className={className}>
       {!loading && !error && !data?.subscriptions?.length && (
         <SubscriptionListItemWrapper>
           <SubscriptionListItemContent>
-            <strong>Kein aktives Abo</strong>
+            <strong>
+              Kein has noch kein aktives Abo. <Link href={subscribeUrl}>Jetzt ein Abo lösen.</Link>
```
michael-scheurer, line 31: dito with typo

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -476,42 +501,61 @@ const NavPaper = ({
...
+          {hasUnpaidInvoices && profileUrl && (
             <Button
               LinkComponent={Link}
-              href={loginUrl}
+              href={profileUrl}
+              variant="contained"
+              color="warning"
+              onClick={closeMenu}
+              startIcon={<MdWarning />}>
+              Offene Rechnung
```
michael-scheurer, line 512:
```suggestion
              Rechnung
```

michael-scheurer (changes requested):

libs/api/src/lib/graphql/subscription-public.ts
```diff
@@ -51,6 +52,39 @@ export const GraphQLPublicSubscription = new GraphQLObjectType<SubscriptionWithR
...
+          subscription.extendable &&
+          !subscription.deactivation &&
+          +add(new Date(), {months: 1}) > +subscription.paidUntil &&
```
michael-scheurer, line 82: Logic doesn't work as expected. Either in front-end or here. Please test.

Tested 7th of January 2025 for the following screenshot. And also tested with paidUntil 5th of January and 5th of March. None of them works.
image

michael-scheurer (changes requested):

libs/api/src/lib/graphql/subscription-public.ts
Itrulia: And use it where?

libs/membership/website/src/lib/has-active-subscription.tsx
Itrulia: This is the dedicated function.

libs/membership/website/src/lib/has-unpaid-invoices.tsx
Itrulia: This is the re-usable function for that.

libs/api/src/lib/graphql/subscription-public.ts
Itrulia: The logic was actually correct, but the query above for the unpaid invoice wasn't

libs/membership/website/src/lib/has-unpaid-invoices.tsx
michael-scheurer: I can live with it. In my code I just try to make life easier / safer for the following. So, that my fellows don't need to care about the fact, that he has to check for canceledAt and paidAt. I would probably just go with something like this:

`export function isActiveInvoice (invoice: Invoice): boolean => !invoice.canceledAt && !invoice.paidAt`

This is maybe because I am coming from OOP, where you would have such a function within the Invoice class. Defined it once, used it everywhere with exact same logic.

PR #1688, opened 2024-12-02
feat(sharing): Reorder og:images

tomaszdurka:


Itrulia (changes requested):
This now breaks whatsapp

Itrulia (approved):

PR #1700, opened 2024-12-10
feat(metadata): Reorder og images

tomaszdurka:


Itrulia (approved):
I hate when services dont implement the spec correctly.

PR #1706, opened 2024-12-16
feat(website): add width/height

Itrulia:


libs/image/website/src/lib/image.tsx
Itrulia, line 1: Haven't tested how layout is affected yet everywhere. Used to be in the in the past and then removed because of some layout issues, the CSS has changed since then and we use the EmotionCache provider, so don't think the old issues are still a thing.

michael-scheurer: What is this change about?
I'll gladly review it, when you are ready and have tested your code :)

Itrulia: Performance

Itrulia:
Tested on: Tsri, Gruppetto, Mannschaft.
Ready for approval @LukasSkywalker @michael-scheurer  :D

PR #1717, opened 2024-12-20
Add full text index

LukasSkywalker:


LukasSkywalker:
See https://github.com/wepublish/wepublish/pull/1923

PR #1719, opened 2025-01-08
feat(block-content-website): add image source to image gallery

Itrulia:


libs/block-content/website/src/lib/image-gallery/image-gallery-block.tsx
```diff
@@ -35,7 +35,10 @@ export const ImageGalleryBlock = ({images, className}: BuilderImageGalleryBlockP
           <ImageListItem key={index}>
             <Image image={image.image!} />
 
-            <Paragraph gutterBottom={false}>{image.caption ?? image.image?.title}</Paragraph>
+            <Paragraph gutterBottom={false}>
```
michael-scheurer, line 38: imo the implementation should be identical like in the image-block.tsx:51 to keep consistency

michael-scheurer (changes requested):

libs/block-content/website/src/lib/image-gallery/image-gallery-block.tsx
Itrulia: We used to show the title before, as it wasn't part of the task I rather not discuss this inside this quick improvement PR.

PR #1724, opened 2025-01-14
revert: events federation

Itrulia:


Itrulia:
@tomaszdurka FYI, events were broken since federation as the V2 events were used in V1.

libs/event/api/src/lib/event.model.ts
```diff
@@ -37,8 +36,7 @@ registerEnumType(EventStatus, {
 })
 
 @ObjectType()
-@Directive('@key(fields: "id")')
-export class Event {
+export class EventV2 {
```
Itrulia, line 39: Renamed just to be sure as imported events use it

elias-summermatter (approved):

PR #1728, opened 2025-01-15
feat(tsri): Add possibility to show different banners based on login status

LukasSkywalker:
This change allows media to choose whether a banner should be shown only for logged in, logged out or for every user. By default, the banner is shown for everyone.


apps/tsri/src/components/website-builder-overwrites/banner/tsri-banner.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+[data-role='OTHER'] button {
+  background-color: ${theme.palette.primary.main};
+  color: ${theme.palette.primary.contrastText};
+  border: 2px solid ${theme.palette.primary.contrastText};
+}
+`
+)
+
+export function TsriBanner(props: BuilderBannerProps) {
```
Itrulia, line 46: Can just export TsriBannerInner as TsriBanner, no?

libs/api/prisma/schema.prisma
```diff
@@ -1262,12 +1262,19 @@ model Banner {
...
   showOnArticles Boolean @default(true)
   showOnPages    Page[]
+  showForLoginStatus LoginStatus @default(ALL)
 
   actions BannerAction[]
 
   @@map("banners")
 }
 
+enum LoginStatus {
```
Itrulia, line 1272: Naming schema is different from other enums we have I believe

Itrulia (approved):

PR #1735, opened 2025-01-20
feat(share): Integrate ProLitteris Tracking Pixel

elias-summermatter:


apps/api-example/src/app/index.ts
```diff
@@ -19,6 +19,8 @@ import {Application} from 'express'
 import {DefaultURLAdapter} from '../urlAdapters'
 import {readConfig} from '../readConfig'
 import {MannschaftURLAdapter} from '../urlAdapters/URLAdapter-mannschaft'
+import {TrackingPixelProvider} from '../../../../libs/tracking-pixel/api/src/lib/tracking-pixel-provider/tracking-pixel-provider'
```
Itrulia, line 22: Imports are not using the library but the file directly

elias-summermatter: Still in draft ;-) dont worry

apps/api-example/src/readConfig.ts
```diff
@@ -153,13 +154,20 @@ type Turnstile = {
   siteKey: string
 }
 
+interface ProLitteris extends ProLitterisCountPixelProps {
+  type: 'prolitteris'
+}
+
+type TackingPixels = ProLitteris
```
LukasSkywalker, line 161: Tacking -> Tracking :)

apps/api-example/src/nestapp/app.module.ts
```diff
@@ -132,6 +134,35 @@ import {PollModule} from '@wepublish/poll/api'
...
       global: true
     }),
+    TrackingPixelModule.registerAsync({
+      imports: [ConfigModule],
+      useFactory: async (config: ConfigService) => {
+        const trackingPixelProviders: TrackingPixelProvider[] = []
+        const configFile = await readConfig(config.getOrThrow('CONFIG_FILE_PATH'))
+
+        const trackingPixelProvidersRaw = configFile.trackingPixelProviders
+        if (trackingPixelProvidersRaw) {
```
michael-scheurer, line 144: Would love to remove parts of this heavy nesting with

`if (!trackingPixelProvidersRaw) {return trackingPixelProviders}`

And the same within the for-loop:
`if (trackingPixelProvider.type !== 'prolitteris') {continue}`

libs/api/prisma/schema.prisma
```diff
@@ -123,14 +123,33 @@ model Article {
...
+  prolitteris @map("prolitteris")
+}
+
+model ArticleTrackingPixels {
+  id                        String                    @id @default(cuid())
+  article                   Article                   @relation(fields: [articleId], references: [id], onDelete: Cascade)
+  articleId                 String
+  createdAt                 DateTime                  @default(now())
+  modifiedAt                DateTime                  @updatedAt
+  trackingPixelProviderID   String
```
michael-scheurer, line 145: I would move the trackingPixelProvider into a separate table. I am missing normalization here.

michael-scheurer: If we want to add further fields later, it would really help. Maybe we would like to add a name or a description. Or even configurations...

michael-scheurer (commented):
Really cool work! I've got 2 annotations.

michael-scheurer (approved):

PR #1736, opened 2025-01-21
refactor(website): replace mui styled with emotion

Itrulia:
TODO's:

* [x] Add linting rule to disallow importing `styled` from mui


libs/ui/src/lib/theme.tsx
```diff
@@ -1,11 +1,22 @@
...
+  Theme as MaterialTheme
+} from '@mui/material'
 import {createBreakpoints} from '@mui/system'
 
 const {
   palette: {augmentColor},
   breakpoints: originalBreakpoints
 } = createTheme()
 
+declare module '@emotion/react' {
```
Itrulia, line 14: This is the magic that allows us to use the emotion styled now, which fixes our SWC problem

PR #1747, opened 2025-02-03
feat(bajour): Add Spotify link to Basel Briefing (BA-204)

LukasSkywalker:


apps/bajour/src/components/briefing/basel-briefing.tsx
```diff
@@ -391,9 +397,23 @@ export const BaselBriefing = ({teasers, blockStyle}: BaselBriefingProps) => {
                 </Author>
               )}
 
-              <ReadMoreButton variant="outlined" color="inherit" size="small">
-                Ganzes Briefing
-              </ReadMoreButton>
+              <ButtonRow>
+                {(blockStyle as BriefingType) === BriefingType.BaselBriefing && (
+                  <ReadMoreButton
```
Itrulia, line 402: Should use LinkComponent as property instead of using onClick as this now breaks browser semantics.

apps/bajour/src/components/briefing/basel-briefing.tsx
```diff
@@ -391,9 +397,23 @@ export const BaselBriefing = ({teasers, blockStyle}: BaselBriefingProps) => {
                 </Author>
               )}
 
-              <ReadMoreButton variant="outlined" color="inherit" size="small">
-                Ganzes Briefing
-              </ReadMoreButton>
+              <ButtonRow>
+                {(blockStyle as BriefingType) === BriefingType.BaselBriefing && (
```
Itrulia, line 401: could use 

```suggestion
                {isBaselBriefing({blockStyle}) && (
```

instead

michael-scheurer (approved):

apps/bajour/src/components/briefing/basel-briefing.tsx
LukasSkywalker: @Itrulia The entire component is already wrapped in a `<LinkWrapper>` so the entire element is clickable. Nesting interactive elements is not allowed. We could wrap the individual parts in a link though. What do you think?

Itrulia: Then it shouldn't be interactable else we fail WCAG guidelines. 
Either have the elements themselves interactable or move the Spotify link out into a different part (and design)

PR #1753, opened 2025-02-08
feat(comments-api): spam protection via akismet

Itrulia:
Small idea to have a cron checking the comment queue every few minutes to automatically filter the comment queue.

Module could be only loaded for specific media's or maybe even via settings in editor?


package.json
```diff
@@ -211,6 +211,7 @@
     "@apollo/gateway": "^2.8.4",
     "@apollo/server": "^4.9.1",
     "@apollo/subgraph": "^2.2.3",
+    "@cedx/akismet": "^17.1.0",
```
Itrulia, line 214: Requires node22

needs `npm i --force`

libs/comments/api/src/lib/spam-protection/spam-protection.service.ts
```diff
@@ -0,0 +1,97 @@
...
+          take: 1
+        }
+      }
+    })
+
+    for (const comment of openComments) {
+      const result = await this.client.checkComment(
+        new Comment({
+          author: new Author({
+            email: comment.user?.email,
```
Itrulia, line 50: GDPR? I guess would just needed to be added to the privacy policy.

PR #1761, opened 2025-02-12
Wnti 1 add mitmachen block

LukasSkywalker:


Itrulia (approved):

libs/membership/website/src/lib/payment-amount/payment-amount-picker/payment-amount-picker.tsx
```diff
@@ -88,14 +86,33 @@ export const PaymentAmountPickerItem = forwardRef<HTMLButtonElement, PaymentAmou
...
 
+    const pickerItems = useMemo(() => {
+      switch (siteTitle) {
+        case 'Wnti':
+          return slug?.includes('spende') ? [10000, 15000, 20000] : [1000, 1500, 2000]
+        default:
+          return [1000, 1500, 2000]
+      }
+    }, [siteTitle, slug])
+
```
LukasSkywalker, line 115: This is a workaround until we (maybe?) allow to set the proposed amounts in the editor.

PR #1762, opened 2025-02-12
fix(bajour): make primary banner optional

LukasSkywalker:
## Summary by CodeRabbit

- **Bug Fixes**
  - Updated primary banner handling to gracefully manage cases when a banner isn’t available, ensuring a smoother and more reliable display.
  - Enhanced banner display safety so that interfaces render consistently without runtime issues when banner data is missing.
- **New Features**
  - The `primaryBanner` query now allows for a null return value, indicating that a primary banner may not always be present.
  - Improved error handling in the banner resolver to throw an exception when a requested banner is not found.


LukasSkywalker:
@coderabbitai review plz

Itrulia (approved):

PR #1765, opened 2025-02-13
fix(tsri): ad header causing a big padding when none are defined

Itrulia:


apps/tsri/src/components/tsri-ad-header.tsx
```diff
@@ -67,11 +74,12 @@ export default function TsriAdHeader({authors}: {authors?: BuilderAuthor[]}) {
 
           <TsriAdvertiserContent>
             {isPromo(advertiser) ? (
-              <b>Rubrik Kultur wird präsentiert von: </b>
```
Itrulia, line 70: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b

LukasSkywalker (approved):

PR #1773, opened 2025-02-19
BA-204: Shorten briefing button label on mobile

LukasSkywalker:
This reduces the size of the briefing buttons on mobile devices and makes sure they do not wrap their content on two lines.

## Summary by CodeRabbit

- **New Features**
	- Enhanced briefing buttons to display extended text on larger screens, providing a clearer and more responsive user interface across different devices.

### DIFF
--- a/apps/bajour/src/components/briefing/basel-briefing.tsx
+++ b/apps/bajour/src/components/briefing/basel-briefing.tsx
@@ -318,6 +318,14 @@ export const Avatar = styled(Image)`
   }
 `
 
+const DesktopOnly = styled('span')`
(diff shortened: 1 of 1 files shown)


apps/bajour/src/components/briefing/basel-briefing.tsx
```diff
@@ -318,6 +318,14 @@ export const Avatar = styled(Image)`
   }
 `
 
+const DesktopOnly = styled('span')`
```
Itrulia, line 321: I guess should be HideOnMobile instead?

PR #1785, opened 2025-02-26
fix(image-website, block-content-website): load ATF images eagerly

Itrulia:


libs/image/website/src/lib/image.context.tsx
```diff
@@ -9,7 +9,7 @@ export const useImageProps = (props: BuilderImageProviderProps): BuilderImagePro
 
   return useMemo(
     () =>
-      mergeDeepLeft(
```
Itrulia, line 12: merged the opposite way than wanted

michael-scheurer (approved):

PR #1801, opened 2025-03-04
feat(payrex): Fix payrexx payment without preAuthorization if not use…

elias-summermatter:
…d as offsession paymentprovider


elias-summermatter:
@tomaszdurka please make a thorough review :-)

michael-scheurer (approved):

PR #1803, opened 2025-03-05
feat(flimmer): onboarding

michael-scheurer:


apps/flimmer/pages/_app.tsx
```diff
@@ -104,6 +105,19 @@ const dateFormatter = (date: Date, includeTime = true) =>
...
 
+const MitmachenInner = (props: SubscribePageProps) => (
+  <SubscribePage fields={['firstName']} {...props} />
+)
+
+const MitmachenInnerStyled = styled(MitmachenInner)`
+  border: 1px solid ${({theme}) => theme.palette.accent.main};
+  border-radius: ${({theme}) => theme.shape.borderRadius}px;
+  padding: ${({theme}) => theme.spacing(4)};
+  ${RichTextBlockWrapper} {
```
Itrulia, line 116:
```suggestion

  ${RichTextBlockWrapper} {
```

apps/flimmer/pages/mitmachen.tsx
Itrulia, line 1: I see flimmer having a subscribe page set in the _app but this is deleted, why? This page is required if we have subscriptions right now.

libs/utils/website/src/lib/pages/subscribe-page.tsx
```diff
@@ -7,7 +7,7 @@ import {ComponentProps} from 'react'
 import {ssrAuthLink} from '../auth-link'
 import {getSessionTokenProps} from '../get-session-token-props'
 
-type SubscribePageProps = Omit<ComponentProps<typeof SubscribeContainer>, ''>
+export type SubscribePageProps = Omit<ComponentProps<typeof SubscribeContainer>, ''>
```
Itrulia, line 10:
```suggestion type SubscribePageProps = Omit<ComponentProps<typeof SubscribeContainer>, ''>
```

apps/flimmer/pages/_app.tsx
```diff
@@ -104,6 +105,19 @@ const dateFormatter = (date: Date, includeTime = true) =>
     ? `${format(date, 'dd. MMMM yyyy')} um ${format(date, 'HH:mm')}`
     : format(date, 'dd. MMMM yyyy')
 
+const MitmachenInner = (props: SubscribePageProps) => (
```
Itrulia, line 108:
```suggestion
const MitmachenInner = (props: ComponentProps<typeof SubscribePage>) => (
```

Itrulia (changes requested):

apps/flimmer/pages/mitmachen.tsx
michael-scheurer: we added it simply as a page within the editor using the subscription block. works very well and the media has the freedom to change the mitmachen page with text, images etc.

Itrulia: That page is still required though else it breaks workflows.

The navigation for example has to be hardcoded to a page right now, can't SSR with user data on the block so you can't send a JWT querystring with it and have the user be logged in etc.

libs/utils/website/src/lib/pages/subscribe-page.tsx
Itrulia: oh damn, was menat to revert back to `type SubscribePageProps = Omit<ComponentProps<typeof SubscribeContainer>, ''>` not fully remove it.

apps/flimmer/pages/mitmachen.tsx
michael-scheurer: done

Itrulia (approved):

PR #1806, opened 2025-03-05
enh(wep): profile view with better navigation and alignments

michael-scheurer:
Enhances profile dashboard. Unpaid invoice is highlighted, subscription and invoice sheets are centered where needed to align with the profile section. Also links back to the profile and titles are provided for a better navigation experience to the user.


elias-summermatter (approved):

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -199,6 +199,18 @@ export const NavbarLogoWrapper = styled('div')`
 
 export const NavbarSpacer = styled('div')``
 
+const OpenInvoiceButton = styled(Button)`
+  &::after {
+    content: 'Rechnung';
```
Itrulia, line 204: Shouldn't be done with before and after as it's an accessibility issue.
Use https://mui.com/system/display/

Itrulia (changes requested):

PR #1812, opened 2025-03-12
feat: crowdfunding

michael-scheurer:
This pull request adds crowdfunding functionality to editor and Website-Builder.

**Editor**
- List view and single view to create and edit a crowdfunding with multiple goals, filters and external funding
image

- Select crowdfunding within a page or article as block
image

**Website**
- Shows the current crowdfunding goal and the total funded amount.
image


libs/block-content/website/src/lib/crowdfunding/crowdfunding-block.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import {CrowdfundingBlock} from './crowdfunding-block'
+import {mockCrowdfundingBlock} from '@wepublish/storybook/mocks'
+
+export default {
+  component: CrowdfundingBlock,
+  title: 'Blocks/Crowdfunding'
+} as Meta
+
+export const Default: StoryObj = {
+  args: mockCrowdfundingBlock()
```
Itrulia, line 11: 👍🏼

libs/block-content/website/src/lib/crowdfunding/crowdfunding-block.tsx
```diff
@@ -0,0 +1,132 @@
...
+  >(() => {
+    return crowdfunding?.activeCrowdfundingGoal
+  }, [crowdfunding?.activeCrowdfundingGoal])
+
+  const progress = useMemo<number>(
+    () => activeCrowdfunding?.progress || 0,
+    [activeCrowdfunding?.progress]
+  )
+
+  const revenue = useMemo<number>(() => crowdfunding?.revenue || 0, [crowdfunding?.revenue])
```
Itrulia, line 71:
```suggestion
  const revenue = crowdfunding?.revenue || 0,
```

Probably useMemo uses more CPU cycles than we save here (same with the ones below

libs/block-content/website/src/lib/crowdfunding/crowdfunding-block.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const goalAmount = useMemo<number>(
+    () => activeCrowdfunding?.amount || 0,
+    [activeCrowdfunding?.amount]
+  )
+  const goalDescription = useMemo(
+    () => activeCrowdfunding?.description,
+    [activeCrowdfunding?.description]
+  )
+
+  function formatMoney(amount: number): string {
```
Itrulia, line 82: We have `formatCurrency` in membership/website, should use that instead.

In general, this function shouldn't be inside the component as it will be recreated for each render.

libs/block-content/website/src/lib/crowdfunding/crowdfunding-block.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return Math.round(amount / 100).toLocaleString('de-CH', {
+      minimumFractionDigits: 2,
+      maximumFractionDigits: 2
+    })
+  }
+
+  // no crowdfunding object is available
```
Itrulia, line 89: nit: unnecessary comment as the simple if below says the same thing

libs/block-content/website/src/lib/crowdfunding/crowdfunding-block.tsx
```diff
@@ -0,0 +1,132 @@
...
+  if (!crowdfunding)
+    return (
+      <Alert severity="error">
+        <AlertTitle>Crowdfunding nicht verfügbar.</AlertTitle>
```
Itrulia, line 93: Do we want to show an error or just hide the component?

libs/block-content/website/src/lib/crowdfunding/crowdfunding-block.tsx
```diff
@@ -0,0 +1,132 @@
...
+            variant="determinate"
+            color="primary"
+            value={progress}
+            sx={{height: '60px'}}
+          />
+          <CfProgressBarInner>
+            <CfProgressBarInnerItem>
+              {goalDescription && (
+                <Tooltip title={goalDescription}>
+                  <IconButton>
```
Itrulia, line 115: get this from the website builder

libs/crowdfunding/api/package.json
Itrulia, line 1: We don't want to access the DB anymore in tests but mock it, so this shouldn't be needed.

libs/crowdfunding/editor/package.json
Itrulia, line 1: Don't need this here

libs/crowdfunding/editor/src/lib/create-crowdfunding-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const {StringType} = Schema.Types
+  const validationModel = Schema.Model({
+    name: StringType().isRequired()
+  })
+
+  const [createCrowdfunding, {loading}] = useCreateCrowdfundingMutation({
+    client,
+    onError: showError,
+    onCompleted: (crowdfunding: CreateCrowdfundingMutation) => {
+      console.log('should close', shouldClose)
```
Itrulia, line 43: Should probably be removed

libs/crowdfunding/editor/src/lib/create-crowdfunding-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export const CreateCrowdfundingForm = () => {
+  const navigate = useNavigate()
+  const {t} = useTranslation()
+  const client = useMemo(() => getApiClientV2(), [])
+
+  const closePath = '/crowdfundings'
+
+  const [crowdfunding, setCrowdfunding] = useState({} as CreateCrowdfundingInput)
```
Itrulia, line 31: The way `setCrowdfunding` is currently used *can* lead to bugs as setStates can be batched and thus `crowdfunding` be outdated.
To make this easier we can use `useReducer` here instead, so instead of doing:

```
setCrowdfunding({
      ...crowdfunding,
      goals: [...(crowdfunding.goals || []), goal]
    })
```

one would do 

```
setCrowdfunding({
      goals: [...(crowdfunding.goals || []), goal]
    })
```

The useReducer would look like this:


```suggestion
  const [crowdfunding, setCrowdfunding] = useReducer((state: CreateCrowdfundingInput, action: Partial<CreateCrowdfundingInput>) => ({...state, ...action}), {})
```

libs/crowdfunding/editor/src/lib/crowdfunding-list.tsx
```diff
@@ -0,0 +1,99 @@
...
+import {CrowdfundingDeleteModal} from './crowdfunding-delete-modal'
+
+const {Column, HeaderCell, Cell: RCell} = RTable
+
+function CrowdfundingList() {
+  const {t} = useTranslation()
+
+  const [crowdfundingDelete, setCrowdfundingDelete] = useState<Crowdfunding | undefined>(undefined)
+
+  const client = useMemo(() => getApiClientV2(), [])
```
Itrulia, line 26:
```suggestion
  const client = getApiClientV2()
```

the function getApiClientV2 is already memoized.

libs/crowdfunding/editor/src/lib/edit-crowdfunding-form.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const {t} = useTranslation()
+
+  const closePath = '/crowdfundings'
+
+  const [crowdfunding, setCrowdfunding] = useState<UpdateCrowdfundingInput>({
+    id: crowdfundingId,
+    name: ''
+  })
+
+  const client = useMemo(() => getApiClientV2(), [])
```
Itrulia, line 37:
```suggestion
  const client = getApiClientV2()
```

the function getApiClientV2 is already memoized.

libs/crowdfunding/editor/src/lib/edit-crowdfunding-form.tsx
```diff
@@ -0,0 +1,131 @@
...
+  useCrowdfundingQuery({
+    client,
+    variables: {
+      id: id!
+    },
+    skip: !id,
+    onError: showError,
+    onCompleted: data => {
+      const {__typename, ...inputWithoutTypename} = data.crowdfunding
```
Itrulia, line 46: can also use the helper: `stripTypename`

libs/crowdfunding/api/src/lib/crowdfunding.model.ts
```diff
@@ -0,0 +1,98 @@
...
+  activeCrowdfundingGoal?: CrowdfundingGoalWithProgress
+}
+
+@ArgsType()
+export class CrowdfundingId {
+  @Field(() => ID)
+  id!: string
+}
+
+@InputType()
```
Itrulia, line 74: This should be an ArgsType not an InputType.

This currently causes us to have to:

```
variables: {
	input: {
		...
	}
}
```

instead of 

```
variables: {
	...
}
```

Itrulia (changes requested):

libs/crowdfunding/api/src/lib/crowdfunding-dataloader.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    return this.dataloader.load(...parameters)
+  }
+
+  public loadMany(
+    ...parameters: Parameters<DataLoader<string, CrowdfundingWithActiveGoal | null>['loadMany']>
+  ) {
+    return this.dataloader.loadMany(...parameters)
+  }
+
+  private async loadCrowdfundingWithActiveGoals(
```
Itrulia, line 38: Like coderabbit said this is an actual bug

libs/crowdfunding/api/src/lib/crowdfunding-dataloader.service.ts
```diff
@@ -0,0 +1,74 @@
...
+        goals: true,
+        memberPlans: true
+      }
+    })
+
+    if (!crowdfunding) {
+      return []
+    }
+
+    const crowdfundingService = new CrowdfundingService(this.prisma)
```
Itrulia, line 58: Don't instantiate services manually. Use the DI Container for this.
But generally dataloaders should not load more than 1 thing, this should probably be in a resolver.

PR #1820, opened 2025-03-19
feat(website): replace navbar urls with full btn props

michael-scheurer:
Add some flexibility to navbar.


Itrulia:
whats the rationale for this change? ButtonProps also doesnt seem like a good choice as the props are more like LinkProps due to LinkComponent.

PR #1824, opened 2025-03-21
style: changed prettier config

Itrulia:
@tomaszdurka @LukasSkywalker @elias-summermatter @michael-scheurer: Please discuss! This is just my idea of some improvements. If you disagree I want to hear it.

The only thing I care about are the trailing commas!

 If you wanna see an example of the changes, you can checkout the PR and run:
```bash
$ npm run prettier
```

I have not committed them yet as it would be around 2k files changed.


.prettierrc
```diff
@@ -1,9 +1,11 @@
 {
-  "printWidth": 100,
```
Itrulia, line 2: Recommended by the prettier team to not be longer than 80

.prettierrc
```diff
@@ -1,9 +1,11 @@
...
-  "semi": false,
```
Itrulia, line 3: Due to how typescript works, we currently have some ugly things going on because this isn't set to true.

`;(GuardedProfile as any).getInitialProps` for example as we need the semicolon to tell the browser it's not a function.

.prettierrc
```diff
@@ -1,9 +1,11 @@
...
+  "printWidth": 80,
+  "semi": true,
   "singleQuote": true,
-  "bracketSpacing": false,
```
Itrulia, line 5: makes things more readable imo

.prettierrc
```diff
@@ -1,9 +1,11 @@
...
-  "jsxBracketSameLine": true,
-  "trailingComma": "none",
```
Itrulia, line 7: Allows for better refactoring, not constantly having to add a comma when commenting something out and back in.

.prettierrc
```diff
@@ -1,9 +1,11 @@
 {
-  "printWidth": 100,
-  "semi": false,
+  "printWidth": 80,
+  "semi": true,
   "singleQuote": true,
-  "bracketSpacing": false,
-  "jsxBracketSameLine": true,
```
Itrulia, line 6: makes things more readable when set to false imo.

.prettierrc
michael-scheurer: I like to have it without. But there are some rare cases it can lead to faults. I think @LukasSkywalker will be all for it ;) So I could agree.

.prettierrc
michael-scheurer: hallelujah <3

michael-scheurer (approved):

.prettierrc
LukasSkywalker: Oooooh yeah. Michael knows me too well. I prefer the explicit semicolon.

.prettierrc
LukasSkywalker: And better diffs. +1!

.prettierrc
LukasSkywalker: Ha, this one is funny. I looked it up in the docs, because `bracketSpacing` sounds as if it would add spaces to brackets, which I prefer not to have, but it does seem to add them to braces, which is nice.

LukasSkywalker (approved):
This really addresses all pains I had with the current prettier config. Explicit semicolons, trailing commas, spacing in braces. Nice.

.prettierrc
```diff
@@ -1,9 +1,11 @@
...
-  "trailingComma": "none",
+  "bracketSpacing": true,
+  "bracketSameLine": false,
+  "experimentalTernaries": true,
```
elias-summermatter, line 7: Do we really want to use experimental stuff?

Itrulia: It will be default next version, it's experimental so people can migrate to it beforehand.

PR #1839, opened 2025-03-31
fix(editor): slate-react is not a controlled component anymore

Itrulia:
see: https://github.com/ianstormtaylor/slate/issues/4992


apps/editor/src/app/routes/polls/pollEditView.tsx
```diff
@@ -90,16 +90,12 @@ function PollEditView() {
   useEffect(() => {
     if (data?.poll) {
       setPoll(data.poll)
-    } else {
-      setPoll(undefined)
```
Itrulia, line 94: This lead to a race condition

PR #1847, opened 2025-04-02
feat(api): introduce TeaserSlots (api, prisma, blocks)

tomaszdurka:
This pull request introduces **Teaser Slots**, a new content block intended to replace both the *Teaser Grid* and *Teaser List* components.

**Teaser Slots** are composed of individual slots that are **autofilled by default**. Autofill can be configured to use specific tags, giving you control over the type of content that appears. You can also manually override any slot with a specific teaser — such as an article, event, page, or custom teaser.

Manual and automatic teasers can **coexist** within the same block. When a slot is manually filled, it’s skipped during autofill, and the remaining empty slots are filled automatically.

The system preserves the state of loaded teasers across different *Teaser Slots* blocks. This prevents duplication, even when using multiple blocks with different filters. **Manually set teasers are also considered** — if a teaser is placed manually in a specific slot, it will not be included again through autofill in the same or other blocks.

Let me know if you have any questions or feedback!


libs/ui/editor/src/lib/blocks/blockMap.tsx
```diff
@@ -39,6 +39,8 @@ import {EditorBlockType, TeaserListBlockSort, TeaserType} from '@wepublish/edito
 import {isFunctionalUpdate} from '../utility'
 import {TeaserListBlock} from './teaserListBlock'
 import {SubscribeBlock} from './subscribeBlock'
+import {TeaserSlotsBlock} from './teaserSlotsBlock'
+import {TeaserSlotType} from '@wepublish/block-content/api'
```
Itrulia, line 43: Importing API here breaks the build. should be imported from `@wepublish/editor/api-v2`

libs/block-content/api/src/lib/teaser-slot/teaser-slots.model.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  @Field(() => [TeaserSlot])
+  slots!: Array<TeaserSlot>
+
+  @Field(() => [Teaser], {
+    // not nullable but teaser grid/flex can be and it uses the same property,
+    // so they have to be the same
+    nullable: 'items'
+  })
+  teasers!: Array<typeof Teaser | null>
```
Itrulia, line 29:
```suggestion
  teasers!: Array<typeof Teaser>
```

libs/block-content/website/src/lib/teaser/teaser-slots-block.tsx
Itrulia, line 1: Could we maybe change the output of the block that the website is using, so the website doesn't need to care if it's  auto filled or not?

Then we could possibly even re-use the teaser-grid component.

libs/ui/editor/src/lib/atoms/editorTemplate.tsx
```diff
@@ -1,10 +1,21 @@
...
-  max-width: 920px;
+  max-width: 1220px;
+
+  > * > * {
+    margin-left: auto;
+    margin-right: auto;
+    max-width: 920px;
+
+    :has(${TeaserSlotsBlockWrapper}) {
+      max-width: 1220px;
```
Itrulia, line 16: This makes it look weird in the editor

michael-scheurer (approved):
Now we should try to deprecate all the other teaser blocks. Do you have some idea about it?

Itrulia (changes requested):
.

libs/block-content/api/src/lib/teaser/teaser-flex.model.ts
```diff
@@ -64,3 +64,7 @@ export class TeaserGridFlexBlockInput extends OmitType(
   @Field(() => [FlexTeaserInput])
   flexTeasers!: FlexTeaserInput[]
 }
+
+export function isTeaserGridFlexBlock(block: BaseBlock<BlockType>): block is TeaserGridFlexBlock {
```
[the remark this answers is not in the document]
Itrulia: Others dont have that tho

Itrulia (approved):

libs/block-content/api/src/lib/teaser/slot-teasers-loader.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      this.loadedTeasers.push(...teasers)
+      return teasers
+    }
+
+    if (teaserType === TeaserType.Event) {
+      const pages = await this.eventService.getEvents({
+        filter: {
+          tags: filter?.tags
+        },
```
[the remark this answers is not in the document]
Itrulia: @tomaszdurka I think the AI is right here

PR #1849, opened 2025-04-03
Fix teaser grid to accept nulls

tomaszdurka:


Itrulia (changes requested):
Why?

libs/block-content/api/src/lib/teaser/teaser-list.model.ts
```diff
@@ -52,7 +52,7 @@ export class TeaserListBlock extends BaseBlock<BlockType.TeaserList> {
     // so they have to be the same
     nullable: 'items'
   })
-  teasers!: Array<typeof Teaser>
+  teasers!: Array<typeof Teaser | null>
```
Itrulia, line 55: teaser list teasers are never null

Itrulia:
Oh never seen the comment. This fixes one of the bugs!

PR #1853, opened 2025-04-09
Adjust WePublish website

tomaszdurka:


Itrulia (changes requested):
This doesn't work as the main project is called wepublish

PR #1873, opened 2025-04-28
feat(hauptstadt): use http only cookie for login

Itrulia:


libs/authentication/website/src/lib/session.context.ts
```diff
@@ -1,10 +1,10 @@
 import {UserSession, User} from '@wepublish/website/api'
-import {createContext, Dispatch, useContext} from 'react'
+import {createContext, useContext} from 'react'
 
 export const AuthTokenStorageKey = 'auth.token'
 
 export const SessionTokenContext = createContext<
-  [User | null, boolean, Dispatch<UserSession | null>] | null
```
Itrulia, line 7: Allows now for async

libs/utils/website/src/lib/async-session.provider.tsx
```diff
@@ -0,0 +1,75 @@
...
+      }
+    })
+
+    const setCookieAndToken = useCallback(
+      async (newToken: UserSession | null) => {
+        setToken(newToken)
+
+        if (newToken) {
+          sessionStorage.setItem(AuthTokenStorageKey, JSON.stringify(newToken))
+          await fetch('/api/cookie', {
```
Itrulia, line 24: Sets token now via API, also sets it in sessionStorage so that the authLink has access to it.

libs/utils/website/src/lib/auth-link.tsx
```diff
@@ -17,16 +21,22 @@ export const authLink = new ApolloLink((operation, forward) => {
...
   new ApolloLink((operation, forward) => {
-    operation.setContext((context: DefaultContext) => ({
-      ...context,
-      headers: {
-        ...context.headers,
-        authorization: token ? `Bearer ${typeof token === 'function' ? token() : token}` : ''
-      },
-      credentials: 'include'
-    }))
+    operation.setContext(async (context: DefaultContext) => {
```
Itrulia, line 28: Supports now async

libs/utils/website/src/lib/pages/profile/profile-page.tsx
```diff
@@ -152,11 +152,12 @@ GuardedProfile.getInitialProps = async (ctx: NextPageContext) => {
       req: ctx.req,
       res: ctx.res,
       expires: new Date(data.data.createSessionWithJWT.expiresAt),
-      sameSite: 'strict'
+      sameSite: 'strict',
+      httpOnly: true // @TODO: Config
```
Itrulia, line 156: Gotta be set to env variable before merging

libs/utils/website/src/lib/async-session.provider.tsx
```diff
@@ -0,0 +1,75 @@
...
+            method: 'DELETE'
+          })
+        }
+      },
+      [getMe]
+    )
+
+    const setupInitialSessionToken = useCallback(async () => {
+      if (sessionToken) {
```
Itrulia, line 44: This is when the token comes from `getInitialProps` or `getServerSideProps`, then we don't need to fetch it from the API

libs/utils/website/src/lib/auth-link.tsx
```diff
@@ -3,7 +3,11 @@ import {AuthTokenStorageKey} from '@wepublish/authentication/website'
 import {getCookie} from 'cookies-next'
 
 export const authLink = new ApolloLink((operation, forward) => {
-  const {token} = JSON.parse(getCookie(AuthTokenStorageKey)?.toString() ?? '{}')
+  const {token: cookieToken} = JSON.parse(getCookie(AuthTokenStorageKey)?.toString() ?? '{}')
+  const {token: sessionStorageToken} = JSON.parse(
```
Itrulia, line 7: Add sessionStorage support to it, no need for separate authLink IMO so just baked it into here.

PR #1888, opened 2025-05-06
refactor(editor): react deprecation & possibleTypes build time

Itrulia:


apps/editor/src/main.tsx
```diff
@@ -26,42 +27,6 @@ Sentry.init({
   replaysOnErrorSampleRate: 1.0
 })
 
-// See: https://www.apollographql.com/docs/react/data/fragments/#fragments-on-unions-and-interfaces
-export async function fetchIntrospectionQueryResultData(url: string) {
```
Itrulia, line 30: Now generated from codegen just like v2

PR #1891, opened 2025-05-07
fix(editor): tag list in article filter being unusable & pagination

Itrulia:


LukasSkywalker (approved):
Nice one! Works beautifully :)

PR #1910, opened 2025-05-13
feat(api, editor): implement paywall

Itrulia:


Itrulia (commented):
Removed settings from v1 as they were unused

libs/api/src/index.ts
```diff
@@ -34,32 +34,3 @@ export * from './lib/challenges/algebraicCaptchaChallenge'
 export * from './lib/challenges/cfTurnstile'
 
 export {getUserForCredentials} from './lib/graphql/user/user.queries'
-
-export * from '@wepublish/banner/api'
```
Itrulia, line 38: Removed this as we don't really do packages anymore. We also have Membership V2 model that would clash

libs/api/src/lib/graphql/mutation.private.ts
```diff
@@ -766,18 +764,6 @@ export const GraphQLAdminMutation = new GraphQLObjectType<undefined, Context>({
         deleteComment(id, authenticate, comment)
     },
 
-    // Settings
-    // ==========
-
-    updateSettingList: {
```
Itrulia, line 772: Removed settings from v1 as they were unused

libs/ui/editor/src/lib/atoms/memberPlan/selectMemberPlan.tsx
Itrulia, line 1: Similar to selectTags as a reusable component (that supports pagination)

libs/paywall/api/src/lib/paywall.service.ts
Itrulia, line 1: Missing tests

PR #1917, opened 2025-05-19
feat(teaser-list): Add ability to sort articles by updated-at

tomaszdurka:
**Short Description:**
The teaser list now supports sorting by “Updated At,” in addition to “Hot and Trending” and “Published At.” The “Updated At” value starts as the initial publish date but updates whenever the article is republished. This lets you influence sorting by republishing with a new date.


Itrulia (changes requested):
Missing sort application for pages (& possibly events). Else perfect!

Itrulia (approved):

PR #1920, opened 2025-05-19
feat(api): improve loading times of revisions by introducing archivedAt

Itrulia:
Adding `archivedAt` to revisions allows us to always fetch the last 3 revisions because:

1. There can only be 1 pending revision at once
2. There can only be 1 draft revision not being archived
3. We sort by "publishedAt" descending

If we are loading 5 articles we used to have 16 queries. 1 for loading 5 articles and 3 per article for the revisions.
Now if we load 5 articles, we are doing 6 queries. 1 for loading 5 articles and 1 per article for the revisions.

The same applies to pages.

This also fixes the `draft` filter on articles & pages as we can now check if there's an un-archived draft version.

```description
Increase loading speed of articles & pages
Fixing the draft filter for articles & pages
```


libs/api/prisma/migrations/20250519153015_add_archived_at_revisions/migration.sql
```diff
@@ -0,0 +1,29 @@
+-- AlterTable
+ALTER TABLE "articles.revisions" ADD COLUMN     "archivedAt" TIMESTAMP(3);
+UPDATE "articles.revisions" SET "archivedAt" = CURRENT_TIMESTAMP WHERE "publishedAt" IS NULL;
```
Itrulia, line 3: Set all draft revisions to archived

libs/api/prisma/migrations/20250519153015_add_archived_at_revisions/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+
+WITH latest AS (
+    SELECT id
+    FROM "articles.revisions"
+    WHERE "publishedAt" IS NULL
+    ORDER BY "createdAt" DESC
+    LIMIT 1
+)
+UPDATE "articles.revisions"
+SET "archivedAt" = NULL
```
Itrulia, line 13: set latest draft revisions to be unarchived

libs/api/prisma/migrations/20250519153015_add_archived_at_revisions/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+
+WITH latest AS (
+    SELECT id
+    FROM "articles.revisions"
+    ORDER BY "createdAt" DESC
+    LIMIT 1
+)
+UPDATE "articles.revisions"
+SET "archivedAt" = NULL
+WHERE id IN (SELECT id FROM latest) AND "publishedAt" IS NULL;
```
Itrulia, line 13: check for `publishedAt` here to make sure only the latest revision is affected. If we check it in the `WITH` above, it would affect the latest `draft`, even if there should be no more draft revision

libs/article/api/src/lib/article.service.ts
```diff
@@ -194,6 +194,14 @@ export class ArticleService {
         hidden,
         disableComments,
         revisions: {
+          updateMany: {
```
Itrulia, line 197: When creating a new revision, update the old ones to be archived.

image

Made sure this works in the correct order.

libs/article/api/src/lib/article.service.ts
```diff
@@ -630,7 +638,8 @@ const createDraftFilter = (filter: Partial<ArticleFilter>): Prisma.ArticleWhereI
     return {
       revisions: {
         some: {
-          publishedAt: filter.draft ? null : {not: null}
+          publishedAt: filter.draft ? null : {not: null},
+          archivedAt: null
```
Itrulia, line 642: should be set to archivedAt null even when draft is false, as only draft can be archived.

libs/article/api/src/lib/article-revision-dataloader.service.ts
```diff
@@ -19,45 +19,20 @@ export class ArticleRevisionDataloaderService implements Primeable<RevisionMap>
...
-              publishedAt: {
-                gt: new Date()
-              }
-            },
-            orderBy: {
-              createdAt: 'desc'
-            }
-          }),
-          this.prisma.articleRevision.findFirst({
+          this.prisma.articleRevision.findMany({
```
Itrulia, line 22: @elias-summermatter could be further optimized by using raw queries. Do you think it's worth doing it in this step or should we make a ticket as an improvement?

Going from 1+n to 1+1 in terms of queries would be nearly always a win I am guessing.

libs/article/api/src/lib/article.service.ts
```diff
@@ -275,7 +283,8 @@ export class ArticleService {
     // Unpublish existing pending revisions
     await this.prisma.articleRevision.updateMany({
       data: {
-        publishedAt: null
+        publishedAt: null,
+        archivedAt: new Date()
```
Itrulia, line 287: Archive unpublished pending revisions.

elias-summermatter (changes requested):
We discuss this tomorrow during the weekly @tomaszdurka @Itrulia @LukasSkywalker  @michael-scheurer

LukasSkywalker (approved):

PR #1922, opened 2025-05-22
feat(api,editor,website,bajour): add square logo to peer profile & peer information use cta image

Itrulia:
Best of Wepublish uses now the square logo if possible


apps/editor/src/app/locales/de.json
```diff
@@ -953,6 +953,7 @@
         "editPeerInfo": "Peer-Info bearbeiten",
         "getRemote": "Peer Profil abrufen",
         "image": "Bild",
+        "squareImage": "Rechteckiges Bild",
```
Itrulia, line 956:
```suggestion
        "squareImage": "Quadratisches Bild",
```

libs/api/src/lib/graphql/peer.ts
```diff
@@ -20,6 +20,7 @@ export const GraphQLPeerProfileInput = new GraphQLInputObjectType({
   fields: {
     name: {type: new GraphQLNonNull(GraphQLString)},
     logoID: {type: GraphQLString},
+    squareLogoId: {type: GraphQLString},
```
LukasSkywalker, line 23: `squareLogoId` is nullable in the GraphQL API but is required in `PeerProfileEditPanel.tsx`. As far as I can tell, they both could be non-nullable?

libs/peering/website/src/lib/peer-information.tsx
```diff
@@ -25,31 +22,19 @@ export const PeerInformationCTA = styled('div')`
...
+    elements: {Image, Button, Link},
     blocks: {RichText}
   } = useWebsiteBuilder()
 
   return (
     <PeerInformationWrapper className={className}>
       <PeerInformationName>
-        {profile?.logo && <Image image={profile.logo} square css={logoStyles} />}
-
-        <H5 component={'span'}>{profile?.name}</H5>
```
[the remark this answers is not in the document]
Itrulia: Yes, fullly replaced with the cta banner

libs/api/src/lib/graphql/peer.ts
Itrulia: we want user to have to set them, but it has to be nullable in graphql as its a new field.

All the other fields are the same

LukasSkywalker (approved):

PR #1924, opened 2025-05-27
feat(hauptstadt): Design

Itrulia:
Things I touched that needs to be checked on other medias:

- [x] Banner
- [x] Teaser
- [x] Article Authors
- [x] Title Block (uses now H2 instead of H3, + Pretitle)
- [x] Break Block
- [x] Search Page
- [x] Peer Information

To revert before merging:

- [ ] Remove urlAdapter config change from `.github/workflows/on-pr-review.yml`

Todo:
- [x] Add environment variable to determine if async session provider should be used or not


elias-summermatter:
@Itrulia can you close this one since we have an other HS branch?

PR #1931, opened 2025-05-28
feat(website): has 87 handle all payment special cases

michael-scheurer:


libs/membership/website/src/lib/subscribe/subscribe-response-handler.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const {createSubscription} = subscribeResponse.data
+
+    // subscription was paid with existing credit card.
+    // in this case intentSecret can be undefined. So this check must run before checking if intentSecret is empty!
+    if (createSubscription.state === 'paid') {
+      window.location.href = successUrl
+      return
+    }
+
+    if (!createSubscription.intentSecret) {
```
Itrulia, line 74: No intentSecret is expected for some payment providers like Bexio

libs/membership/website/src/lib/subscribe/subscribe-response-handler.tsx
```diff
@@ -0,0 +1,130 @@
...
+    // handle fallback
+    handleError(['Unbekannter Fehler!'])
+  }, [subscribeResponse, handleError, memberPlan, fetchPage])
+
+  // entry point. whenever a subscribe response is passed, this component handles it.
+  useEffect(() => {
+    handleSubscribeResponse()
+  }, [subscribeResponse, handleSubscribeResponse])
+
+  async function onCloseStripeElement(success: boolean) {
```
Itrulia, line 101: This now redirects me even when I close the dialog manually.

libs/website/builder/src/lib/membership.interface.ts
```diff
@@ -36,10 +36,9 @@ export type BuilderSubscriptionListProps = Pick<
   onExtend?: (subscriptionId: string) => Promise<void>
 }
 
-export type BuilderInvoiceListItemProps = FullInvoiceFragment & {
+export type BuilderInvoiceListItemProps = {
   className?: string
-  isSepa?: boolean
-  canPay: boolean
+  invoice: FullInvoiceFragment
```
Itrulia, line 41: We generally don't want to pass full domain models down

PR #1933, opened 2025-06-02
Api v1 public to v2

tomaszdurka:


libs/comments/api/src/lib/comment.resolver.ts
```diff
@@ -0,0 +1,131 @@
...
+    const userId = session?.user?.id || null
+    return this.ratingSystemService.getUserCommentRatings(comment.id, userId)
+  }
+
+  @ResolveField(() => Image, {nullable: true})
+  async guestUserImage(@Parent() comment: Comment) {
+    if (!comment.guestUserImageID) {
+      return null
+    }
+    return {id: comment.guestUserImageID}
```
Itrulia, line 101: Dont think these resolve fields works, no?

libs/comments/api/src/lib/comment.service.ts
```diff
@@ -0,0 +1,360 @@
...
+@Injectable()
+export class CommentService {
+  constructor(
+    private readonly prisma: PrismaClient,
+    private readonly ratingSystem: RatingSystemService,
+    private readonly settingsService: SettingsService,
+    private readonly challengeService: ChallengeService
+  ) {}
+
+  async getPublicChildrenCommentsByParentId(parentId: string, userId: string | null) {
```
Itrulia, line 50: Name probably shouldn't referene piublic

libs/comments/api/src/lib/comment.service.ts
```diff
@@ -0,0 +1,360 @@
...
+      orderBy: {
+        modifiedAt: 'desc'
+      },
+      include: {
+        revisions: {orderBy: {createdAt: 'asc'}},
+        overriddenRatings: true
+      }
+    })
+
+    return comments.map(this.mapCommentToPublicComment)
```
Itrulia, line 67: We shouldn't map these anymore in services.

libs/comments/api/src/lib/comment.service.ts
```diff
@@ -0,0 +1,360 @@
...
+    private readonly settingsService: SettingsService,
+    private readonly challengeService: ChallengeService
+  ) {}
+
+  async getPublicChildrenCommentsByParentId(parentId: string, userId: string | null) {
+    const comments = await this.prisma.comment.findMany({
+      where: {
+        AND: [
+          {parentID: parentId},
+          {OR: [userId ? {userID: userId} : {}, {state: CommentState.approved}]}
```
Itrulia, line 55: This now doesn't return unapproved comments anymore for admins. Use preview mode.

libs/comments/api/src/lib/comment.input.ts
```diff
@@ -0,0 +1,45 @@
+import {Field, InputType} from '@nestjs/graphql'
+import {CommentItemType} from './comment.model'
+import {GraphQLRichText} from '@wepublish/richtext/api'
+import {Node} from 'slate'
+import {CommentItemType as PCommentItemType} from '@prisma/client'
+import {ChallengeInput} from '@wepublish/challenge/api'
+
+@InputType()
+export class CommentInput {
```
Itrulia, line 9: We shouldn't duplicate all these properties but get them from the model.

Itrulia (changes requested):
While this moved things to v2, basically we copied over all the bad parts of v1 over to v2. Which kinda wasn't the idea of the whole rewrite.

1. Using more dataloaders
2. Getting rid of unnecessary mapping of types around
3. Getting rid of the unnecessary nesting of the input variable
4. Using lazy loading
5. Duplicating models

Only looking at comments theres multiple things that won't work anymore.

libs/user-subscription/api/project.json
Itrulia, line 1: shouldn't this be just `subscription` and not `user-subscription`?

PR #1934, opened 2025-06-03
fix(editor): avoid audience chart overflow

michael-scheurer:
This fix resolves a display overflow issue on the audience dashboard, ensuring proper layout rendering.

### DIFF
--- a/apps/editor/src/app/routes/audience/audience-chart.tsx
+++ b/apps/editor/src/app/routes/audience/audience-chart.tsx
@@ -36,7 +36,6 @@ export function AudienceChart({clientFilter, audienceStats, loading}: AudienceCh
     t,
     i18n: {language}
   } = useTranslation()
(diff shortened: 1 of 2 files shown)


apps/editor/src/app/routes/audience/audience-chart.tsx
```diff
@@ -49,8 +49,10 @@ export function AudienceChart({clientFilter, audienceStats, loading}: AudienceCh
 
   // Ensure the ResponsiveContainer does not render outside the viewport during initial load.
```
Itrulia, line 50: can't we now just do `const readyRenderChart = !!audienceStats.length`

LukasSkywalker (changes requested):
Not directly related to this PR, but maybe you could add it while you're working on this.

`apps/editor/src/app/routes/audience/audience-filter.tsx` should have a key prop on the `<Col>` on line 146.

PR #1950, opened 2025-06-11
Api v1 public to v2 (second part, payments, and whatever is left)

tomaszdurka:


libs/api/src/lib/media/karmaMediaAdapter.ts
```diff
@@ -96,7 +96,7 @@ export class KarmaMediaAdapter implements MediaAdapter {
   }
 
   async getImageURL(
-    {id, filename, extension, focalPoint}: ImageWithFocalPoint,
+    {id, filename, extension, focalPoint}: Image,
```
elias-summermatter, line 99: Why did you remove her the focal point from the karma media Adapter getImageURL function?

PR #1951, opened 2025-06-11
Add paywall bypass

LukasSkywalker:


libs/paywall/website/src/lib/use-paywall-bypass-token.ts
```diff
@@ -0,0 +1,25 @@
+import {useEffect} from 'react'
+import {useRouter} from 'next/router'
+import {PaywallBypassService} from './paywall-bypass'
+
+export const usePaywallBypassToken = () => {
```
Itrulia, line 5: No need to be a hook

libs/paywall/website/src/lib/use-paywall-bypass-token.ts
```diff
@@ -0,0 +1,25 @@
...
+  const router = useRouter()
+
+  useEffect(() => {
+    if (!router.isReady) return
```
Itrulia, line 9: No need to check, when router isn't ready, router.query is empty

libs/paywall/website/src/lib/paywall-bypass.ts
```diff
@@ -0,0 +1,42 @@
+const BYPASS_STORAGE_KEY = 'paywall_bypass_token'
+const BYPASS_EXPIRY_HOURS = 4
+
+interface BypassTokenData {
+  token: string
+  expiresAt: number
+}
+
+export class PaywallBypassService {
```
Itrulia, line 9: No need for a static class. Simply export functions.

libs/paywall/website/src/lib/use-paywall-bypass-token.ts
```diff
@@ -0,0 +1,25 @@
+import {useEffect} from 'react'
+import {useRouter} from 'next/router'
```
Itrulia, line 2: Unfortunately, the website builder shouldn't rely on the framework implementation directly (removes vendor lock in of Next).

Next actually allows and suggests to use the history API directly: https://nextjs.org/docs/app/getting-started/linking-and-navigating#windowhistoryreplacestate

PR #1956, opened 2025-06-11
feat(website, api): allow preview without slug

Itrulia:


libs/nest-modules/src/lib/urls/url-adapter.ts
```diff
@@ -20,15 +20,15 @@ export class URLAdapter {
...
+    return article.slug ? `${this.baseURL}/a/${article.slug}` : `${this.baseURL}/a/id/${article.id}`
   }
 
   async getArticlePreviewUrl(article: Article) {
     return `${await this.getArticleUrl(article)}?preview`
   }
 
   async getPageUrl(page: Page) {
-    return `${this.baseURL}/${page.slug}`
+    return page.slug != null ? `${this.baseURL}/${page.slug}` : `${this.baseURL}/id/${page.id}`
```
Itrulia, line 31: checking for != here as front slug is empty.

PR #1961, opened 2025-06-17
feat(website, editor): add JWT to preview link

Itrulia:


libs/utils/website/src/lib/with-session-provider.tsx
Itrulia, line 1: Moved to HoC so that `withJwtHandler` has access to it.

libs/ui/editor/src/lib/blocks/richTextBlock/rich-text-block.tsx
```diff
@@ -45,6 +45,12 @@ const Divider = styled(ToolbarDivider)`
   height: 1.8em;
 `
 
+const SlateWrapper = styled.div`
```
Itrulia, line 48: Fixes https://wepublish.atlassian.net/browse/TSRI-142

elias-summermatter:
@Itrulia i reviewed the api part :-)

libs/editor/api/src/lib/schemas/auth.graphql
```diff
@@ -42,3 +40,10 @@ mutation CreateSessionWithJWT($jwt: String!) {
     token
   }
 }
+
+query CreateJWTForUser($userId: String!) {
```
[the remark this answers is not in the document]
elias-summermatter: Yes that sounds good :-) because i think it would not work otherwise :-)

michael-scheurer (approved):

PR #1962, opened 2025-06-22
Improve payment handling

Itrulia:


libs/payment/api/src/lib/payment-provider/never-charge-payment-provider.ts
```diff
@@ -22,7 +22,7 @@ export class NeverChargePaymentProvider extends BasePaymentProvider {
   async createIntent(props: CreatePaymentIntentProps): Promise<Intent> {
     return {
       intentID: `no_charge_${new Date().getTime()}`,
-      intentSecret: 'no_charge',
```
Itrulia, line 25: empty intentSecret = redirect to success

PR #2037, opened 2025-08-18
fix: banners width

tomaszdurka:


elias-summermatter:
@LukasSkywalker & @Itrulia can we close this in your view? If not why?

PR #2052, opened 2025-08-26
fix(core): Hotfix exclude webhooks from json parser

elias-summermatter:


apps/api-example/src/main.ts
```diff
@@ -22,7 +24,22 @@ async function bootstrap() {
...
+  const skipPrefixes = [`/${MAIL_WEBHOOK_PATH_PREFIX}`, `/${PAYMENT_WEBHOOK_PATH_PREFIX}`] as const
+  const skipPrefixesLength = skipPrefixes.length
+  const jsonParser = json({limit: MAX_PAYLOAD_SIZE})
+
+  // Apply JSON parsing only when the path doesn't match any webhook prefix (plain for loop for maximum speed and minimal overhead)
+  const conditionalJson: RequestHandler = (req: Request, res: Response, next: NextFunction) => {
+    const path: string = req.path ?? req.url
+    for (let i = 0; i < skipPrefixesLength; i++) {
+      const p = skipPrefixes[i]
+      if (path === p || path.startsWith(p + '/')) return next()
```
LukasSkywalker, line 37: Maybe we can add `{}` around the return here so we don't repeat https://mfukar.github.io/2014/02/22/Apple-SSL-bug.html

LukasSkywalker (approved):

PR #2072, opened 2025-09-04
refactor(api): make all guids strings on a database level

Itrulia:
As figured out by @dsheyp, when one tries to query an id that is not a guid/uuid, the database throws an error.
This is not desired in my opinion in applications like these as we map urls to ids, this keeps the ids being the same structure but with a more loose database type.

### DIFF
--- a/libs/api/prisma/migrations/20250904071505_make_all_guids_in_the_database_strings/migration.sql
+++ b/libs/api/prisma/migrations/20250904071505_make_all_guids_in_the_database_strings/migration.sql
@@ -1,0 +1,176 @@
+/*
+  Warnings:
+
+  - The primary key for the `banner_actions` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - The primary key for the `banners` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - The primary key for the `block-content.styles` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - The primary key for the `consents` table will be changed. If it partially fails, the table could be left without primary key constraint.
+  - The primary key for the `crowdfunding_goals` table will be changed. If it partially fails, the table could be left without primary key constraint.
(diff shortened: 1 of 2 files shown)


libs/api/prisma/migrations/20250904071505_make_all_guids_in_the_database_strings/migration.sql
```diff
@@ -0,0 +1,176 @@
...
+ALTER TABLE "_CrowdfundingToMemberPlan" ALTER COLUMN "A" SET DATA TYPE TEXT;
+
+-- AlterTable
+ALTER TABLE "_PaymentMethodToSubscriptionFlow" ALTER COLUMN "B" SET DATA TYPE TEXT;
+
+-- AlterTable
+ALTER TABLE "banner_actions" DROP CONSTRAINT "banner_actions_pkey",
+ALTER COLUMN "id" SET DATA TYPE TEXT,
+ALTER COLUMN "bannerId" SET DATA TYPE TEXT,
+ADD CONSTRAINT "banner_actions_pkey" PRIMARY KEY ("id");
```
LukasSkywalker, line 67: UUIDs are pretty heavily normalized when being written to the database. Because the tables using a UUID were added when the database was already migrated to PostgreSQL, this probably doesn't affect us. But it might still be more cautious to use `citext` instead of `text` here. That way, comparisons and the unique constraint keep being case insensitive and the index is used regardless of the casing of the query string.

elias-summermatter: Why not just rewrite existing non conform uuids with:
`UPDATE articles
SET id = gen_random_uuid()
WHERE NOT (
  uuid_column ~* '^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$'
);?
`

elias-summermatter (approved):

PR #2074, opened 2025-09-04
fix(website): ssr auth link not properly working

Itrulia:


libs/membership/website/src/lib/invoice-list/invoice-list-container.tsx
```diff
@@ -27,7 +27,7 @@ export function InvoiceListContainer({filter, className}: InvoiceListContainerPr
   } = useInvoicesQuery({
     onCompleted(data) {
```
Itrulia, line 28: apparently onCompleted callbacks gets called on server side too when the data is there already

libs/utils/website/src/lib/auth-link.tsx
```diff
@@ -1,9 +1,14 @@
 import {ApolloLink, DefaultContext} from '@apollo/client'
 import {AuthTokenStorageKey} from '@wepublish/authentication/website'
 import {getCookie} from 'cookies-next'
+import {Observable} from 'zen-observable-ts'
 
 export const authLink = new ApolloLink((operation, forward) => {
-  const {token} = JSON.parse(getCookie(AuthTokenStorageKey)?.toString() ?? '{}')
+  const {token: cookieToken} = JSON.parse(getCookie(AuthTokenStorageKey)?.toString() ?? '{}')
+  const {token: sessionStorageToken} = JSON.parse(
```
Itrulia, line 8: This was copied over from hauptstadt, hauptstadt has the token in session storage as the cookie is httpOnly. Left it here as it makes no difference.

dsheyp (approved):
- great
-tested together with feat/SHARE-59-404page

PR #2078, opened 2025-09-08
feat: SHARE-60 canonical meta tags

dsheyp:
- seed: added tags for testing
- media-websites: pages with paging: added canonical tag

### DIFF
--- a/libs/tag/website/src/lib/tag.tsx
+++ b/libs/tag/website/src/lib/tag.tsx
@@ -85,10 +85,6 @@ export function Tag({
           <Head>
             <link
               rel="canonical"
-              href={canonicalUrl}
-            />
-            <link
-              rel="canonical"
               key="canonical"
               href={canonicalUrl}
             />


apps/babanews/pages/a/index.tsx
```diff
@@ -49,24 +50,31 @@ export default function ArticleList() {
     return 1
   }, [data?.articles.totalCount])
 
+  const canonicalUrl = '/a'
```
Itrulia, line 53: Canonical urls should be absolute ideally (but I understand why they aren't here due to SSR) but can be relative. So I guess it's fine to keep it like this.

apps/babanews/pages/a/tag/[tag].tsx
```diff
@@ -60,6 +61,10 @@ export default function ArticleListByTag({tagId}: ArticleListByTagProps) {
     return 1
   }, [data?.articles.totalCount])
 
+  const canonicalUrl = useMemo(() => {
```
Comment (Itrulia) [inline apps/babanews/pages/a/tag/[tag].tsx:64]: No need to useMemo for this as not enough CPU cycles are saved (and possibly even more used) + extra memory.

apps/babanews/pages/a/tag/[tag].tsx
```diff
@@ -71,19 +76,24 @@ export default function ArticleListByTag({tagId}: ArticleListByTagProps) {
...
+          <Head>
+            <link rel="canonical" href={canonicalUrl} />
+          </Head>
+          <Pagination
+            page={page ?? 1}
+            count={pageCount}
+            onChange={(_, value) =>
+              replace(
+                {
+                  query: value > 1 ? {...query, page: value} : (delete query.page, {...query})
```
Comment (Itrulia) [inline apps/babanews/pages/a/tag/[tag].tsx:89]: Don't think we actually need or want to remove it from the URL.

apps/babanews/pages/a/index.tsx
```diff
@@ -49,24 +50,31 @@ export default function ArticleList() {
...
-                query: {...query, page: value}
-              },
-              undefined,
-              {shallow: true, scroll: true}
-            )
-          }
-        />
+        <>
+          <Head>
+            <link rel="canonical" href={canonicalUrl} />
```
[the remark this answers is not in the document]
dsheyp: sure

apps/babanews/pages/a/tag/[tag].tsx
Comment (Itrulia) [reply apps/babanews/pages/a/tag/[tag].tsx:88]: Since the task is for SEO I assumed it was for the canonical tag, the URL outside of canonical tag has no influence on SEO.

elias-summermatter:
@dsheyp anything unclear here? :-)

dsheyp:
... everything's clear now.

apps/babanews/pages/a/tag/[tag].tsx
Comment (dsheyp) [reply apps/babanews/pages/a/tag/[tag].tsx]: agreed. URL changes have been reverted

libs/tag/website/src/lib/tag.tsx
```diff
@@ -78,15 +81,29 @@ export function Tag({
...
-          count={pageCount}
-          onChange={(_, value) =>
-            onVariablesChange?.({
-              skip: (value - 1) * take,
-            })
-          }
-        />
+        <>
+          <Head>
+            <link
```
Itrulia: I don't think this is needed/intended twice  here

Itrulia (approved):

PR #2079, opened 2025-09-09
feat: HAS-22

Itrulia:


elias-summermatter:
@Itrulia  or @michael-scheurer can you please merge this branch as fast as possible. I think otherwise we run into merge conflicts all the time... Please try in the future to split pr in multiple small ones that we can merge during the projects into dev branch

Itrulia:
@elias-summermatter a lot of this has been split out already, the problem was that I needed to touch shared components & shared logic where I had no time to make sure it works on the other medias.


And then it grew and grew as other commits relied on previous commits etc.

PR #2089, opened 2025-09-11
fix(core): Remove not working wordpress migration 

elias-summermatter:


elias-summermatter:
@Itrulia can you have a look

PR #2098, opened 2025-09-16
fix(onlinereports): OR-94 prevent escaping slashes in image-block caption

dsheyp:


michael-scheurer (approved):

Itrulia (commented):
{{ caption }} should be {{- caption }} instead imo.

But this in general now allows to inject scripts from the CMS hm...

dsheyp:
@Itrulia you're right about the possibility with that solution to insert anything in caption/source from the cms, including scripts. but wouldn't using  the {{- caption }} notation achieve the same? i think we need to define which special chars should be allowed in caption/source and then write our own custom escape function. what do you think?

Itrulia:
Yeah we have to look into it, basically why I didnt put this in requires changes.


Using {{- instead of escapeValue false allows the translation to decide instead of the code. As theres no technical requirement it to be unescaped.

PR #2112, opened 2025-09-18
fix(onlinereports): OR-96 fix vertical margins and line-heights in mobile view (ads, titles, pre-titles)

dsheyp:
… also addresses horizontal overflow on super small devices < 300px width


dsheyp:
- was deployed to production
- a new PR was created to merge into master

PR #2126, opened 2025-09-24
Extend crowd funding block has 116

Pakicetus97:


libs/utils/website/src/lib/pages/profile/profile-page.tsx
```diff
@@ -81,6 +90,47 @@ function ProfilePage(props: ProfilePageProps) {
     fetchPolicy: 'cache-only'
   })
 
+  const [activeTab, setActiveTab] = useState<SubscriptionTabValue>('subscription')
+
+  const filterActiveSubscriptions = useCallback(
```
Itrulia, line 95: Does not need to be inside the component.

Itrulia (changes requested):

PR #2135, opened 2025-09-25
fix(onlinereports): OR-100 comments: fix login/registration &  forceNonSystemShare for sharing OR comments

dsheyp:


libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -61,27 +56,34 @@ export function LoginForm({
...
   const onSubmit = handleSubmit(({ email, requirePassword, password }) => {
     if (requirePassword) {
+      setShowStandByMessage(true);
       return onSubmitLoginWithCredentials(email, password);
     }
 
     return onSubmitLoginWithEmail(email);
   });
 
+  useEffect(() => {
```
Itrulia, line 83: Should not use an effect for this

libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -145,7 +147,18 @@ export function LoginForm({
...
 
-        {error && <Alert severity="error">{error.message}</Alert>}
+        {showStandByMessage && !error && !loginLinkSent && (
+          <Alert severity="info">
+            Einen Moment bitte, die eingegebenen Anmeldedaten werden
+            überprüft...
+          </Alert>
+        )}
+
+        {error && (
```
Itrulia, line 157: This error message is wrong as it can error for different reasons.

libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -61,27 +56,34 @@ export function LoginForm({
   const {
     elements: { Alert, TextField },
   } = useWebsiteBuilder();
+  const [showStandByMessage, setShowStandByMessage] = useState(false);
```
Itrulia, line 59: this is essentially a loading state, should not be done this way (and can never be set to false again).

Get the login state either via awaiting the onSubmit or the parents apollo state.

libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -122,7 +124,7 @@ export function LoginForm({
               label={'Email'}
               error={!!error}
               helperText={error?.message}
-              ref={autofocus}
+              {...register('email')}
```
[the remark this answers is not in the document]
dsheyp: see previous comment...
sure! will revert to the ref and autofocus() function we previously had and instead use a timeout of 300ms before accessing DOM and setting focus

libs/authentication/website/src/lib/login-form/login-form.tsx
dsheyp: I see. just realized that it didn't do what it was intended for and "fixed" it on the go 😂  will remove

libs/authentication/website/src/lib/login-form/login-form.tsx
dsheyp: reverted

libs/authentication/website/src/lib/login-form/login-form.tsx
dsheyp: you're right. pushed an improved solution

libs/authentication/website/src/lib/login-form/login-form.tsx
Itrulia: I just checked, it works fine for me locally and on https://www.hauptstadt.be/login (the console.log should be gone tho).

Where did it not work for you? Did you test it in storybook (as there's no SSR so this prob won't work)?

libs/authentication/website/src/lib/login-form/login-form.tsx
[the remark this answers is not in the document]
Itrulia: It's more that in general register should not be used together with Controller in react-hook-form: https://react-hook-form.com/get-started#IntegratingwithUIlibraries

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
Itrulia, line 1: What is the goal here? this changes the design and interaction for everyone and set's fairly inconsistent spacings.

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
```diff
@@ -265,112 +328,125 @@ export const CommentEditor = ({
...
+
   return (
     <>
-      <CommentEditorWrapper
-        className={className}
-        onSubmit={submit}
-        modalOpen={modalOpen}
-      >
-        {!hasUser && (
+      {(anonymousCanComment || hasUser) && (
```
Itrulia, line 347: Editor isn't shown unless a user can comment already.

libs/authentication/website/src/lib/login-form/login-form.tsx
dsheyp: for me, it doesn't work for onlinereports, neither locally nor in production. i just checked staging for onlinereports - and there it does work. also just checked tsri in production - and it didn't work. either we have a commit that broke something or this feature has an inconsistent behaviour, not sure

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
dsheyp: you're right, this will affect all media websites. so far, i have only tested it for onlinereports and for OR it does introduce more consistent spacings. i've made some screenshots of the current situation and my new proposal. pasted it on jira - [please have a look](https://wepublish.atlassian.net/browse/OR-100)

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
dsheyp: just checked. same behaviour for gruppetto: after providing non-existing email, modal closes but editor stays open

Itrulia: @dsheyp you just fixed that behaviour already: https://github.com/wepublish/wepublish/pull/2135/files#diff-e2d61fc12c52c0507a9f92cf3fc04648608e034ad4c38e16c0329a24473d55f0L56

dsheyp: my last comment gave me the idea to actually fix the afterLoginCallback, to await and check the result. then we can still have the blurred editor in the background....

dsheyp: pushed a solution where afterLoginCallback is only called after a successful login. the disadvantage of this solution is, that if the user performs a logout while commenting the ui doesn't reflect that change and the user might keep interacting with the still visible editor

libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -70,13 +80,17 @@ export function LoginForm({
...
-    reValidateMode: 'onChange',
+    mode: 'onSubmit',
+    reValidateMode: 'onBlur',
   });
 
   const onSubmit = handleSubmit(({ email, requirePassword, password }) => {
     if (requirePassword) {
-      return onSubmitLoginWithCredentials(email, password);
+      setAwaitingWithCredentials(true);
+      return (async () => {
```
Itrulia, line 90: Look up async action I think its called, comments should have it.

Helps you abstracting this better.

libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -122,7 +136,11 @@ export function LoginForm({
               label={'Email'}
               error={!!error}
               helperText={error?.message}
-              ref={autofocus}
+              ref={
```
Itrulia, line 139: this now gets called every rerender when theres a delay set

libs/authentication/website/src/lib/login-form/login-form.tsx
```diff
@@ -43,9 +44,16 @@ const loginFormSchema = z.union([
   withCredentialsFormSchema,
 ]);
 
-const autofocus = (node: HTMLElement | null) => {
+const autofocus = (node: HTMLElement | null, focusDelay?: number) => {
```
Itrulia, line 47: Adding a delay is probably fixing the symptom not the cause. But fine for now (going live for OR but want to revisit that before merged into dev).

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
```diff
@@ -408,23 +474,41 @@ export const CommentEditor = ({
...
-                afterLoginCallback={handleAfterLoginCallback}
-              />
-
-              <Register>
-                <Button
-                  css={buttonStyles}
-                  onClick={registerRedirect}
+              <ExistingAccountsLoginWrapper>
+                <H1
+                  css={existingAccountsHeadingStyles}
```
Itrulia, line 479: If you want to use the styling of H4, then use H4. 

As discussed, you should add a comoponent to it which controls the html element. H1 is not the same as h1.

Also needs to be gotten from useWebsiteBuilder as it is an overrideable component.

libs/comments/website/src/lib/comment-editor/comment-editor.tsx
```diff
@@ -38,18 +39,90 @@ export const CommentEditorActions = styled('div')`
 export const LoginWrapper = styled('div')`
```
Itrulia, line 39: Theres a lot of moving part for something that should be a fairly simple layout in my opinion.

Should we take a look at it together?

PR #2138, opened 2025-09-26
fix(api): bounded cache in admin api

Itrulia:
image
image

https://www.apollographql.com/docs/apollo-server/v3/performance/cache-backends#ensuring-a-bounded-cache
https://www.apollographql.com/docs/kotlin/advanced/persisted-queries

we don't use the persisted query feature so turned that off.


elias-summermatter:
@Itrulia very nice if tests are fix merge :-)

Itrulia:
@elias-summermatter tests are already broken for a while. I think my hotfix for bajour actually fixes them as it seems like they are about periodic job.

PR #2147, opened 2025-09-30
f/GRUP 21 added max header links

Pakicetus97:
- feat(website): added powered by WePublish logo
- fix(webiste): build error
- feat(website): added first implementation
- feat(website): added default for headerMax and teststories

### DIFF
--- a/libs/navigation/website/src/lib/navbar/navbar.stories.tsx
+++ b/libs/navigation/website/src/lib/navbar/navbar.stories.tsx
@@ -185,38 +185,6 @@ const navigations = [
   },
 ] as Navigation[];
 
-const extendedHeaderLinks = [
-  {
-    __typename: 'PageNavigationLink',
-    label: 'Mitglied werden',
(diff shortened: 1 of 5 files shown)


Itrulia (changes requested):
Can we do this with CSS instead? CSS already gives us this option so we don't have to reimplement this.


This would be as simple as:


```typescript
const GruppettoNavbar = styled(NavbarContainer)`
  ${NavbarLink}:nth-child(n+3) {
    display: none;
  }
`
```

libs/navigation/website/src/lib/navbar/navbar.stories.tsx
```diff
@@ -206,6 +238,20 @@ export const Default = {
   },
 };
 
+export const WithExtendedHeaderLinks = {
```
Itrulia, line 241: dont think changes to this file are needed anymore

Itrulia (changes requested):

Itrulia (approved):

PR #2152, opened 2025-10-01
f/HAS-116-subscribe-block

Itrulia:


dsheyp:
i'm currently reviewing this PR. i noticed that once a user creates a donation, in their profile it will be handled as a subscription (wording: "Aktive Abos", "Abo ist unbezahlt", "Abo kündigen", "Anderes Abo lösen.") will it (for now?) be handled like this, without a separate donation-section in the users profile?

dsheyp:
i noticed: when creating a donation as an admin user in the editor, under "Advanced Settings" some settings for subscriptions are showing (like: "Configure for trial subscription", "Migration of subscriptions...", "Extendable") I guess for donations "Recurring" (if users should be able to choose to make their donation a recurring one) would be the wording instead of "Extendable"

dsheyp (approved):

PR #2165, opened 2025-10-06
feat(onlinereports): OR-104 implemented social icons in custom OR footer

dsheyp:


apps/onlinereports/src/components/footer.tsx
```diff
@@ -24,19 +57,170 @@ const OnlineReportsFooterContainer = styled(FooterContainer)`
...
-      wepublishLogo="hidden"
+    <FooterWrapper
+      className={className}
+      ref={ref}
     >
+      <FooterPaper
+        main={mainItems}
+        categories={[]}
+        children={children} // eslint-disable-line react/no-children-prop
+        style={
```
Itrulia, line 147: dont use inline styles please. Either extract them using styled or use css/sxnif available.

apps/onlinereports/src/components/footer.tsx
```diff
@@ -24,19 +57,170 @@ const OnlineReportsFooterContainer = styled(FooterContainer)`
...
-      categorySlugs={[['footer']]}
-      wepublishLogo="hidden"
+    <FooterWrapper
+      className={className}
+      ref={ref}
     >
+      <FooterPaper
+        main={mainItems}
+        categories={[]}
+        children={children} // eslint-disable-line react/no-children-prop
```
Itrulia, line 146: there for a reason, use transclusion for children

apps/onlinereports/src/components/footer.tsx
```diff
@@ -24,19 +57,170 @@ const OnlineReportsFooterContainer = styled(FooterContainer)`
...
+        style={
+          !!iconItems?.links.length || wepublishLogo !== 'hidden' ?
+            { paddingBottom: 0 }
+          : {}
+        }
+      />
+
+      {(!!iconItems?.links.length || wepublishLogo !== 'hidden') && (
```
Itrulia, line 154: FYI online reports will always have a hidden wep logo.

Itrulia (changes requested):

apps/onlinereports/src/components/footer.tsx
dsheyp: ok, thx 👍 i'll keep the wep logo code in the custom OR-footer so if this would change one day it would already be in place....

PR #2201, opened 2025-10-16
feature(hauptstadt): added shadow to navigation bar

Pakicetus97:


dsheyp (changes requested):
I have two points, as discussed:

1.
- I think the drop-shadow should only be visible if navbarState includes `navbarState.Diagonal` --> e.g. if the line is visible and in a diagonal position
--> pass navbarState as an argument/prop into `NavBackgroundWrapper` (the same way it is implemented for `NavbarInnerWrapper`) 

2.
- observing  the width of `NavBackgroundWrapper` in the webdeveloper toolbar shows that the width of `NavBackgroundWrapper` only changes around the moment a breakpoint is reached
- looking at it more closely reveals that the widths at which it's changing are defined in the theme as styleOverrides [maxWidth](https://github.com/wepublish/wepublish/blob/7298a55c4daf2240a25a968c8a4bf9ddb1adb835/apps/hauptstadt/src/theme.tsx) for MuiToolbar (line 381 ff)
- using a ResizeObserver plus a resize-eventListener in JS seems a bit too much and it would be preferable to go with a "less intense" solution, if possible
- the beforementioned maxWidth-values can be used to give a css-max-width-property to the `NavBackgroundWrapper`, which will rely on media-queries
- the `NavBackgroundWrapper` can then be centered using margin: 0 auto;
- the css clip-path property then becomes static and doesn't change with different screen sizes
- this allows to implement the shadow as a "fully static" solution, solely relying on media-queries, with no need to dynamically calculate width on resize

PR #2424, opened 2026-02-22
feat(core): Add Chromatic

elias-summermatter:


libs/storybook/mocks/src/index.ts
```diff
@@ -1,3 +1,10 @@
+import { faker } from '@faker-js/faker';
+
+// Fixed seed + reference date so all faker-generated values are identical
+// across every Chromatic build, regardless of module load order.
+faker.seed(123);
```
Itrulia: Could we check if it's chromatic via ENV variables? I would love to have different seeds in non test environments so that we can see different outputs.

libs/navigation/website/src/lib/navbar/navbar.tsx
```diff
@@ -266,7 +266,7 @@ export function Navbar({
 
   const categories = useMemo(
     () =>
-      categorySlugs.map(categorySlugArray =>
+      (categorySlugs ?? []).map(categorySlugArray =>
```
Itrulia: Fix the mocks instead as categorySlugs can not be null/undefined according to the types

PR #2595, opened 2026-04-01
fix(search) fix search results pagination showing too many pages

dsheyp:
* avoid showing empty search result pages

### DIFF
--- a/libs/page/api/src/lib/page.service.ts
+++ b/libs/page/api/src/lib/page.service.ts
@@ -389,7 +389,7 @@ export class PageService {
   async performPageFullTextSearch(searchQuery: string): Promise<string[]> {
     try {
       const foundPageIds = await this.prisma.$queryRaw<Array<{ id: string }>>`
-        SELECT p.id
+        SELECT DISTINCT a.id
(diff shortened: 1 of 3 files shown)


LukasSkywalker (approved):

libs/page/api/src/lib/page.service.ts
```diff
@@ -389,7 +389,7 @@ export class PageService {
   async performPageFullTextSearch(searchQuery: string): Promise<string[]> {
     try {
       const foundPageIds = await this.prisma.$queryRaw<Array<{ id: string }>>`
-        SELECT p.id
+        SELECT DISTINCT a.id
```
LukasSkywalker: Hie

LukasSkywalker: @dsheyp

LukasSkywalker (approved):

LukasSkywalker (approved):

libs/page/api/src/lib/page.service.ts
dsheyp: thanks!

PR #2611, opened 2026-04-10
feat(share): improve slider customization

dsheyp:

PR #2612, opened 2026-04-10
feat(tsri): redesign image slider TSRI-187

dsheyp:
* depends on https://github.com/wepublish/wepublish/pull/2611


apps/tsri/src/components/tsri-image-slider.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+const SwipeHintBase = ({
+  className,
+  onAnimationEnd,
+}: {
+  state: 'showing' | 'fading';
+  className?: string;
+  onAnimationEnd?: () => void;
+}) => (
+  <div
```
Itrulia: Do we need a div around it? Couldn't we do it directly with the svg?

Itrulia (approved):

apps/tsri/src/components/tsri-image-slider.tsx
dsheyp: no, div isn't needed - removed

PR #2625, opened 2026-04-15
fix(tsri, hauptstadt, website): updated at showing even when just a few seconds apart

Itrulia:


apps/hauptstadt/src/components/hauptstadt-article-date.tsx
```diff
@@ -16,14 +17,28 @@ export const HauptstadtArticleDate = ({
 
   const updated =
     !!article.latest.publishedAt &&
-    article.latest.publishedAt !== article.publishedAt;
+    !!differenceInDays(
```
Itrulia, line 20: 0 = same day, goes into negative and positive

dsheyp (approved):

PR #2632, opened 2026-04-16
feat(core) on pod creation instead of exit log a warning in case of missing MAILCHIMP_API_KEY

dsheyp:


deployment/map-secrets.sh
```diff
@@ -18,13 +18,13 @@ if [[ $MODE == "restore" ]]; then
   if [[ -f secrets_name.list ]]; then
     for n in $(cat secrets_name.list); do
 
-      # Exit if en env is missing!
       if [[ -z ${!n} ]]; then
-        echo "Secret env ${n} not set!"
-        exit 99
+        echo "WARNING: Secret env ${n} not set! Substituting empty string."
+        sed -i "s/\"@@@${n}@@@\"/\"\"/g" $(grep -rl "\"@@@${n}@@@\"" /wepublish)
```
elias-summermatter, line 23: Is this maybe same for other envs for other media? Did you check?

dsheyp: it seems MAILCHIMP_API_KEY is the only one using `secret_env`, while all other values (like GA_ID) are using plain `env`

the `secret_env` is being used by cultur, wnti, tsri

dsheyp: review03 works well with bajour. both back- & frontend were deployed with no errors --> root cause for pod-exit seems to be confirmed (missing MAILCHIMP_API_KEY) --> my change should fix this (the frontend should gracefully handle the missing MAILCHIMP_API_KEY env var. the error occurs during pod deployment

elias-summermatter (changes requested):
This is not a good idea i think....

PR #2661, opened 2026-04-22
feat(core): Add mailersend as mailprovider

elias-summermatter:


libs/mail/api/src/lib/mail-provider/mailersend-mail-provider.ts
```diff
@@ -0,0 +1,214 @@
...
+    req,
+  }: WebhookForSendMailProps): Promise<MailLogStatus[]> {
+    if (req.method !== 'POST') {
+      return [];
+    }
+
+    const signature = req.headers['signature'];
+    const sigValue = Array.isArray(signature) ? signature[0] : signature;
+
+    // MailerSend signs the raw request body. Express parses JSON by default
```
Itrulia, line 107: This is handled by the `incomingRequestHandler` in the `MailWebhookMiddleware` and not parsed by default. You can set that per provider

libs/mail/api/src/lib/mail-provider/mailersend-mail-provider.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    const eventType: string | undefined = req.body?.type;
+    const state =
+      eventType ? mapMailersendEventToMailLogState(eventType) : null;
+
+    const mailLogID = extractMailLogId(req.body?.data?.email?.tags);
+
+    const mailLogStatuses: MailLogStatus[] = [];
+    if (state !== null && mailLogID !== undefined) {
```
Itrulia, line 127:
```suggestion
    if (state != null && mailLogID != null) {
```

Itrulia (changes requested):

PR #2668, opened 2026-04-28
fix(man): do not load hot and trending data

michael-scheurer:
this pr is meant as a test only. it makes sure hot and trending data are not loaded from the api.


michael-scheurer:
@elias-summermatter do you still use this pr? it was meant for testing at mannschaft. shall I close it?

PR #2669, opened 2026-04-28
feat(core): Add wepublish infra configurator

elias-summermatter:


apps/api-example/src/nestapp/app.module.ts
```diff
@@ -583,6 +588,14 @@ import {
       },
       inject: [ConfigService],
     },
+    {
+      provide: 'CONFIGURATOR_BOOTSTRAP',
```
Itrulia, line 592: Can we use symbols instead of magic strings?

PR #2677, opened 2026-05-01
feat: allow disabling blocks so that they dont show up

Itrulia:
Also fixes constant telling to the user that there has been unsaved changes


apps/api-example/prisma/seed.ts
Itrulia, line 1: It's annoying to have to re-setup 2fa locally all the time (this is just for local)

libs/ui/editor/src/lib/atoms/blockList.tsx
```diff
@@ -339,16 +347,21 @@ interface ListItemWrapperProps {
   disabled?: boolean;
 
   onDelete?: () => void;
+  onDisable?: (disable: boolean) => void;
   onMoveUp?: () => void;
   onMoveDown?: () => void;
-  onStyleChange?: (blockStyle?: BlockStyle['id']) => void;
+  onStyleChange?: (
```
Itrulia, line 353: Changes to this are related to constantly being asked to save as there's been unsaved changes. This fixes it

libs/website/api/src/lib/omit-disabled-blocks.ts
Itrulia, line 1: Removes the disabled blocks so they never even hit the frontend apart from the API call

dsheyp (approved):

PR #2681, opened 2026-05-05
feat: banner add hide and hide for options

Itrulia:


apps/editor/src/app/locales/fr.json
```diff
@@ -199,6 +199,66 @@
         "tags": "Tags"
       }
     },
+    "banner": {
```
Itrulia, line 203: @dsheyp @michael-scheurer AI generated, please double check for me 🇫🇷

apps/hauptstadt/src/components/hauptstadt-banner.tsx
Itrulia, line 1: Now handled via the settings

libs/api/prisma/migrations/20260505191446_add_collapsible_and_hide_for_minutes_properties_to_banner/migration.sql
```diff
@@ -0,0 +1,21 @@
+-- AlterTable
+ALTER TABLE "_BannerToPage" ADD CONSTRAINT "_BannerToPage_AB_pkey" PRIMARY KEY ("A", "B");
```
Itrulia, line 2: These are regenerated by Prisma. Prisma handles these differently in v7 or 6

prisma.config.ts
```diff
@@ -1,24 +1,18 @@
...
-if (fs.existsSync(envPath)) {
-  for (const line of fs.readFileSync(envPath, 'utf8').split('\n')) {
-    const match = line.match(/^\s*([\w.-]+)\s*=\s*(.*)?\s*$/);
-    if (match && !process.env[match[1]]) {
-      process.env[match[1]] = match[2]?.replace(/^['"]|['"]$/g, '') ?? '';
-    }
-  }
-}
+config({
+  path: ['.env'],
```
Itrulia, line 7: only .env so that you don't run migrations from .env.local by accident (saved my ass a few times that old prisma behaved like this).

apps/editor/src/app/locales/fr.json
dsheyp: made some minor changes and re-pushed fr.json

dsheyp (approved):

apps/editor/src/app/locales/fr.json
Itrulia: Thank you!

PR #2683, opened 2026-05-06
feat: focalPoint on image, properties as JSONB

Itrulia:
Improve database queries by moving focalPoint and properties directly onto the table


libs/user/api/src/lib/user.service.ts
```diff
@@ -143,11 +142,7 @@ export class UserService {
         ...input,
         active: true,
         password: hashedPassword,
-        properties: {
-          createMany: {
-            data: properties ?? [],
-          },
-        },
+        properties: properties as any,
```
Itrulia, line 145: Unfortunately JsonValue isn't compatible with JsonInputValue by Prisma...

libs/payment/api/src/lib/payment-provider/payrexx-subscription-payment-provider.ts
```diff
@@ -459,14 +454,15 @@ export class PayrexxSubscriptionPaymentProvider extends BasePaymentProvider {
     return this.prisma.subscription.findFirst({
       where: {
         properties: {
-          some: {
-            key: 'payrexx_external_id',
```
Itrulia, line 463: Couldn't even find where this is ever set...

LukasSkywalker (approved):

PR #2685, opened 2026-05-07
feat(fazetten): use the wepublish-site logo-wall for fazetten

dsheyp:


apps/fazetten/src/components/fazetten-article.tsx
```diff
@@ -1,6 +1,14 @@
+import styled from '@emotion/styled';
 import { Article } from '@wepublish/article/website';
 import { createWithTheme } from '@wepublish/ui';
 
 import { ContentTheme } from '../theme';
+import { StyledTeaserSlotsLogoWall } from './teaser-layouts/teaser-slots-logo-wall';
 
-export const FazettenArticle = createWithTheme(Article, ContentTheme);
+const StyledArticle = styled(Article)`
```
Itrulia, line 8: Bruchts glaub nümm

apps/fazetten/src/mui-augmentation.d.ts
Itrulia, line 1: Could we move this to a shared place? With interface merging we don't have to worry when we override it again I imagine.

Itrulia (approved):

apps/fazetten/src/components/fazetten-article.tsx
dsheyp: will leave it here as a stub - in case the LogoWall-design will have to be adjusted when the block is rendered in articles instead of on pages (stuff like "how many cols does it occupy, show header --> yes/no, header-typography-styles, ....)

apps/fazetten/src/mui-augmentation.d.ts
dsheyp: for my understanding - is this what you mean:
* now, that several apps have their own mui-augmentation.d.ts file --> move all definitions shared by the apps to libs/ui/theme.tsx
* only leave the definitions that are app-specific in each apps mui-augmentations.d.ts

PR #2693, opened 2026-05-08
perf(article): drop in-memory tag pre-pass, rely on SQL relation filter

dsheyp:


LukasSkywalker:
https://github.com/wepublish/wepublish/pull/2835

PR #2694, opened 2026-05-08
perf(article): skip count(*) when not explicitly required, cache count and rows for 60s

dsheyp:


libs/article/api/src/lib/article.resolver.ts
```diff
@@ -97,7 +99,8 @@ export class ArticleResolver {
   })
   public articles(
     @Args() args: ArticleListArgs,
-    @PreviewMode() isPreview: boolean
+    @PreviewMode() isPreview: boolean,
+    @Info() info: GraphQLResolveInfo
```
Itrulia, line 103: Could use https://www.npmjs.com/package/graphql-fields-list instead and make a proper decorator for it

PR #2696, opened 2026-05-08
perf(website): bypass next/link to stop prefetch storm

dsheyp:


libs/utils/website/src/lib/next-wepublish-link.tsx
Itrulia, line 1: This is not just an a11y issue, this is fixing a symptom with insane side effects.
This completely negates the performance improvements for the user.

Itrulia (changes requested):

Itrulia:
superseeded by https://github.com/wepublish/wepublish/pull/2706

PR #2699, opened 2026-05-08
fix(website): use public API URL and 60s SSR timeout during deployment-next-build

dsheyp:


Itrulia:
Discussed in meeting as not needed, @elias-summermatter increases timeout on build instead

PR #2704, opened 2026-05-11
fix(cultur, tsri): mailchimp potentially breaking pages

Itrulia:
And improved logging


apps/wnti/src/components/tsri-teaser.tsx
Itrulia, line 1: Never had the fetching for it done, so wouldn't have worked. IIRC it's because wnti was a copy of tsri

dsheyp (approved):

PR #2716, opened 2026-05-14
feat: allow to switch between payment amount picker and slider

michael-scheurer:


elias-summermatter:
@michael-scheurer please fix as discuss with Karl

PR #2725, opened 2026-05-18
feat(editor-integrations): frontend tracking

dsheyp:


Itrulia:
Was this prompted from somewhere (like a jira ticket)? This is something I've been working on (but currently on hold due to other stuff).
The implementation here introduces a few new things we have not planned to do like this and has the issue that it's not properly loading the data at the correct time

This goes hand in hand with the self service for Themes etc.

PR #2743, opened 2026-05-26
fix(api): make check intent state a hard fail

Itrulia:
Without this, we will book money even when we don't know if it has been paid already or not

### DIFF
--- a/libs/membership/api/src/lib/periodic-job/subscription.service.ts
+++ b/libs/membership/api/src/lib/periodic-job/subscription.service.ts
@@ -455,23 +455,29 @@ export class SubscriptionService {
     }
     const payments = await this.payments.findByInvoiceId(invoice.id);
     for (const payment of payments) {
-      if (!payment || !payment.intentID) continue;
+      if (!payment || !payment.intentID) {
+        continue;
+      }
+
       try {
         const intentState = await paymentProvider.checkIntentStatus({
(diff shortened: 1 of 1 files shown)


LukasSkywalker (approved):

libs/membership/api/src/lib/periodic-job/subscription.service.ts
```diff
@@ -455,23 +455,29 @@ export class SubscriptionService {
...
         logger('checkInvoiceState').error(
-          'Checking payment <%s> with intent %s on payment provider %s soft failed with error: %s',
+          'Checking payment <%s> with intent %s on payment provider %s hard failed with error: %s',
           payment.id,
           payment.intentID,
           paymentProvider.getName(),
           e
         );
+
+        throw e;
```
michael-scheurer, line 480: How about expired credit cards where charging the payment mean wasn't pissible. Is this also going to be a hard fail? Because in these cases it must try again every night until subscription is cancelled. Please check on this. This is very important.
@LukasSkywalker @elias-summermatter 
@Itrulia

Itrulia: Thats not what the intent check does. The intent check checks on the payment provider if it has been paid or not.

This essentially charged the user again if the intent check failed

PR #2759, opened 2026-05-30
feat: add vouchers

Itrulia:


libs/membership/website/src/lib/subscribe/subscribe.tsx
```diff
@@ -385,12 +444,17 @@ export const Subscribe = <T extends Exclude<BuilderUserFormFields, 'flair'>>({
   });
 
   const onSubmit = handleSubmit(data => {
+    if (subscribeInfo.data?.createSubscriptionInfo.voucherValid === false) {
```
Itrulia, line 447: If I add it as a validator via .refine, it doesn't update the validation until blurring the input. If I add it via `root.XXX` pressing the submit button removes the validation.

This way it shows the error and blocks it from doing anything.

libs/ui/editor/src/lib/atoms/memberPlan/selectMemberPlans.tsx
Itrulia, line 1: Renamed selectMemberPlan to selectMemberPlans to reflect it selecting multiple

dsheyp (approved):

apps/editor/src/app/locales/de.json
```diff
@@ -1839,6 +1839,32 @@
         "anyMemberPlan": "Alle Abo-Pläne"
       }
     },
+    "voucher": {
+      "navbar": "Gutschein",
```
Itrulia:
```suggestion
      "navbar": "Gutscheine",
```

PR #2765, opened 2026-06-02
fix(api): readd unique email validation

Itrulia:


libs/user/api/src/lib/user.service.ts
```diff
@@ -131,13 +131,22 @@ export class UserService {
...
     input.email = input.email.toLowerCase();
     await Validator.createUser.parse(input);
     await Validator.createAddress.parse(address);
 
+    if (
+      await this.prisma.user.findUnique({
+        where: { email: input.email },
+      })
+    ) {
+      throw new BadRequestException('Email already in use');
```
Itrulia, line 147: Brings back the correct error message

LukasSkywalker (approved):

PR #2768, opened 2026-06-02
feat(editor): crud mailtemplates within the editor

michael-scheurer:
- Create, edit and delete mail templates under /mailtemplates (pushed to the provider; "Auf … ansehen" still works).
- Visual editor (toolbar + desktop/tablet/mobile preview) and HTML editor (CodeMirror with highlighting), switchable without losing content.
- Editable subject line.
- Placeholder sidebar — searchable, click to insert; the full list (incl. optional_*) is now generated from the actual mail-sending logic.
- fix: Template CRUD for Mandrill and Mailgun; "unused" status now also counts system-mail usage.
- fix: Renamed the e-mail integration Mailchimp → Mandrill (+ logo); new create/update/delete permissions.
- Adds @uiw/react-codemirror + @codemirror/lang-html for the html mailtemplate editor.

image


Itrulia (changes requested):
My issue with this is mostly that we made our own rich text editor here that is imo massive tech debt instead of waiting for tiptap (or pushing to get tiptap done).

Ideally this would also use Puck and React-Email as we discussed a few times before

Itrulia (approved):

PR #2771, opened 2026-06-02
feat(editor, website): version history

michael-scheurer:
image


michael-scheurer:
Now with pagination and lightweight revision query instead of full-blown revisions with blocks etc.
Also prepared for future filtering.
Added the user information too.

@Itrulia I was considering to create a dedicated model and service for article reviews and page reviews. I decided to keep it in the existing articles and pages. wdyt?

image

Itrulia:
Should still be the base revision model. All good

Itrulia (approved):

PR #2781, opened 2026-06-04
fix(website): typings

Itrulia:


libs/website/api/src/lib/schemas/article.graphql
```diff
@@ -83,6 +90,11 @@ fragment SlimArticle on Article {
     ...SlimArticleRevision
   }
 
+  paywall {
```
Itrulia, line 93: Needed for hauptstadt teasers

dsheyp (commented):
looks good to me.

PR #2784, opened 2026-06-05
feat(website-builder): add a siblings prop to each block

dsheyp:


Itrulia (commented):
Whats the idea behind this?

Blocks *shouldnt* know about their siblings. If they are in a flex block, the flex block should handle the layout between them.

dsheyp:
> Whats the idea behind this?
> 
> Blocks _shouldnt_ know about their siblings. If they are in a flex block, the flex block should handle the layout between them.

* we had spoken about the siblings stuff before and it was your idea to make it globally available
* but it's nothing new and it is already in production --> in this pr i ended up mixing two things in one pr - that's why it looks like the two things would be somehow related, but they aren't
* let's briefly talk about in a call

PR #2789, opened 2026-06-05
feat(tsri): TSRI-189 adjust styles for promo articles

dsheyp:


apps/tsri/src/hooks/use-advertisers.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return [
+    {
+      type: BlockFormat.Paragraph,
+      children: [{ text: bioText }],
+    },
+  ];
+}
+
+export const useAdvertisers = (
```
Itrulia, line 33: Could probably just be a function and not a hook (doesn't matter here though) especially with react compiler coming at some point in this codebase :D

apps/tsri/src/components/tsri-article-author.tsx
```diff
@@ -20,6 +24,11 @@ export const TsriArticleAuthor = styled(ArticleAuthor)`
     ${({ theme }) => theme.breakpoints.up('md')} {
       width: 140px;
     }
+
+    img {
```
Itrulia: Can we use `ImageWrapper` here? I generally don't like targeting html elements because you don't get any compile time help and you might put an element above.

dsheyp: solved - made img a styled component

apps/tsri/src/hooks/use-advertisers.tsx
dsheyp: agreed - but will leave it as is

Itrulia (approved):

apps/tsri/src/hooks/use-advertisers.tsx
Itrulia: For sure!

PR #2806, opened 2026-06-11
fix(editor): website analytics settings avoid stale cached data and falsely activated analytics providers

dsheyp:
* as soon as plausible is checked active --> piwik pro will be too --> and if key empty will error


Itrulia:
fixxed by https://github.com/wepublish/wepublish/pull/2808

PR #2810, opened 2026-06-12
fix(api): remove committed dev JWT keys

CMZCoder:
## Summary
- remove the tracked local Ed25519 JWT private/public key pair from `.env` and `docker-compose.yml`
- add a shared JWT key resolver for the API example that keeps production strict while generating a process-local Ed25519 pair for local development/test when both env vars are omitted
- use the resolver in mail login JWTs, sessions, media adapter wiring, and API example seeding

## Validation
- `npm ci`
- `git diff --check`
- `npx nx run api-example:test --runInBand --skip-nx-cache`
- `npx nx run api-example:lint --skip-nx-cache` (passes with existing warnings)
- `npx nx run api-example:build --skip-nx-cache`
- `trufflehog git file:///home/void/ExternalRepos/wepublish --since-commit 8be726e397876aa158002bcad559fbe60919876c --branch pr/dev-jwt-key-fallback --results verified --fail --fail-on-scan-errors --no-update --concurrency=2 --json` (0 verified, 0 unverified)
- `trivy fs --scanners secret --skip-dirs node_modules --skip-dirs dist --skip-dirs .git --skip-dirs .nx --severity HIGH,CRITICAL --exit-code 1 .`

## Notes
This is intentionally narrow: it only removes committed development JWT key material and preserves local developer ergonomics without relaxing production requirements.

### DIFF
--- a/.env
+++ b/.env
@@ -8,9 +8,11 @@ API_URL=http://localhost:4000
 #API_URL_INTERNAL=
 WEP_ONE_URL=https://one-admin.wepublish.cloud
 
-# Api (Ed25519 key pair - newlines escaped as \n)
-JWT_PRIVATE_KEY="-----BEGIN PRIVATE KEY-----\nMC4CAQAwBQYDK2VwBCIEIJWbdZ23hpurtFiEpnulSfJ1U4TwbkiDWmbGZSDVx3nR\n-----END PRIVATE KEY-----"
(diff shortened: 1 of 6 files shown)


LukasSkywalker:
@CMZCoder Why that change? The code worked with the JWT, and didn't require additional code.

Itrulia:
Those are dev only keys, there's no need to complicate this.

PR #2811, opened 2026-06-12
fix(media): harden remote URL checks

CMZCoder:
## Summary
- validate media remote-object checks before `fetch`: HTTP(S) only, production private literal hosts blocked, production DNS resolutions to private ranges blocked
- disable automatic redirect following for the HEAD reachability check
- stop logging full object URLs, keeping only origin plus status/error context
- replace disabled media CSP with a restrictive media-safe Helmet policy while preserving cross-origin embedding of media responses

## Validation
- `git diff --check`
- `npx nx run media:test --runInBand --skip-nx-cache`
- `npx nx run media:lint --skip-nx-cache` (passes with existing controller warnings only)
- `npx nx run media:build --skip-nx-cache`
- `trufflehog git file:///home/void/ExternalRepos/wepublish --since-commit 8be726e397876aa158002bcad559fbe60919876c --branch pr/media-remote-url-hardening --results verified --fail --fail-on-scan-errors --no-update --concurrency=2 --json` (0 verified, 0 unverified)

## Notes
This is intentionally limited to the media service security hardening. It does not include dependency bumps or broader package-lock changes from the previous broad branch.

### DIFF
--- a/apps/media/src/app/assertRemoteFileIsAccessible.spec.ts
+++ b/apps/media/src/app/assertRemoteFileIsAccessible.spec.ts
@@ -1,0 +1,103 @@
+import { promises as dns } from 'dns';
+import { assertRemoteFileIsAccessible } from './assertRemoteFileIsAccessible';
+
+describe('assertRemoteFileIsAccessible', () => {
+  const originalFetch = global.fetch;
+  const originalNodeEnv = process.env.NODE_ENV;
+
+  afterEach(() => {
+    global.fetch = originalFetch;
+    process.env.NODE_ENV = originalNodeEnv;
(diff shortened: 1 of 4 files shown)


LukasSkywalker:
@CMZCoder Why would we not want to log full object urls?

dsheyp:
> @CMZCoder Why would we not want to log full object urls?

@CMZCoder @LukasSkywalker to save disk space?

this one https://media.example.com/image.jpg gives me a 404 - could you please re-up it so that we can test?
thank you.

Itrulia:
We control the url, there's no need to check anything.
The CSP change is not doing anything as we are just serving images.

PR #2812, opened 2026-06-12
fix(deployment): harden production deploy script

CMZCoder:
## Summary
- replace shell-interpolated deployment git commands with `execFile("git", args)`
- validate project names before using them in paths, tag patterns, or git arguments
- filter deployment changelog entries in JavaScript instead of piping through shell `egrep`
- make the deploy script import-safe and add node:test coverage for helper behavior

## Validation
- `git diff --check`
- `node --test deployment/deploy_production.test.js`
- `node --check deployment/deploy_production.js`
- `npx prettier --check deployment/deploy_production.js deployment/deploy_production.test.js`
- `npx eslint deployment/deploy_production.js deployment/deploy_production.test.js` (files are ignored by current ESLint config; no source signal)
- `trufflehog git file:///home/void/ExternalRepos/wepublish --since-commit 8be726e397876aa158002bcad559fbe60919876c --branch pr/deploy-script-hardening --results verified --fail --fail-on-scan-errors --no-update --concurrency=2 --json` (0 verified, 0 unverified)

## Notes
This keeps the existing deployment flow intact: dirty-worktree guard, changelog preview, exact project-name confirmation, deploy tag creation, and tag push. The change is limited to command safety and testability.

### DIFF
--- a/deployment/deploy_production.js
+++ b/deployment/deploy_production.js
@@ -1,80 +1,52 @@
-const {exec} = require('child_process');
+const { execFile } = require('child_process');
 const fs = require('fs');
 const path = require('path');
 const readline = require('readline');
-const chalk = require("chalk");
+const chalk = require('chalk');
 
-if (!process.argv[2]) {
-  console.error('Error: Project name is required.');
-  process.exit(1);
+const repositoryRoot = path.resolve(__dirname, '..');
+
+function validateProjectName(projectName) {
+  if (
+    typeof projectName !== 'string' ||
+    !/^[a-zA-Z0-9][a-zA-Z0-9_-]*$/.test(projectName)
+  ) {
+    throw new Error('Invalid project name.');
+  }
+
+  return projectName;
 }
 
-const projectName = process.argv[2];
-
-const rootPath = path.dirname(require.main.filename);
-const configPath = path.resolve(rootPath, '../apps', projectName, 'deployment.config.json');
-
-if (!fs.existsSync(configPath)) {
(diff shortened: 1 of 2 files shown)


deployment/deploy_production.js
```diff
@@ -1,108 +1,209 @@
...
-  console.error('Error: Project name is required.');
-  process.exit(1);
-}
-
-const projectName = process.argv[2];
+const repositoryRoot = path.resolve(__dirname, '..');
 
-const rootPath = path.dirname(require.main.filename);
-const configPath = path.resolve(rootPath, '../apps', projectName, 'deployment.config.json');
+function validateProjectName(projectName) {
```
Itrulia, line 9: What is the value of validating the project name here?

deployment/deploy_production.js
```diff
@@ -1,108 +1,209 @@
...
       }
-    });
+    );
   });
 }
 
-async function deploy() {
+async function deploy(projectName) {
+  const status = await checkUncommittedChanges();
+  if (status.trim() !== '') {
```
Itrulia, line 143:
```suggestion
  if (status.trim()) {
```

deployment/deploy_production.test.js
Itrulia, line 1: Is this being ran somewhere? If not, I don't see the value in having a test.
It's also not using the same testing framework that we normally use.

CMZCoder:
Withdrawing this. After re-checking the script and the review comments, this PR over-scopes a small internal deployment helper and adds test structure outside the repo convention without enough upstream value. The `execFile` idea is not enough by itself to justify the churn here. Sorry for the review noise.

PR #2815, opened 2026-06-12
fix(image): default remote fetch MIME type

CMZCoder:
## Summary
- normalize the remote response `content-type` before handing fetched images to the media adapter
- keep valid string MIME types, support array-shaped header values, and fall back to `application/octet-stream` for missing or unusable values
- add a focused `ImageFetcherService` regression test for valid and invalid content-type headers

## Validation
- red/green: `npx jest --config libs/image/api/jest.config.ts --runInBand libs/image/api/src/lib/image-fetcher/image-fetcher.service.spec.ts` failed before the service change because numeric `content-type` was passed through as `mimetype`; the same command passes after the fix
- `npx eslint libs/image/api/src/lib/image-fetcher/image-fetcher.service.ts libs/image/api/src/lib/image-fetcher/image-fetcher.service.spec.ts`
- `npx prettier --check libs/image/api/src/lib/image-fetcher/image-fetcher.service.ts libs/image/api/src/lib/image-fetcher/image-fetcher.service.spec.ts`
- `git diff --check`
- `trufflehog git ... --since-commit $(git merge-base HEAD origin/master) --branch pr/image-fetcher-mimetype-fallback --results verified --fail --fail-on-scan-errors --no-update --concurrency=2` (0 verified, 0 unverified findings)

## Note
`libs/image/api` is intentionally listed in `.nxignore` because the current graph has an image/peer cycle, so I validated this slice with direct Jest and ESLint commands instead of `nx run image-api:*`.

### DIFF
--- a/libs/image/api/src/lib/image-fetcher/image-fetcher.service.spec.ts
+++ b/libs/image/api/src/lib/image-fetcher/image-fetcher.service.spec.ts
@@ -1,0 +1,40 @@
+import { HttpService } from '@nestjs/axios';
+import { of } from 'rxjs';
+
+import { ImageFetcherService } from './image-fetcher.service';
+
+describe('ImageFetcherService', () => {
+  const buildService = (headers: Record<string, unknown>) => {
+    const get = jest.fn().mockReturnValue(
+      of({
+        data: new ArrayBuffer(8),
+        headers,
+      })
+    );
+
+    return {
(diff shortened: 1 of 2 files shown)


Itrulia:
We only use the mimetype to save as metadata to s3 and map that back to a file extension.

PR #2818, opened 2026-06-14
feat: replace richtext editor with tiptap

Itrulia:


libs/richtext/api/src/lib/slate-to-pm-migrator.ts
Itrulia, line 1: Migration file

libs/api/prisma/migrations/20260420120001_migrate_slate_to_prosemirror/migration.sql
Itrulia, line 1: Rename old columns, make new columns and rename json keys

apps/onlinereports/src/render-element.tsx
Itrulia, line 1: @dsheyp Please double check, dunno what we should do with the new h1-3

apps/tsri/src/components/break-blocks/tsri-sidebar-content.tsx
```diff
@@ -184,6 +183,11 @@ export const TsriSidebarContent = ({
     return undefined;
   }, [index, count, siblings]);
 
+  const nodes = richText?.content ?? [];
+  const firstNode = nodes[0];
+  const startsWithHeadingTwo =
```
Itrulia, line 188: @dsheyp Please double check, dunno what we should do with the new h1-3 (naming bc it used to be heading 2..)

apps/babanews/tsconfig.json
Itrulia, line 1: Needed for tiptap

apps/onlinereports/src/render-element.tsx
dsheyp: H2 --> becomes Überschrift 4 --> the migrations seems to work, they render correctly

apps/tsri/src/components/break-blocks/tsri-sidebar-content.tsx
dsheyp: the migration from H2 to Überschrift 4 worked

dsheyp (approved):

PR #2819, opened 2026-06-14
feat(website): inteded routing via query

Itrulia:
requested by Hauptstadt

Mainly used to redirect the user back to the mitmachen page after trying to register a new account

### DIFF
--- a/apps/babanews/pages/login.tsx
+++ b/apps/babanews/pages/login.tsx
@@ -1,6 +1,7 @@
 import styled from '@emotion/styled';
 import { Typography } from '@mui/material';
 import {
+  IntendedRouteExpiryInSeconds,
   IntendedRouteStorageKey,
   LoginFormContainer,
   useUser,
@@ -12,7 +13,8 @@ import {
   SessionWithTokenWithoutUser,
 } from '@wepublish/website/api';
 import { useWebsiteBuilder } from '@wepublish/website/builder';
-import { deleteCookie, getCookie } from 'cookies-next';
+import { deleteCookie, getCookie, setCookie } from 'cookies-next';
+import { add } from 'date-fns';
 import { NextPageContext } from 'next';
 import { useRouter } from 'next/router';
 import { useEffect, useRef } from 'react';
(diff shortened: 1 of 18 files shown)


libs/authentication/website/src/lib/auth.constants.ts
```diff
@@ -1,2 +1,2 @@
 export const IntendedRouteStorageKey = 'auth.intended';
-export const IntendedRouteExpiryInSeconds = 2 * 60;
+export const IntendedRouteExpiryInSeconds = 5 * 60;
```
Itrulia, line 2: When using the email flow, the 2min might have expired already

libs/errors/website/src/lib/api-alert.tsx
```diff
@@ -24,16 +24,23 @@ export function ApiAlert({ error, ...props }: BuilderApiAlertProps) {
     elements: { Link },
   } = useWebsiteBuilder();
 
+  const loginLink =
+    typeof window !== 'undefined' ?
+      `/login?inteded=${window.location.pathname}${window.location.search}`
```
LukasSkywalker: inteded -> intended :)

apps/babanews/pages/login.tsx
```diff
@@ -37,6 +39,17 @@ export default function Login({ sessionToken }: LoginProps) {
     }
   }, [sessionToken, setToken]);
 
+  if (
+    router.query.intended &&
+    (router.query.intended as string).startsWith('/')
+  ) {
+    setCookie(IntendedRouteStorageKey, router.query.intended, {
```
LukasSkywalker, line 46: Wäre das nicht ein Kandidat für ein useEffect, damit es nicht bei jedem Render gesetzt wird?

LukasSkywalker (changes requested):

apps/babanews/pages/login.tsx
Itrulia: No, useEffect runs after the render, this would then also run after the redirect.
There's also not really anything rerendering the page either

libs/errors/website/src/lib/api-alert.tsx
Itrulia:
```suggestion
      `/login?intended=${window.location.pathname}${window.location.search}`
```

LukasSkywalker (approved):

PR #2825, opened 2026-06-17
fix(editor): display status as "used" for system-mail-only templates

dsheyp:


libs/mail/api/src/lib/mail-context.ts
```diff
@@ -197,14 +197,26 @@ export class MailContext implements MailContextInterface {
...
+        },
+      }),
+      this.prisma.userFlowMail.findMany({
+        include: {
+          mailTemplate: true,
+        },
+      }),
+    ]);
+
+    return [
```
Itrulia, line 213: nitpick:

```
[
  ...intervals.map(intervals => interval.mailTemplate?.externalMailTemplateId), 
  ...userFlowMails.map(userFlowMail => userFlowMail.mailTemplate?.externalMailTemplateId)
].flat()
```

Itrulia (approved):

PR #2832, opened 2026-06-25
migrate(tiptap): additional migrations for the slate to tiptap transition

dsheyp:


libs/richtext/editor/src/lib/richtext-editor.tsx
```diff
@@ -166,20 +166,25 @@ export const RichtextEditor = forwardRef<HTMLDivElement, RichtextEditorProps>(
       [autofocus]
     );
     const providerValue = useMemo(() => ({ editor }), [editor]);
+    const editorReady = !!editor && !editor.isDestroyed;
 
     useEffect(() => {
+      if (!editor) {
```
Itrulia:
```suggestion
      if (!editorReady) {
```

should probably this like the others, no?

libs/richtext/editor/src/lib/richtext-editor.tsx
```diff
@@ -190,52 +195,56 @@ export const RichtextEditor = forwardRef<HTMLDivElement, RichtextEditorProps>(
...
-              borderBottom: 1,
-              borderColor: 'divider',
-            }}
-          >
-            <MenuBar />
-          </Box>
-
-          <BubbleMenu />
-          <DragHandle />
+          {editorReady && (
```
Itrulia, line 198: Should probably just return null if editorReady is falsy instead of doing multiple checks.

Or is there a reason that wouldn't work?

libs/richtext/src/lib/to-plaintext.ts
```diff
@@ -1,15 +1,46 @@
...
-import { RichtextElements } from './json-format.interface';
+import {
+  RichtextElements,
+  RichtextJSONDocument,
+} from './json-format.interface';
+
+const nodeToPlaintext = (node: unknown): string => {
+  if (!node || typeof node !== 'object') {
+    return '';
+  }
```
Itrulia, line 10:
```suggestion
  }
  
```

libs/richtext/api/src/lib/slate-to-pm-migrator.ts
Itrulia, line 1: can we format this like the rest of the newer code? Like new lines around if blocks etc

libs/richtext/api/src/lib/slate-to-pm-migrator.ts
```diff
@@ -756,4 +777,250 @@ export class SlateToPmMigrator {
...
+    await this.remigrateRevisionTable(
+      'slate.remigrateBuggyPages',
+      'pages.revisions',
+      SlateToPmMigrator.BUGGY_TEXT_NODE,
+      blocks => this.migrateBlocksFromSlate(blocks)
+    );
+  }
+
+  // GENERAL: entity richtext columns re-migrated from their `slate*` source.
+  private static readonly BUGGY_ENTITY_COLUMNS: ReadonlyArray<{
```
Itrulia: no need for typing. Can do `as const` at the end instead

libs/richtext/api/src/lib/slate-to-pm-migrator.ts
```diff
@@ -13,11 +22,23 @@ const slateToPm = (content: any): RichtextElements | [] => {
...
 
+      // Links carry their label in `children`, not `title`; fall back to the url.
+      const text =
+        isLink ?
+          slateText(child) || child.title || child.url || ''
+        : child.text;
+
+      // ProseMirror rejects text nodes whose `text` is not a non-empty string
+      // (RangeError: Invalid text node in JSON / empty text nodes not allowed).
+      if (typeof text !== 'string' || text.length === 0) {
```
Itrulia, line 33: do we need to check for 0 or falsy?

libs/richtext/editor/src/lib/richtext-editor.tsx
dsheyp: yes, will update accordingly in next commit

libs/richtext/src/lib/to-plaintext.ts
dsheyp: ok, will update accordingly in next commit

libs/richtext/api/src/lib/slate-to-pm-migrator.ts
dsheyp: yes, will reformat in next commit

libs/richtext/api/src/lib/slate-to-pm-migrator.ts
dsheyp: yes, will update accordingly in next commit

libs/richtext/editor/src/lib/richtext-editor.tsx
dsheyp: not 100% sure here, but:

nested editor inside flex blocks caused issues --> 
* it seems EditorContent and paper-wrapper have to always stay mounted
* on a not-ready editor MenuBar/BubbleMenu/DragHandle/FooterActions will crash because each does useCurrentEditor().editor! and editor.isActive(...)
* returning null while !editorReady would also unmount EditorContent and the frame during the editor's not-ready/recreate window

libs/richtext/api/src/lib/slate-to-pm-migrator.ts
dsheyp: * make sure text IS a string
* if it is a string, make sure it is a non-empty string

checking for none-falsy only would allow e.g. 9 as a valid text --> but it's still invalid for tiptap --> two step check should eliminate that

Itrulia: Im talking about text.length, that should either not exist, be 0 or above 0.

Itrulia (approved):

PR #2834, opened 2026-06-29
improve(richtext-editor): re-migrate richtext in flex-block to avoid nulls

dsheyp:
* the ghost-link fix changes behavior #2832 already merged (removes the || title || url fallback that over-corrected) --> this here is an intentional follow-up correction, not a regression


libs/richtext/api/src/lib/slate-to-pm-migrator.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  new SlateToPmMigrator(undefined as any, undefined as any);
+
+const linkMarks = (doc: any): { text: string; href: string }[] => {
+  const out: { text: string; href: string }[] = [];
+  const walk = (node: any): void => {
+    if (Array.isArray(node)) {
+      node.forEach(walk);
+      return;
+    }
+    if (!node || typeof node !== 'object') {
```
Itrulia, line 14:
```suggestion

    if (!node || typeof node !== 'object') {
```

libs/richtext/api/src/lib/slate-to-pm-migrator.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      return;
+    }
+    if (node.type === 'text') {
```
Itrulia, line 18:
```suggestion

    if (node.type === 'text') {
```

libs/richtext/api/src/lib/slate-to-pm-migrator.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      for (const mark of node.marks ?? []) {
+        if (mark.type === 'link') {
+          out.push({ text: node.text, href: mark.attrs.href });
+        }
+      }
+    }
```
Itrulia, line 24:
```suggestion
    }
    
```

libs/richtext/api/src/lib/slate-to-pm-migrator.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    (node.content ?? []).forEach(walk);
+  };
+  walk(doc);
```
Itrulia, line 29:
```suggestion

  walk(doc);
  
```

Itrulia (approved):

PR #2851, opened 2026-07-10
feat(membership): exact per-periodicity member plan pricing

dsheyp:
Adds MemberPlan.periodicityPricing (nullable Json, additive migration) holding exact min/target/max amounts in cents per payment period for every enabled periodicity. Subscription monthlyAmount becomes a GraphQL Float so the derived monthly value can carry fractions (e.g. 500.00/year -> 4166.67 cents/month); all invoice-creating paths now round the period total through a single calculatePeriodAmount helper, making any period price exactly representable.

- billing: rounding in legacy renew, periodic-job cron, Payrexx provider and crowdfunding revenue; importSubscription now multiplies by periodicity like every other path
- validation: public create/update checks the period amount against the per-periodicity minimum (fallback: derived from amountPerMonthMin)
- editor: per-periodicity min/target/max price panel on the member plan form (delta vs derived monthly price is displayed, not entered); subscription edit view edits the exact period amount and derives the fractional monthly value, so saving no longer truncates period-priced subscriptions
- website: subscribe defaults/validation/summary texts and member plan picker prefer the exact per-periodicity prices; upgrade + slider rounding
- csv export rounds monthlyAmount; regenerated schema-v2 + graphql clients


libs/api/prisma/migrations/20260710120000_add_member_plan_periodicity_pricing/migration.sql
Itrulia, line 1: This migration isn't needed anymore, no?

libs/api/prisma/migrations/20260818120000_member_plan_periodicity_pricing_table/migration.sql
```diff
@@ -0,0 +1,40 @@
...
+    CONSTRAINT "member.plans.periodicity-prices_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE UNIQUE INDEX "member.plans.periodicity-prices_memberPlanId_periodicity_key" ON "member.plans.periodicity-prices"("memberPlanId", "periodicity");
+
+-- AddForeignKey
+ALTER TABLE "member.plans.periodicity-prices" ADD CONSTRAINT "member.plans.periodicity-prices_memberPlanId_fkey" FOREIGN KEY ("memberPlanId") REFERENCES "member.plans"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- DataMigration: copy existing JSONB entries into the new table
```
Itrulia: This migration was only needed for dev, no?

libs/block-content/api/src/lib/subscribe/subscribe-block.model.ts
```diff
@@ -24,6 +24,15 @@ registerEnumType(SubscribeBlockField, {
   name: 'SubscribeBlockField',
 });
 
+export enum SubscribePeriodicityDisplay {
+  Dropdown = 'dropdown',
+  OfferCards = 'offerCards',
```
Itrulia: What's offer cards?

libs/member-plan/api/src/lib/member-plan.model.ts
```diff
@@ -60,6 +60,31 @@ export class AvailablePaymentMethodInput extends OmitType(
...
+@ObjectType()
+export class PeriodicityPrice {
+  @Field(() => PaymentPeriodicity)
+  periodicity!: PaymentPeriodicity;
+
+  @Field({ nullable: true })
+  label?: string;
+
+  @Field(() => Int, { nullable: true })
+  amountMin?: number;
```
Itrulia: This should be required, no?

libs/api/prisma/schema.prisma
```diff
@@ -703,6 +703,24 @@ model AvailablePaymentMethod {
...
 }
 
+model MemberPlanPeriodicityPrice {
+  id         String   @id @default(dbgenerated("gen_random_uuid()"))
+  createdAt  DateTime @default(now())
+  modifiedAt DateTime @updatedAt
+
+  periodicity  PaymentPeriodicity
+  label        String?
+  amountMin    Int?
```
Itrulia: Should be required, no?

libs/api/prisma/migrations/20260711090000_add_member_plan_default_payment_periodicity/migration.sql
Itrulia: could we merge this with the other migration?

libs/api/prisma/migrations/20260710120000_add_member_plan_periodicity_pricing/migration.sql
dsheyp: there are two media-sites that already have it on prod so we need to keep this migration

Itrulia (approved):

PR #2852, opened 2026-07-13
feat(subscribe-block): improve periodicity pricing

dsheyp:


libs/block-content/website/src/lib/subscribe/subscribe-block.tsx
```diff
@@ -52,9 +55,20 @@ export const SubscribeBlock = ({
...
       discountCode,
+      periodicity,
     },
   } = useContext(BuilderRouterContext);
 
+  const defaultPaymentPeriodicity = useMemo(
+    () =>
+      Object.values(PaymentPeriodicity).find(
+        value =>
+          typeof periodicity === 'string' &&
```
Itrulia, line 66: Can it ever not be string?

Itrulia (approved):

PR #2894, opened 2026-07-20
feat(crowdfunding): campaign end date + daysRemaining

dsheyp:


libs/crowdfunding/api/src/lib/crowdfunding.resolver.ts
```diff
@@ -117,4 +117,18 @@ export class CrowdfundingResolver {
       subscriptions: await this.subscriptions(parent),
     });
   }
+
+  @ResolveField(() => Number, { nullable: true })
+  daysRemaining(@Parent() parent: PCrowdfunding) {
```
Itrulia, line 122: That should be on the client

libs/api/prisma/schema.prisma
```diff
@@ -1455,6 +1455,7 @@ model Crowdfunding {
   name                    String
   countSubscriptionsFrom  DateTime?
   countSubscriptionsUntil DateTime?
```
Itrulia, line 1457: Whats the difference between endsAt and countSubscriptionsUntil?

dsheyp: i noticed there's a countSubscriptionsUntil field and thought about using it for the endsAt purpose - but then voted against. reason(s):
* maybe due to its naming it definitely looked more like somethings that would be used for some type of statistics (which are probably not implemented yet), let's say to let admins plot a graph for a specific period and get insights on how their campaign performed over time
* that there exists a counter-part field (countSubscriptionsFrom) helped for me to read these two fields more as something targeted at counting subscriptions over a period (not necessarly the full campaign period, though) than describing a static campaign attribute
* thinking about it again, i realize that the way i looked at it was more like arguments to a method getSubscriptions() --> probably shoud have just adjusted the label a bit and used these fields

Itrulia: Those are the start and enddates of the campaign

dsheyp: exactly what i would have needed - didn't see the obvious that already existed

PR #2922, opened 2026-08-04
feat: add vouchers to upgrade and improve text handling

Itrulia:


libs/block-content/api/src/lib/subscribe/subscribe-block.resolver.ts
Itrulia, line 1: Changes in here fix that ids are given out that don't exist anymore

LukasSkywalker (approved):

PR #2929, opened 2026-08-11
fix: some master fixes from big update

Itrulia:


libs/block-content/website/src/lib/block-styles/image-slider/image-slider.tsx
```diff
@@ -84,6 +86,12 @@ export const ImageSlider = ({
                 <Image
                   caption={image.caption}
                   image={image.image}
+                  css={css`
```
Itrulia: Note for myself: Move this out and style this via styled

LukasSkywalker (approved):

PR #2933, opened 2026-08-12
feat(munotgloeggli): self service

Itrulia:


apps/munotgloeggli/src/theme.tsx
```diff
@@ -66,14 +65,7 @@ export const ttNorms = localFont({
   ],
 });
 
-const {
-  palette: { augmentColor },
-} = WePTheme;
-
-const theme = createTheme(WePTheme, {
-  palette: {
-    primary: augmentColor({ color: { main: '#dc0d15' } }),
```
Itrulia, line 75: Needs to be set on self service

LukasSkywalker (approved):

PR #2937, opened 2026-08-14
fix(media): ignore nx cache during build

LukasSkywalker:


Dockerfile
```diff
@@ -261,7 +261,7 @@ COPY . .
 COPY ./apps/media/package.json ./package.json
 COPY ./apps/media/package-lock.json ./package-lock.json
 RUN npm ci --no-audit --no-fund
-RUN npx nx build media --ignore-nx-cache && \
```
Itrulia, line 264: Please also fix for api-example and editor

Itrulia (approved):

PR #2940, opened 2026-08-17
refactor(api): db add timezone to every date

Itrulia:


libs/api/prisma/migrations/20260817121431_add_timezones_to_dates/migration.sql
```diff
@@ -0,0 +1,440 @@
+-- DropView recreated below. SQL blocks changing the column types while used in a view
+DROP VIEW IF EXISTS "articles.revisions.published";
+DROP VIEW IF EXISTS "articles.revisions.pending";
+DROP VIEW IF EXISTS "articles.revisions.draft";
+DROP VIEW IF EXISTS "pages.revisions.published";
+DROP VIEW IF EXISTS "pages.revisions.pending";
+DROP VIEW IF EXISTS "pages.revisions.draft";
+
+-- AlterTable
```
Itrulia, line 9: I think from prisma upgrade or went missing when adding the goodies

LukasSkywalker (approved):

PR #2948, opened 2026-08-20
fix(core): REF-23 send payment successs mail in all cases not just in nighly job

elias-summermatter:
… nighly job


libs/membership/api/src/lib/renewal-mail/renewal-success-mail.service.ts
```diff
@@ -0,0 +1,167 @@
...
+      return;
+    }
+
+    const claim = await this.prisma.invoice.updateMany({
+      where: {
+        id: invoiceId,
+        renewalSuccessMailSentAt: null,
+        suppressRenewalSuccessMail: false,
+      },
+      data: { renewalSuccessMailSentAt: new Date() },
```
LukasSkywalker: If the mail delivery fails on Line 118, the mail is never retried, because it is already marked as sent.

libs/membership/api/src/lib/periodic-job/periodic-job.service.ts
```diff
@@ -36,6 +36,7 @@ import { Action } from '../subscription-event-dictionary/subscription-event-dict
 import { SubscriptionService } from './subscription.service';
 import { PeriodicJobRunObject } from './periodic-job.type';
 import { getMaxTake } from '@wepublish/utils/api';
+import { RenewalSuccessMailService } from '../renewal-mail/renewal-success-mail.service';
```
LukasSkywalker: This could import `InvoicePaidNotifier`, and then call `await this.invoicePaid.notify(id);` on line 437 directly. That makes sure the invoicePaid notification is called the same way everywhere.

Would also allow to remove `onInvoicePaid` in `renewal-success-mail.service.ts`, since it duplicates the error handling of the InvoicePaidNotifier.

`RenewalMailModule` can then be reduced

```diff
-  exports: [RenewalSuccessMailService, INVOICE_PAID_LISTENER],
+  exports: [INVOICE_PAID_LISTENER],
```

LukasSkywalker (approved):

PR #2949, opened 2026-08-21
feat(reflekt): REF-20 improve the subscription-upgrade

dsheyp:


apps/reflekt/src/components/reflekt-subscribe-crowdfunding.tsx
```diff
@@ -272,10 +272,9 @@ export const ReflektCrowdfundingMemberPlanItem = forwardRef<
   const radioGroup = useRadioGroup();
   const isChecked = props.checked ?? radioGroup?.value === id;
   const radioInputRef = useRef<HTMLInputElement>(null);
-  const {
-    formState: { errors },
-    setValue,
-  } = useFormContext();
+  const form = useFormContext() as ReturnType<typeof useFormContext> | null;
```
Itrulia, line 275: should only be null if no provider is set, should imo not happen and throw an error

libs/membership/website/src/lib/upgrade/upgrade.tsx
```diff
@@ -322,6 +325,7 @@ export const Upgrade = ({
               {...field}
               onChange={memberPlanId => field.onChange(memberPlanId)}
               memberPlans={availableMemberplans}
+              alwaysShow
```
Itrulia, line 328: This property is not meant to be used like this but to allow medias to hide/show it.

Like:

```
export const HauptstadtMemberPlanPicker = styled(
  forwardRef<HTMLButtonElement, BuilderMemberPlanPickerProps>(
    function HauptstadtMemberPlanPicker(props, ref) {
      return (
        <MemberPlanPicker
          {...props}
          alwaysShow
        />
      );
    }
  )
)`
  ${MemberPlanPickerRadios} {
    gap: ${({ theme }) => theme.spacing(4)};
  }
`;
```

PR #2955, opened 2026-08-24
fix(editor): harden preview

dsheyp:


apps/editor/src/app/routes/pages/pageEditor.tsx
```diff
@@ -729,10 +729,12 @@ function PageEditor() {
                             { previewJwt: token },
                             targetOrigin
                           );
-                          window.removeEventListener('message', handleMessage);
                         }
                       };
                       window.addEventListener('message', handleMessage);
+                      setTimeout(() => {
+                        window.removeEventListener('message', handleMessage);
```
elias-summermatter: Why this timeout? Why 60sec? Is there not other conditon where the 60sec can be reached? Is there maybe a better whay to detect this which is more clear?

elias-summermatter (approved):

Itrulia (approved):

PR #2959, opened 2026-08-25
chore(core): Add npm script to reset password locally

LukasSkywalker:


tools/set-password.js
```diff
@@ -0,0 +1,72 @@
...
+const { Client } = require('pg');
+
+const [email, password] = process.argv.slice(2);
+
+if (!email || !password) {
+  console.error('Usage: npm run set-password -- <email> <password>');
+  process.exit(1);
+}
+
+const databaseUrl = (
```
Itrulia, line 21: I guess we only want to ever use .env as a fail safe instead of using the dotenv package. So that we dont accidentally wipe a prods database?

PR #2973, opened 2026-08-31
fix(wnti): Fix rich text list styles

ahmedbajra:
Use `rem` instead of `em` in custom `RichText`.

### DIFF
--- a/apps/wnti/src/components/tsri-richtext.tsx
+++ b/apps/wnti/src/components/tsri-richtext.tsx
@@ -3,6 +3,6 @@ import { RichTextBlock } from '@wepublish/block-content/website';
 
 export const TsriRichText = styled(RichTextBlock)`
   .MuiTypography-body1 {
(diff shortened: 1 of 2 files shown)


apps/wnti/src/components/tsri-richtext.tsx
```diff
@@ -3,6 +3,6 @@ import { RichTextBlock } from '@wepublish/block-content/website';
 
 export const TsriRichText = styled(RichTextBlock)`
   .MuiTypography-body1 {
-    font-size: 1.25em;
```
elias-summermatter, line 6: I this Problem just on the wnti richtext or also somewhere else where the Block is used?

ahmedbajra: Fair point, Flimmer also used em font size for rich text. Added another commit.

dsheyp (approved):

PR #2988, opened 2026-09-02
feat(hauptstadt): nachtleben archiv HAS-177

dsheyp:


libs/article/api/src/lib/article.model.ts
```diff
@@ -287,6 +287,11 @@ export class ArticleFilter {
   authors?: string[];
   @Field(() => [String], { nullable: true })
   tags?: string[];
+  @Field(() => [String], {
+    nullable: true,
+    description: 'Only include articles that have every one of these tags',
+  })
+  tagsAll?: string[];
```
Itrulia:
```suggestion
  allTagsIn?: string[];
```

to fit tagsNotIn

libs/article/api/src/lib/article.resolver.ts
```diff
@@ -279,12 +279,16 @@ export class ArticleResolver {
 
   @ResolveField(() => String, { nullable: true })
   async url(@Parent() parent: PArticle) {
-    return this.urlAdapter.getArticleUrl(parent);
+    const tags = await this.tagDataLoader.load(parent.id);
```
Itrulia: can we load this inside the hauptstadt url adapter instead? So we don't have to load it for all medias (same with the one below)

apps/hauptstadt/pages/arch/[slug].tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return <StandaloneArticlePage slug={slug as string} />;
+}
+
+export const getStaticPaths = () => ({
+  paths: [],
+  fallback: 'blocking',
+});
+
+export const getStaticProps: GetStaticProps = async ({ params }) => {
```
Comment (Itrulia) [inline apps/hauptstadt/pages/arch/[slug].tsx]: Could we import this from /a/[slug]?

apps/hauptstadt/src/components/standalone-article-page.tsx
Itrulia, line 1: We should import from /a/[slug] instead of moving this out. To ensure it's always the same

Itrulia (approved):

PR #2997, opened 2026-09-03
feat(tsri): TSRI-195 hideAuthor hides author on teasers & hides article on author page

dsheyp:


libs/article/api/src/lib/article.service.ts
```diff
@@ -1047,11 +1047,16 @@ const createAuthorFilter = (
       },
     };
 
+    const publishedAuthorFilter: Prisma.ArticleRevisionWhereInput = {
```
Itrulia: We should probably still be allowed to filter on this even when hideAuthor is true. As it's hiding it from the UI

Itrulia (changes requested):

libs/article/api/src/lib/article.service.ts
dsheyp: agreed. hideAuthor is presentation, not visibility --> will make the exclusion an opt-in filter field (like includeHidden) defaulting to today's behavior and tsri's author page opts in

Itrulia (approved):

PR #2998, opened 2026-09-04
feat(haupstadt): Allow ordering authors and setting their roles

LukasSkywalker:


apps/api-example/prisma/seed.ts
```diff
@@ -799,11 +799,15 @@ async function seedArticles(
```
LukasSkywalker, line 799: Would have preferred to pack `role` and `position` into the author object to get

```js
{ name: 'Foo', position: 1, role: 'Text' }
```

instead of needing to create `RevisionAuthor` and nesting `Author` inside there

```js
{ position: 1, role: 'Text', author: { name: 'Foo' } }
```

because that would have required much fewer destructuring changes. This interferes with the Apollo cache though, because an author with different roles in different articles shares a spot in the cache. Hence this larger refactoring.

Itrulia (approved):

apps/hauptstadt/src/components/hauptstadt-author-chip.tsx
```diff
@@ -1,5 +1,5 @@
 import { BuilderAuthorChipProps } from '@wepublish/website/builder';
 
-export function HauptstadtAuthorChip({ author }: BuilderAuthorChipProps) {
-  return author.name;
+export function HauptstadtAuthorChip({ author, role }: BuilderAuthorChipProps) {
+  return role ? `${author.name} (${role})` : author.name;
```
Itrulia, line 4: role.trim() potentially? Or maybe doing this when saving

LukasSkywalker: I already trim when saving (as well as when peering, and the migration script as well): https://github.com/wepublish/wepublish/pull/2998/changes#diff-75427e8639a6b68f7d22dd825b68f5b9aea12dfd11980473414f3fb94fc76979R33
Should we also trim when displaying?

EDIT: Link doesn't work that well, so here's the diff:
image

Itrulia: Then all good and no trim needed

PR #3013, opened 2026-09-08
feat(editor): Add "current date" placeholder to mail template

ahmedbajra:


libs/membership/api/src/lib/mail-send/placeholder-check.ts
Itrulia, line 1: missed 2 console logs

ahmedbajra: Find & replace isn't what it used to be :)

Itrulia (approved):

PR #3017, opened 2026-09-09
feat(editor): tell an external app who is in the iframe and which article is open

jolandaspiess:
**Draft on purpose.** The editor side of an assistant we are piloting for hauptstadt; opening it early so the two design questions at the bottom can be answered before anyone polishes anything.

## What it does

An external app registered in the editor gets loaded today with nothing but its url, so it never learns who is sitting in the iframe or what they are working on. Two commits change that, using the `createExternalAppToken` mutation that already exists on the api but that no editor code called.

**1. Token in the url fragment.** `ExternalAppIframeView` mints a token once the app is loaded and renders the iframe only when it is there:

```
<url>#token=<jwt>
```

The token travels in the fragment, never as a query parameter: fragments reach no server and no log, and the app is expected to strip it from the address bar after trading it for its own session. A failed mutation is shown rather than swallowed, because loading the app without a token would silently sign the editor out of it.

**2. The app beside the article it works on.** Every app with `target = IFRAME` gets a button in the article editor's toolbar, next to Metadata, labelled with its own name and icon. It opens a drawer holding the app, which is told the open article in the same fragment:

```
<url>#token=<jwt>&articleId=<id>&revision=draft
```

so the app needs no article picker. The button appears only for a saved article, and the token is minted when the drawer opens, never on page load: the editor asks for no token for a tool nobody used.

The fragment is built in one place, `useExternalAppSrc`, which both callers use. `AppIcon` moves out of the external-apps dashboard into its own file because two callers need it now.

## How it was checked

- 97 tests in `editor` green, 15 of them new: the hook, the panel and the full-page view.
- `eslint` adds no warning, `prettier --check` clean, `tsc -p apps/editor/tsconfig.app.json --noEmit` clean.
- Measured against a local editor and a real external app: the minted jwt carries `iss` of the api and `aud` of the registered app url, the app trades it for its own session, and it comes up on the open article without a login form and without an article picker.

## Two questions for the team

1. **Which apps belong in the article editor?** This takes the cheap route: every app with `target = IFRAME`. The honest downside is that any future iframe app shows up there whether it has anything to do with an article or not. A third target value, `ARTICLE_EDITOR` beside `BLANK` and `IFRAME`, would say it properly, but that means a prisma migration, the api model, the schema, codegen and the registration form. Happy to do it that way if you prefer it.
2. **Where should the panel sit?** This follows the pattern the article editor already uses twice: a toolbar button and a drawer of size `md`. A docked sidebar would be closer to "an assistant reading along", but it changes the editor's layout, which felt like your call rather than ours.

Also worth knowing: `createExternalAppToken` carries `@Authenticated()` and no `@Permissions`, unlike the create/update/delete mutations beside it. Any signed-in user can therefore mint a token in their own name for any registered app. That is what makes this change small, and it may well be intended, but it seemed worth naming.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/editor/src/app/routes/articles/articleEditor.tsx
+++ b/apps/editor/src/app/routes/articles/articleEditor.tsx
@@ -8,6 +8,10 @@ import {
   DialogTitle as MuiDialogTitle,
 } from '@mui/material';
 import {
+  ExternalAppsTarget,
+  useExternalAppsQuery,
+} from '@wepublish/editor/api';
+import {
   CreateArticleMutationVariables,
   EditorBlockType,
   FullAuthorFragment,
@@ -77,6 +81,8 @@ import {
 } from 'rsuite';
 
 import { openPreviewWindow } from '../../openPreview';
+import { AppIcon } from '../externalApps/appIcon';
+import { ExternalAppArticlePanel } from '../externalApps/externalAppArticlePanel';
 
 const IconButtonMarginTop = styled(RIconButton)`
   margin-top: 4px;
@@ -228,6 +234,21 @@ function ArticleEditor() {
   );
 
   const articleID = id || createData?.createArticle.id;
+
+  // External apps that run in an iframe get a button beside the metadata, so
+  // an assistant can look at the article that is open instead of asking which
+  // one it is. The token is minted when the drawer opens, never before: the
+  // editor asks for no token for a tool nobody used.
+  const [openExternalAppId, setOpenExternalAppId] = useState<string | null>(
+    null
+  );
+  const { data: externalAppsData } = useExternalAppsQuery({
+    variables: { filter: { target: ExternalAppsTarget.Iframe } },
+  });
+  const externalApps = externalAppsData?.externalApps ?? [];
+  const openExternalApp = externalApps.find(
+    app => app.id === openExternalAppId
+  );
 
   const {
     data: articleData,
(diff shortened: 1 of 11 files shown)


jolandaspiess:
On the red checks, so nobody has to dig: as far as I can tell none of them are caused by this branch.

**Test (both runs).** The test commands themselves pass — `npm run test-backend` and `npm run test-website` are both green, and the three new spec files run inside `test-backend`:

```
✓ editor  src/app/routes/externalApps/useExternalAppSrc.spec.tsx        (6 tests)
✓ editor  src/app/routes/externalApps/externalAppIframeView.spec.tsx    (5 tests)
✓ editor  src/app/routes/externalApps/externalAppArticlePanel.spec.tsx  (4 tests)
```

The failing step is the one after them, `Upload coverage to Codecov`:

```
429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue."}
```

A pull request from a fork gets no repository secrets, so the uploader falls back to tokenless upload and hits the public rate limit.

**Docker base image / Build and push base image.** Same cause, earlier in the job:

```
Secret source: None
##[error]Username and password required
```

The docker login step has no credentials on a fork PR. This step is green on the in-repo PRs open right now (#3016, #3011, #3010).

**Dependency Scan.** `trivy fs` with `severity: CRITICAL,HIGH` and `exit-code: 1`, so it fails on the repository's dependency tree rather than on a diff. It is red on #3016, #3011 and #3010 as well. This branch adds no dependency.

Green: Lint, Typecheck, CodeQL, Secret Scan.

Happy to rebase, split the two commits, or rebuild it around a dedicated `ARTICLE_EDITOR` target value if that is the direction you prefer — see the two questions in the description.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3021, opened 2026-09-10
fix(website): normalize malformed nested lists in richtext rendering

dsheyp:
e.g. [redaktionsstatut](https://tsri.ch/redaktionsstatut)


Itrulia (changes requested):
Can we fix the data instead? As we would need to keep this one going forever

PR #3045, opened 2026-09-18
fix(reflekt): make sure reflekt sitemap lists up to 500 articles

dsheyp:


Itrulia (changes requested):
1. This is not improving SEO, sitemap is for indexing once and not for repeat.
2. It should use pagination instead
3. This wouldn't solve freshly imported articles for example, 1 time sitemap upload for that would still be better.

libs/feed/website/src/lib/sitemap-generator.ts
```diff
@@ -60,7 +96,9 @@ export const generateSitemap =
             <url>
                 <loc>${article.url}</loc>
                 <lastmod>${article.latest.publishedAt}</lastmod>
-
+${
+  qualifiesAsNews(article.publishedAt) ?
```
Itrulia, line 100: why?

libs/feed/website/src/lib/sitemap-generator.ts
```diff
@@ -20,12 +30,38 @@ export const generateSitemap =
...
+      if (newsCutoff == null) {
+        return true;
+      }
+
+      const published = new Date(publishedAt ?? '').getTime();
+
+      return Number.isFinite(published) && published >= newsCutoff;
+    };
+
+    const mostRecentPublishedAt = [
```
Itrulia, line 46: this is incorrect information and google will ignore it according to their guidelines

libs/feed/website/src/lib/sitemap-generator.ts
dsheyp: google is telling developers to do so in their docs for news-sitemaps: [google docs](https://developers.google.com/search/docs/crawling-indexing/sitemaps/news-sitemap?hl=de)

libs/feed/website/src/lib/sitemap-generator.ts
dsheyp: not sure this isn't correct: in some places in their docs google is telling developers to do so: [google docs](https://developers.google.com/search/docs/crawling-indexing/sitemaps/news-sitemap?hl=de)

they still might contradict themeselves in other places, though....

dsheyp:
* in the same place [google docs](https://developers.google.com/search/docs/crawling-indexing/sitemaps/news-sitemap?hl=de) google also makes clear that a sitemap isn't a one-time-thing --> they actually recommend updating it each time something changes as they will re-crawl it
* they even mention that following their recommended practice of removing news older than 2 days may lead to a temporarly empty sitemap, which they say, won't cause any problems in terms of the google search
* an up-to-date sitemap reflecting the latest publications and highlighting them as "news" seems to totally make sense and also seems to be exactly what google is recommending

libs/feed/website/src/lib/sitemap-generator.ts
Itrulia: I dont see that in the docs.

Itrulia: I had to read it multiple times before I caught it, yes I am wrong here. I never seen this when reading the docs

libs/feed/website/src/lib/sitemap-generator.ts
Itrulia: I don't see that in the documentation you linked.

Here's my sources:

https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#additional-notes-about-xml-sitemaps

https://developers.google.com/search/blog/2023/06/sitemaps-lastmod-ping#the-lastmod-element

dsheyp: you're right, lastmod isn't mentioned in the docs i linked and, imho, my new implementation is partly ok but also has an issue that i only noticed now:
* the good:
  * mostRecentPublishedAt is being used for the homepage-lastmod only
  * the homepage is showing the latest articles in the slider(s)
  * lastmod for the homepage will be set to the published-at of the latest article --> makes sense, as updating the slider(s) with the latest article can be considered a significant change according to the google docs you linked and: it would go un-noticed if only looking at the modified-at / published-at of the homepage
* the bad
  * lastmod for the homepage will be set to the published-at of the latest page-change (if more recent than the latest article)
  * --> this doesn't make sense, as the homepage holds no page-teasers --> will remove it
* it should probably be combined with a modified-at-check on the homepage that includes some diff to make sure only siginificant changes according to the docs are considered (--> a bit more complicated and maybe out of scope for now as most home-page-changes will occur due to article-publication-events)

PR #3046, opened 2026-09-18
feat(factuel): missing translations + hide nav menu

ahmedbajra:
* add missing translations across components / pages
* add new `hideMenu` prop for Navbar component
* localize country-select component


apps/editor/src/app/locales/en.json
dsheyp, line 1: 1. a german key-name was used: `"login-fuer-abonnent-innen"` - a deliberate choice?
2. also a german value `"Login für Abonnent*innen"`--> in en.json - a deliberate choice?

ahmedbajra: Must've been the i18n plugin I tried out. I ultimately added another entry directly in the overrides as the plugin didn't quite do what it should have.
- removed

.vscode/settings.json
dsheyp, line 1: according to the previous comment an i18n extension was tried out - was it the one that left these settings? if so the file could be removed if it isn't in used any more

dsheyp (approved):

PR #3047, opened 2026-09-18
fix(factuel): final visual fixes

ahmedbajra:


libs/navigation/website/src/lib/navbar/navbar.tsx
dsheyp, line 1: the new min-height property in the globally defined navbar should probably be guarded a bit from having unintended side-effects, maybe something like

`${({ hideMenu }) => hideMenu && css``min-height: var(--navbar-height, 52px);``}`

single backticks would be correct but gh won't let me write them, so they're doubled

dsheyp (changes requested):

libs/navigation/website/src/lib/navbar/navbar.tsx
ahmedbajra: added with [#e39c5c3](https://github.com/wepublish/wepublish/pull/3047/commits/e39c5c3c2d53097322bb9ecaea149ef210662a67)

dsheyp (approved):

